PR #5, opened 2016-06-30
Changes to package.json to run on Windows

jjeffryes:
- this is submitted just for comparison. Windows seems to need these changes to run properly, but there may be a better way to get around it.


rmisio:
This will work on Win, but not Linux (robbing Peter to pay Paul :)

Closing in liu of #6

PR #7, opened 2016-07-04
Page nav

rmisio:
This PR includes a very-simplified first version of the Page Nav and a few bits of infrastructure.


js/models/LocalSettings.js
```diff
@@ -0,0 +1,36 @@
+import LocalStorageSync from '../utils/backboneLocalStorage';
```
rmisio: Made a model that syncs via LocalStorage to contain our local settings. Rather than directly going into LocalStorage, this gives us a centralized API that relevant areas of the app can read / write from. It also gives us a bunch of model sugar, like a place to centralize validation, change events, etc...

js/views/PageNav.js
```diff
@@ -0,0 +1,67 @@
...
+      remote.getCurrentWindow().unmaximize();
+      // this.$('.js-navMax').attr('data-tooltip', window.polyglot.t('Maximize'));
+    } else {
+      remote.getCurrentWindow().maximize();
+      // this.$('.js-navMax').attr('data-tooltip', window.polyglot.t('Restore'));
+    }
+  }
+
+  render() {
+    loadTemplate('pageNav.html', (t) => {
```
rmisio: I also slightly changed the loadTemplate API in regards to the file path we pass in. Whereas before, we passed in the full path of the template relative to the app root (e.g. `./js/templates/pageNav.html`), now, by default, loadTemplate will assume that the root is the template folder, so the path you would pass in to reference the template from the example above would be simply `pageNav.html`.

You can pass in an optional root argument if you want a different root path (but, I don't think anywhere in v1 we referenced a template outside of the `templates` folder).

jjeffryes: I think the chance of having templates outside of the template folder is pretty small, and that's a fine rule to have in place.

js/start.js
```diff
@@ -0,0 +1,11 @@
+import $ from 'jquery';
+import app from './app';
+import LocalSettings from './models/LocalSettings';
+import PageNav from './views/PageNav.js';
+
+app.localSettings = new LocalSettings({ id: 1 });
+app.localSettings.fetch().fail(() => app.localSettings.save());
```
jjeffryes: Why does the localSettings model save if the fetch fails?

js/utils/backboneLocalStorage.js
```diff
@@ -0,0 +1,17 @@
...
+// We are wrapping the above Backbone localStorage adapter to slightly
+// change its behavior. By default, it overrides Backbone.sync globally,
+// forcing all models and collections to sync to localStorage
+// We are modifying it to leave the default sync in tact, and you can opt in on
+// a per model / collection basis by setting it's sync method to
+// Backbone.LocalStorage.sync. You also still need to set the localStorage attribute,
+// as described in the docs.
+
+import Backbone from 'backbone';
+import localStorageSync from 'backbone.localstorage'; // eslint-disable-line
```
jjeffryes: It's better to disable the specific eslint rule instead of the whole line, just in case there's also an unintentional error, and to make it clear which rule is being avoided.

js/utils/loadTemplate.js
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  _.templateSettings.variable = 'ob';
+
+  let template = templateCache[templateFile];
+
+  if (!template) {
+    const file = fs.readFileSync(path.join(root, templateFile), 'utf8');
+    template = _.template(file);
+    templateCache[templateFile] = template;
```
jjeffryes: Caching the templates is a neat idea. Good work.

js/start.js
rmisio: The fetch will fail the first time you load the app because the model was never saved in local storage. In that case, the save will save whatever the default options are.

js/utils/backboneLocalStorage.js
rmisio: Ha! That's a line a copied from v1, that I assume you put in (because I didn't). But, I agree and I'll clean up your mess (joking!... lol).

js/views/PageNav.js
```diff
@@ -0,0 +1,67 @@
+import electron from 'electron';
+import { View } from 'backbone';
+import loadTemplate from '../utils/loadTemplate';
+import app from '../app';
+
+const remote = electron.remote;
+
+export default class PageNav extends View {
```
jjeffryes: This has the functionality of the appbar view. Do you intend to fold that into the pageNav?

js/utils/backboneLocalStorage.js
jjeffryes: Well, my bad then!

rmisio: totally kidding... you're point is very valid. Anyhow, it's fixed.

js/views/PageNav.js
rmisio: Yes, according to Wolf there won't be an appbar and the controls are back in pageNav, so pageNav is going to have to be there from the start.

Once we start implementing the start-up process we can manage how to handle if the app isn't for the full functionality of the pageNav yet, whether it's disabling the address bar and drop-down menus, not displaying them or something else.

rmisio: @morebrownies 

I guess that also may mean, that once the app is loaded, we never want to obscure the pageNav, which might be a good thing. Any loading process should allow the user to implicitly cancel it by navigating to a different page and it's up to us a devs to handle that case.

morebrownies: Yes, the pageNav / address bar should always be available. Horrible UX to block it.

jjeffryes: As long as it's a hard rule that the nav bar is never covered, I think that's fine. 

But that raises questions about situations where the navbar should not be functional, such as when the server can't be contacted, or during onboarding.

PR #8, opened 2016-07-08
Style sheet

jjeffryes:
- this adds base styles to the app
- the styles are applied to the page nav
- to view the styleSheet, enter window.location.href="../styleSheet.html" in the console


js/templates/pageNav.html
```diff
@@ -1,12 +1,51 @@
...
+      <div class="rowDivV colBorderBackground"></div>
+      <div>
+        <div class="flexVCent">
+          <a class="iconBtn ion-android-notifications"></a>
+          <a class="discSm border colBorder"
+          <% if(ob.navimage){ %>
+          //this will be updated to the real image parameter
+          style="background-image: url(ob.navImage), url('imgs/defaultUser.png')"
+          <% } else { %>
+            style="background-image: url('imgs/defaultUser.png')"
```
rmisio: Since our app is actually served from .tmp, we would need to make image paths relative to there. So, I think it needs to be `../imgs/defaultUser.png`.

js/templates/pageNav.html
```diff
@@ -1,12 +1,51 @@
...
+        <div class="flexVCent">
+          <input type="text" class="flexGrow" placeholder="Type a handle or OpenBazaar ID" />
+        </div>
+      </div>
+      <div class="rowDivV colBorderBackground"></div>
+      <div>
+        <div class="flexVCent">
+          <a class="iconBtn ion-android-notifications"></a>
+          <a class="discSm border colBorder"
+          <% if(ob.navimage){ %>
```
rmisio: Even though we're not linting templates, can we mostly follow the linting style for JS in templates, which would mean a space before and after the parenthesis?

jjeffryes: Good call.

styles/components/_layout.scss
```diff
@@ -0,0 +1,94 @@
+//containers should have one of the standard padding sizes
```
rmisio: Conceptually, what is the difference between grid and layout?

js/templates/pageNav.html
jjeffryes: I wasn't sure how we wanted to handle that, I put a copy of the imgs in .tmp, since it seemed kind of weird to have everything else copied there but not the images. 

We can make the path work from .tmp, but we'll have to ignore IDE errors when the editor can't find the path from the template and styles directory (this will also affect images in css).

styles/components/_layout.scss
jjeffryes: That's a good question. Grid should be for only grid classes, and layout should be for everything that isn't part of the grid. 

That got muddy very quickly since some of the flex classes are very useful for non-grid layouts, but it didn't make sense to put them in a different file.

I'll merge them.

js/templates/pageNav.html
rmisio: Hmm... those errors are IDE specific. Mine doesn't throw them. Would it be a big problem to disable them?

I don't think copying the image files into .tmp is a bad idea, but I don't think it should be a manual thing. I can integrate it into the tooling, but to do it right we'd need to watch the images folder and then copy over any new / updated images and make sure the sequencing is right with the other tooling (browsersync, etc...). It just takes a bit of work and I think we might have bigger fish to fry (especially if this is to accomodate a specific IDE - for all I know, my IDE might be the minority by not flagging such things...).

jjeffryes: I spent a little time looking at how do sync the img folder, I don't think it'd be super difficult. But at the same time the images folder would rarely change, so doing it manually wouldn't be terrible.

For now, I've taken it out, but I feel like there could be issues if we reference images in the css, since the linter will probably flag those and fail the commit.  If that happens we can revisit it.

PR #10, opened 2016-07-12
Router

rmisio:
A initial work-through of the router with some dummy page views.

To test, play around with some of the dummy user guids and handles enumerated in the `usersCl` variable in start.js.

Some urls to try in the nav bar:
\<guid\>
@\<handle\>
@\<handle\>/\<tab\>
@<handle>/channel/\<category\>/\<layer\>
\<guid\>/\<tab\>
\<guid\>/channel/\<category\>/\<layer\>

and via location.href, you could try:
`location.href="transactions"`
`location.href="transactions/<tab>"`

Also feel free to try some bogus routes via both methods.


js/router.js
```diff
@@ -1,15 +1,147 @@
...
+
+    this.pageNavVw = options.pageNavVw;
+
     const routes = [
-      // ['home/:state(/:searchText)', 'home'],
-      // ['userPage/:userID(/:state)(/:itemHash)(/:skipNSFWmodal)', 'userPage'],
-      // [/^@([^\/]+)(.*)$/, 'userPageViaHandle'],
-      // ['transactions/:state(/:orderID)(/:tabState)', 'transactions'],
-      // ['settings/:state', 'settings'],
+      [/^@([^\/]+)[\/]?([^\/]*)[\/]?([^\/]*)[\/]?([^\/]*)$/, 'userViaHandle'],
```
rmisio: You could use regular expressions in routes, which are very powerful, or use the simpler syntax (e.g. in the transactions routes below), but you can't mix the two.

In the regular expression syntax, `[\/]?([^\/]*)` is more or less equivalent to `(:optional-path-frag)`

js/router.js
```diff
@@ -1,15 +1,147 @@
...
+  userViaHandle(handle, ...args) {
+    this.getGuid(handle).done((guid) => {
+      this.user(guid, ...args);
+    }).fail(() => {
+      this.userNotFound();
+    });
+  }
+
+  user(guid, tab, ...args) {
+    tab = tab || 'store'; // eslint-disable-line no-param-reassign
```
rmisio: The linter is not happy because we should really be using ES6's default arg syntax here where `tab = 'store'` could be declared right in the function definition. The problem is that only assigns the default if the argument comes in as `undefined`, whereas the Router is sending unmatched path fragments as `null`. So, we're putting in the lint override and doing it the old school way.

js/router.js
```diff
@@ -1,15 +1,147 @@
...
     ];
 
-    routes.forEach((route) => this.route.apply(this, route));
+    routes.slice(0)
+      .reverse()
+      .forEach((route) => this.route.apply(this, route));
+
+    this.setAddressBarText();
+
+    $(window).on('hashchange', () => {
```
rmisio: The router is updating the address bar in the nav bar based upon the window's hashchanges. So, views shouldn't have to directly update that. All they need to do is to use the routers `navigate` method when their state should be represented by a different url, for example:

```
app.router.navigate('@themes/store')
```

js/utils/selectors.js
```diff
@@ -0,0 +1,17 @@
+import $ from 'jquery';
```
rmisio: Created this module as a place to centralize and cache global selectors (i.e. selectors that live outside any particular view). I took the approach of making them methods that will do the initial DOM query the first time they are called, which means if for some reason someone follows a code path where they are never needed, we'll save ourself an unnecessary DOM query.

js/views/PageNav.js
```diff
@@ -57,11 +58,44 @@ export default class PageNav extends View {
...
+  onKeyupAddressBar(e) {
+    if (e.which === 13) {
+      let text = this.$addressBar.val();
+
+      if (text.startsWith('ob://')) text = text.slice(5);
+
+      if (text.charAt(0) === '@' && text.length > 1) {
+        // a handle
+        app.router.navigate((text.split(' ')[0]), { trigger: true });
+      } else if (text.startsWith('Qm')) {
```
rmisio: For now, I'm assuming v2 guids we'll be prefaced with `Qm`. @cpacia did not seem to 100% guarantee this would be the case, but determining if a string is a valid base58 string is not a reliable way to distinguish between a guid and tag, since `shoes` is a valid base58 string (but obviously not a GUID).

So, I'm thinking we will need some prefix.

styles/_base.scss
```diff
@@ -121,3 +121,13 @@ form {
 .colEmphasis1 {
   background-color: $emphasis1;
 }
+
+#pageContainer {
```
rmisio: if you have a better place for this to live, I'm all ears.

js/router.js
```diff
@@ -1,15 +1,153 @@
...
+      this.setAddressBarText();
+    });
+  }
+
+  setAddressBarText() {
+    if (location.hash.startsWith('#transactions')) {
```
rmisio: So, I ended up going back to making the `transactions` route not have the handle preface, which means we won't be showing it in the address bar, nor allowing it to be accessed via typing it's url into the address bar.

Having worked through it, if we did need to support `@handle/transactions` urls (both showing them in the address bar and giving the bar the ability to route to them), it wouldn't be that difficult to do (the infrastructure is in place).

rmisio:
@morebrownies Feel free to play around with this one (although, I know you're a busy man).

One open question, if someone attempts to navigate to a nonexistent tab on a page (e.g. via the address bar), should we:
a.) Route them to the default tab of that page.
b.) Route them to a generic Page Not Found page.
c.) something else

js/router.js
```diff
@@ -1,15 +1,153 @@
...
+      [/^@([^\/]+)[\/]?([^\/]*)[\/]?([^\/]*)[\/]?([^\/]*)$/, 'userViaHandle'],
+      [/^(Qm[a-zA-Z0-9]+)[\/]?([^\/]*)[\/]?([^\/]*)[\/]?([^\/]*)$/, 'user'],
+      ['transactions', 'transactions'],
+      ['transactions/:tab', 'transactions'],
+      ['*path', 'pageNotFound'],
     ];
 
-    routes.forEach((route) => this.route.apply(this, route));
+    routes.slice(0)
+      .reverse()
```
jjeffryes: Why is the array getting reversed?

js/start.js
```diff
@@ -1,11 +1,49 @@
...
 $('#pageNavContainer').append(pageNav.render().el);
 
+app.router = new ObRouter({
+  usersCl,
+  pageNavVw: pageNav,
+});
+Backbone.history.start();
+
+import { getGuid } from './utils';
+getGuid('sampatt');
```
jjeffryes: What does calling getGuid at the bottom of start.js do?

js/views/TemplateOnly.js
```diff
@@ -0,0 +1,27 @@
+// This is a view that will do nothing more than render
+// a template that it's given when contructed, passing along
+// any context passed into render as context for the template.
```
jjeffryes: This is a good idea for simple "pages".

js/views/TemplateOnly.js
```diff
@@ -0,0 +1,27 @@
...
+    super(options);
+
+    if (!options.template) {
+      throw new Error('Please provide a template');
+    }
+
+    this.options = options;
+  }
+
+  render(context = {}) {
```
jjeffryes: Isn't this setting the context to empty? Or is passing in the context not implemented yet?

styles/_base.scss
jjeffryes: I think base is a good place for the main page container styles. If they get too big we can break them off into another file.

js/router.js
rmisio: Because when we add them in this way using `route.apply`, they end up going in in reverse order. So, I'm just reversing the array so they go in the order we expect, which is how they're listed in code. Otherwise, our default pageNotFound route would be matched and wouldn't allow legitimate routes above it to be matched.

js/start.js
rmisio: This was for debugging purposes and I forgot to remove. I'll remove it.

js/views/TemplateOnly.js
rmisio: If you don't pass it in it defaults to empty, but if you pass it in, it will use whatever you pass in.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

js/router.js
```diff
@@ -1,15 +1,153 @@
...
+    tab = tab || 'inbound'; // eslint-disable-line no-param-reassign
+
+    this.loadPage(
+      new TransactionsPage({ tab }).render()
+    );
+  }
+
+  userNotFound() {
+    this.loadPage(
+      new TemplateOnly({ template: 'error-pages/userNotFound.html' }).render()
```
jjeffryes: Should this be TemplateOnly().render({ template: 'error-pages/userNotFound.html' })?

Or am I misunderstanding how context gets passed in?

rmisio: The template to use is an option of the View, not context that is to be passed to render. The context is actually what data you want to provide to the template. (and the template has no need to know it's own path :)

jjeffryes: Oh, right, it's going into loadTemplate wrappedTmpl. I didn't follow the trail far enough.

morebrownies:
@rmisio I tried getting the app running locally this morning and I've hit some issues. I'll check it out when I get a bit more time to focus on it. 

As for your question, I choose B:

b.) Route them to a generic Page Not Found page.

PR #11, opened 2016-07-15
Various additions

jjeffryes:
- add basic nav menu
- add structure to support sidebar
- stub in settings view


js/templates/settings.html
```diff
@@ -0,0 +1 @@
+<H1>Placeholder for Settings</H1>
```
rmisio: can we lowercase the `h1`?

jjeffryes: Sure.

styles/_base.scss
```diff
@@ -122,12 +122,31 @@ form {
...
     font-size: 30px;
   }
 }
+
+#chatContainer {
+  position: fixed;
+  top: 100px;
+  right: 0;
+  width: $chatClosed;
+  .chatOpen & {
```
rmisio: somewhat minor and i think a linter would help use here, but can't we have a new line before a nested selector (unless it's the first item)?

js/views/Settings.js
```diff
@@ -0,0 +1,43 @@
+import { View } from 'backbone';
+import loadTemplate from '../utils/loadTemplate';
+
+export default class Settings extends View {
```
rmisio: Sorry... I know you were just following my pattern, but I've realized we don't need to specify the class name (since we're not using it anywhere by name). So, the `Settings` could just be removed from this line.

js/views/PageNav.js
```diff
@@ -59,6 +61,10 @@ export default class PageNav extends View {
     }
   }
 
+  navListBtnClick() {
+    this.$('.js-navList').toggleClass('open');
```
rmisio: Is this just a dummy list that will be removed at some point?

If not, then I think we should cache the selector. The page nav will live for the life of the app and it's silly to keep going to the DOM for the same element.

js/templates/pageNav.html
```diff
@@ -32,20 +32,25 @@
...
-          <a class="discSm border colBorder"
+          <a class="discSm border colBorder js-navListBtn"
           <% if (ob.navimage) { %>
           //this will be updated to the real image parameter
           style="background-image: url(ob.navImage), url('../imgs/defaultUser.png')"
           <% } else { %>
             style="background-image: url('../imgs/defaultUser.png')"
           <% } %>
           ></a>
+          <nav class="box border boxList navList colBorder colPrimary padKids shad2 foldDown js-navList">
```
rmisio: Not saying this is right or wrong... just expressing as a fellow developer that it just seems like a whole lot of classes for one little drop down. I'm all for abstraction, but it's just confusing to try and understand what is a 'box'... what is a 'boxList'... difference between that and 'navList'... how 'foldDown' changes things.

It just would be so much easier to have a drop-down component (with 1 main class and possibly a couple of modifiers) and be able to go to that sass file and see exactly what's going on.

You definitely don't want to keep repeating your CSS and making overly bulky css, but is there some type of middle ground to this approach?

jjeffryes: In this case, this is a dummy list just so we can test out going to the page and making the open and close work. 

Once we get styles from Mike, we'll give it specific styling, it'll probably only have a few of the base classes.

js/views/PageNav.js
jjeffryes: Good point. This selector is real, so it should be cached in the render.

js/views/Settings.js
jjeffryes: Don't you need the class to extend the View?

styles/_base.scss
jjeffryes: Yes, normally that's the style I use, I just forgot it there.

js/views/Settings.js
rmisio: You don't need to give the class a name, since you're not referencing it anywhere. So, it could be:

```
export default class extends View {
```

js/views/PageNav.js
```diff
@@ -138,7 +137,10 @@ export default class PageNav extends View {
     });
 
     this.$addressBar = this.$('.js-addressBar');
+    this.$navList = this.$('.js-navList');
+    this.$popMenus = this.$('.js-navPopMenu');
 
+    getBody().on('click', (e) => { this.onBodyClick(e); });
```
rmisio: Why the extra wrapping function? Why not just `getBody().on('click', this.onBodyClick);`?

This shouldn't happen in render, otherwise if we re-render it's going to be bound again and we'll have duplicate handlers. If you do want it to be in render, then at least we should ensure we're removing any existing handlers before adding the new one (`.off(this.onClick)`).

Also, kind of minor, but I would just do this on document and not body. That's just usually where I've seen global handlers added, but if you really prefer body, no problem.

Also, this event handler would need to be removed in remove(). (I know that we're probably never going to remove this view, but IMHO, I still think it's good practice to have a view clean up after itself. Then, if for some reason we ever do have to remove it, we don't have to try and figure out all the cleanup things.)

jjeffryes: No reason on the wrapping function, it's a leftover from earlier when it had more code in it. 

Good point on doing it in render.

I originally had it on document, then I thought it seemed too much to add a getDoc to selectors.js. I'll put that back in. 

Thanks for the reminder on remove.

PR #12, opened 2016-07-15
Modals

rmisio:
This PR brings in the various modals from v1 and makes them available in v2.

To test, execute the following in your JS console:
```
location.hash = 'test-modals'
```

Additionally, to see the loading modal in action, use the address bar to navigate between different users, e.g.:
```
@themes
@sampatt
```


js/start.js
```diff
@@ -39,8 +39,20 @@ app.localSettings.fetch().fail(() => app.localSettings.save());
 const pageNav = new PageNav();
 $('#pageNavContainer').append(pageNav.render().el);
```
rmisio: If we find we're targeting `#pageNavContainer` or `#chatContainer` anywhere else, then we should add it to "selectors" module.

js/views/baseVw.js
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  /**
+   * Shortcut method to instantiate a view and register it as a child.
+   * @param {function} ChildView - The class of the child view (not instance).
+   * @param {...*} args - Remaining arguments will be passed into the constuctor
+   *   when instantiating the child view.
+   * @return {object} The create child view instance.
+   */
+  createChild(ChildView, ...args) {
```
rmisio: This is an addition to the baseVw and in one shot you can create and register a child.

So, instead of:

```
const myChildView = new MyChildView();
this.registerChild(myChildView);
```

you can now:

```
this.createChild(MyChildView);
```

js/views/modals/Test.js
```diff
@@ -0,0 +1,34 @@
...
+import loadTemplate from '../../utils/loadTemplate';
+import TestModal2 from './Test2';
+
+export default class extends BaseModal {
+  constructor(options = {}) {
+    super(options);
+    this.options = options;
+  }
+
+  className() {
```
rmisio: A couple of notable differences in extending from the baseModal (mainly due to some ES6 restrictions). When you declare either a className or events for your view which is extending from the BaseVw, make sure to declare them as functions (ES6 does not support class properties) and make sure to include the value from the base class. For example, notice how in `className` I append `super.className()` to my custom class -and- in events I merge `super.events()` with my own events object.

js/templates/modals/dialog.html
```diff
@@ -0,0 +1,17 @@
...
+<div class="<%= ob.messageClass %>">
+  <p><%= ob.message %></p>
+  <% if (ob.buttons.length) { %>
+  <div class="buttonStrip row">
+    <% ob.buttons.forEach((btn, i) => { %>
+    <%
+      // pending designs, may need to give certain buttons special classes depending
+      // on what position they are in (e.g. first, last...)
+    %>
+      <a class="btn colPrimary colBorder js-<%= btn.fragment %>" data-event-name="<%= btn.fragment %>"><%= btn.text %></a>
```
jjeffryes: The color classes need to be updated here to clrP and clrBr. I shortened the class names in the last PR.

styles/_base.scss
```diff
@@ -91,20 +104,18 @@ form {
...
+  position: relative;  
 }
 
 #chatContainer {
   position: fixed;
   top: 100px;
   right: 0;
   width: $chatClosed;
+  z-index: 1; // chat needs to be on top of modals
+  background-color: rgba(255, 255, 255, 0.6);
```
jjeffryes: I don't think chat has a background in the designs.

styles/_mixins.scss
```diff
@@ -12,4 +12,8 @@
     top: 50%;
     transform: translate(0, -50%);
   }
+}
+
+@mixin chatOpeningTransition($property) {
+  transition: $property 0.5s ease;
```
jjeffryes: Why make this a mixin? Do you think we'll use the same transition time and easing type more than once?

If we are going to re-use it, should it be called chatOpeningTransition?

jjeffryes: D'oh. And then I see it gets used on the page content. That makes sense.

styles/components/_modal.scss
```diff
@@ -0,0 +1,41 @@
...
+  left: 0;
+  right: 0;
+  bottom: 0;
+  background-color: rgba(0, 0, 0, 0.2);
+
+  .modalContent {
+    padding: 25px;
+    min-width: 300px;
+    max-height: 500px;
+    background-color: #fff;
```
jjeffryes: We don't want hard-coded colors, all colors should be set by color classes.

styles/_mixins.scss
rmisio: Yeah, the idea is that whatever needs to animate in response to the chat opening animation should use this mixin to ensure the duration and easing are the same. Otherwise the animations are not in sync. As of now, this is being used by Chat and then the page container and the modals. It's very possible some other areas will pop up that need to use it as well.

styles/_base.scss
rmisio: This was just to test to ensure that chat was coming above the modal overlay. I could remove it, but I think it's kind of useful to ensure that any code changes don't fudge that stacking order. Once we actually implement the chat, this could go away, since with the real chat it will be easy to see it's stacking is off.

PR #15, opened 2016-07-22
Settings template

jjeffryes:
- initial code for the settings modal
- faked model data
- stubbed save and cancel functions


js/views/PageNav.js
```diff
@@ -140,7 +141,8 @@ export default class extends View {
   }
 
   navSettingsClick() {
-    // activate settings modal here
+    new SettingsModal().render().open();
```
rmisio: I think we need to keep a reference to this modal and then check if the modal is already open before launching a new one, maybe:

```
if (!this.settingsModal || !this.settingsModal.isOpen()) {
  this.settingsModal = new SettingsModal().render().open();
}
```

To be honest, since there will be at least one other piece of code that launches the Settings modal (which would be keyboard shortcuts), I think we're going to want to move the launching of the modal into some decoupled module that any interested parties can call. But... I think for the time being it's ok for pageNav to manage it.

js/views/modals/Settings/Settings.js
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  selectTab(targ) {
+    const targTab = targ.data('tab');
+    const TabView = `Settings${targTab}`;
+
+    this.$('.js-tab').removeClass('clrT active');
+    targ.addClass('clrT active');
+
+    if (!this[TabView] && this.TabViews[TabView]) {
```
rmisio: Why is the purpose of the first check?

js/views/modals/Settings/Settings.js
```diff
@@ -0,0 +1,85 @@
...
+    this.currentTabView.save();
+  }
+
+  render() {
+    loadTemplate('modals/settings/settings.html', (t) => {
+      this.$el.html(t(this.options));
+
+      super.render();
+    });
+    this.tabContent = this.$('.js-tabContent');
```
rmisio: Can we put these inside of the loadTemplate callback? Right now loadTemplate works synchronously, but if we ever switched it to be async, then having them outside wouldn't work.

js/views/modals/Settings/SettingsGeneral.js
```diff
@@ -0,0 +1,92 @@
...
+    console.log('cancel general');
+    // cancel the form
+  }
+
+  render() {
+    loadTemplate('modals/settings/settingsGeneral.html', (t) => {
+      this.$el.html(t(this.options));
+    });
+
+    setTimeout(() => {
```
rmisio: This timeout wouldn't be necessary if you put the code inside of the `loadTemplate` callback.

test/views/pageNav.js
```diff
@@ -0,0 +1,52 @@
+import { expect } from 'chai';
```
rmisio: I know this is probably just a test stub, but it's confusing that it's a copy of the loadTemplate test. Can we remove it and just bring it in if and when we actually have a test for the view?

js/views/modals/Settings/SettingsPage.js
```diff
@@ -0,0 +1,72 @@
...
+      { code: 'USA', dataName: 'UNITED_STATES', name: 'United States' },
+      { code: 'DZD', dataName: 'ALGERIA', name: 'Algeria' },
+    ];
+
+    this.options = {
+      // test data
+      countryList: this.countryList,
+    };
+
+    this.options = Object.assign({}, options, this.profileModel, this.options);
```
rmisio: This is a little confusing because normally this.options on a view has to do with "options" related to the view. It seems like you're trying to combine data that you want to provide to the template..?

If that's the case, IMO, then the constructor is not the place to do it. Also, this.options doesn't need to contain data that's intended for the template. I think it's clearer for a view to have it's models separated from it's options, the latter mainly having to do with stuff that's passed in when instantiating the view.

So, if it is indeed a matter of providing the data to the template, this can be done in render:

```
this.$el.html(t({
  countryList: this.countryList,
  ...this.profileModel,
}));
```

(`...<object>` is es7 shorthand to say expand this object here (i.e. merge all of the object's properties here))

(I'm not passing in this.options into the template, because by looking at your genuine options, I don't think you're template needs that info)

and then in the constructor I think all you really need is `this.options = options` if you have a need to have access to any options that were passed in the the view was instantiated.

js/views/modals/Settings/Settings.js
jjeffryes: Code I didn't finish. This was supposed to only create a new view if it hadn't been already created.

js/views/modals/Settings/Settings.js
jjeffryes: Yep.

test/views/pageNav.js
jjeffryes: Yes. I forgot to finish that.

js/views/modals/Settings/SettingsPage.js
jjeffryes: It certainly doesn't need to be on options. Originally I was going to pass the data in as an option from the Settings parent, but didn't fix it after changing my mind.

In any case, it's just fake data until the models are wired into it.

js/views/PageNav.js
jjeffryes: That part is intended to be temporary, I just needed a way to trigger the settings modal.

js/views/modals/Settings/SettingsGeneral.js
```diff
@@ -0,0 +1,90 @@
...
+  render() {
+    loadTemplate('modals/settings/settingsGeneral.html', (t) => {
+      this.$el.html(t({
+        languageList: this.languageList,
+        countryList: this.countryList,
+        currencyList: this.currencyList,
+        ...this.userModel,
+      }));
+
+      setTimeout(() => {
```
jjeffryes: The timeout is required here, without it select2 doesn't pick up the width from the select element it's replacing.

js/views/modals/Settings/SettingsPage.js
```diff
@@ -0,0 +1,66 @@
...
+
+  render() {
+    loadTemplate('modals/settings/settingsPage.html', (t) => {
+      this.$el.html(t({
+        countryList: this.countryList,
+        ...this.profileModel,
+      }));
+    });
+
+    setTimeout(() => {
```
jjeffryes: The timeout is required here, without it select2 doesn't pick up the width from the select element it's replacing.

js/views/modals/Settings/SettingsGeneral.js
```diff
@@ -0,0 +1,90 @@
...
+      className: 'settingsGeneral',
+      events: {
+      },
+      ...options,
+    });
+
+    // temp data. This view will need the user model, the languages, the countries, and the
+    // currencies.
+
+    this.userModel = {
```
jjeffryes: This will be replaced by the real model when that's ready.

js/views/modals/Settings/SettingsPage.js
```diff
@@ -0,0 +1,66 @@
...
+  constructor(options = {}) {
+    super({
+      className: 'settingsGeneral',
+      events: {
+      },
+      ...options,
+    });
+  // temp data. This view will need the user model, the languages, the countries, and the
+    // currencies.
+    this.profileModel = {
```
jjeffryes: This will be replaced by the real model when that's ready.

js/views/modals/Settings/Settings.js
```diff
@@ -0,0 +1,87 @@
...
+    let newTabView;
+
+    this.$('.js-tab').removeClass('clrT active');
+    targ.addClass('clrT active');
+
+    if (this.TabViews[TabView]) {
+      newTabView = this[TabView] || new this.TabViews[TabView]();
+      if (this.currentTabView) this.currentTabView.remove();
+      this.currentTabView = newTabView;
+      this.tabContent.append(newTabView.render().el);
```
jjeffryes: I just remembered we talked about detaching these views instead of removing them. I'll add that in the next time I work on this.

js/views/PageNav.js
rmisio: It's 3 lines of code to ensure duplicate settings models aren't launched. Why not put it in with this merge?

js/views/modals/Settings/Settings.js
```diff
@@ -0,0 +1,87 @@
...
+    this.currentTabView.save();
+  }
+
+  render() {
+    loadTemplate('modals/settings/settings.html', (t) => {
+      this.$el.html(t(this.options));
+
+      super.render();
+
+      this.tabContent = this.$('.js-tabContent');
```
rmisio: Do we not want to follow the convention where variables referencing jQuery selectors are prefaced with a `$`?

js/views/modals/Settings/Settings.js
```diff
@@ -0,0 +1,87 @@
...
+    const opts = {
+      removeOnClose: true,
+      modelContentClass: 'modalContent clrP border clrBr',
+      ...options,
+    };
+
+    super(opts);
+    this.options = opts;
+
+    this.currentTabView = '';
```
rmisio: It's a bit odd that this will hold a view instance, but it's being initialized to an empty string. Does this really need to be initialized at all? `if (this.currentTabView) this.currentTabView.remove();` will work just the same if it's not initialized at all (since it will be undefined).

js/views/modals/Settings/Settings.js
```diff
@@ -0,0 +1,87 @@
...
+  selectTab(targ) {
+    const targTab = targ.data('tab');
+    const TabView = `Settings${targTab}`;
+    let newTabView;
+
+    this.$('.js-tab').removeClass('clrT active');
+    targ.addClass('clrT active');
+
+    if (this.TabViews[TabView]) {
+      newTabView = this[TabView] || new this.TabViews[TabView]();
```
rmisio: Can we pull out the `this[TabView]` check? Since we're never setting that anywhere, this will never be true. If we do decide to cache them at some point, then we could put it back in (although I think the instances should start with a lowercase letter. alternatively we could store them in an appropriately name variable, e.g. `this.tabViews` (lowercase t)).

js/views/modals/Settings/Settings.js
```diff
@@ -0,0 +1,87 @@
...
+
+  selectTab(targ) {
+    const targTab = targ.data('tab');
+    const TabView = `Settings${targTab}`;
+    let newTabView;
+
+    this.$('.js-tab').removeClass('clrT active');
+    targ.addClass('clrT active');
+
+    if (this.TabViews[TabView]) {
```
rmisio: Is this if check really necessary?

If the developer forgot to add a child view class into `this.TabViews` or if they put the wrong name in the html data attribute, wouldn't it be better to just let the compiler throw an exception on the next line (`this.TabViews[TabView]` will bomb)? Instead, what will happen now is that the target for the new tab will be active, but the contents will still be from the previous tab.

js/views/PageNav.js
jjeffryes: I didn't say I wouldn't. ;)

js/views/modals/Settings/Settings.js
jjeffryes: Yes, I just forgot to.

js/templates/pageNav.html
```diff
@@ -43,16 +43,27 @@
...
-            <a href="#transactions" class="listLink"><i class="ion-log-in"></i> Sales</a>
-            <a href="#transactions" class="listLink"><i class="ion-briefcase"></i> Cases</a>
-            <hr>
-            <a class="listLink js-navSettings"><i class="ion-toggle-filled"></i> Settings</a>
+            <div class="listGroup clrS clrBr">
+              <a class="listLink clrS clrBr">Current Node <i
+                class="ion-arrow-right-b floR"></i> </a>
+            </div>
+            <div class="listGroup clrS clrBr">
+              <a class="listLink">My Page<span class="floR clrT2">Cltrl + ?</span> </a>
```
rmisio: I'm not sure if floating is the best approach here. There was an issue in v1 where translations of the menu item text could get really long and when they break to multiple lines, the design doesn't look so nice:

image

Would it be better to a.) make it into a tabular layout so the link text and shortcut text are in separate columns b.) Make the width of the pop menu expandable where it can get wider if the content demands it (of course, along with a max-width so it doesn't get too wide)? The latter would limit, but not completely eliminate text breaking to multiple lines.

rmisio:
I'm also finding that hovering over radio's doesn't consistently update the cursor. It does if I'm over the outer part, but not if I'm on the actual radio or it's label. Maybe the hover css just applies to the .btn, but also needs to be explicitly declared on the input and label?

js/templates/pageNav.html
```diff
@@ -43,16 +43,45 @@
...
-            <a href="#transactions" class="listLink"><i class="ion-briefcase"></i> Cases</a>
-            <hr>
-            <a class="listLink js-navSettings"><i class="ion-toggle-filled"></i> Settings</a>
+            <div class="listGroup clrS clrBr">
+              <a class="listLink clrS clrBr">Current Node <i
+                class="ion-arrow-right-b floR"></i> </a>
+            </div>
+            <div class="listGroup clrS clrBr">
+              <a class="listLink">
+                <span>My Page</span><span class="clrT2">Cltrl + ?</span>
```
rmisio: Really nice work on supporting text that breaks to multiple lines:

image

One minor thing that I won't hold up the merge for is if we come across a really long single word. But, it's entirely possible we won't even come upon this case, so we could totally cross that bridge if we come to it.

image

js/views/modals/Settings/Settings.js
```diff
@@ -0,0 +1,88 @@
...
+    const tabViewName = targ.data('tab');
+    let tabView = this.tabViewCache[tabViewName];
+
+    if (!this.currentTabView || this.currentTabView !== tabView) {
+      this.$('.js-tab').removeClass('clrT active');
+      targ.addClass('clrT active');
+      if (this.currentTabView) this.currentTabView.$el.detach();
+      if (!tabView) {
+        tabView = new this.tabViews[tabViewName]();
+        tabView.render();
```
rmisio: You're never putting the tabView into the cache. Maybe after you create it you just need a `this.tabViewCache = tabView`.

jjeffryes: Looks like I cut that out when I refactored.

PR #18, opened 2016-07-23
Profile model

rmisio:
Implements the profile model. Test page available at `location.hash = 'test-profile'`.


js/models/Profile.js
```diff
@@ -0,0 +1,158 @@
...
+  get socialTypes() {
+    return [
+      'facebook',
+      'twitter',
+      'instagram',
+      'other',
+    ];
+  }
+
+  validate(attrs) {
```
rmisio: So basically if a field has an error we add an entry to an error object that we're returning from this method. The object is keyed by the field name and for each key there would be an array of errors (which might just be one). If a field has no errors, then it wouldn't be added to the object.

For example:

```
{
  'website': ['Please provide an url.'],
  'about': ['This must start with a lowercase q.', 'Some other error']
}
```

If we stay consistent with this approach, then this works hand in hand with our `formError.html` template which is available as template helper to all templates and is used to render the list of errors for a form field.

Anyhow, please see the example page of how to wire things together. It's just one approach, but I basically have the `model.validationError` contents passed into the template in `render`. That attribute is set by Backbone with the results of the most recent validation (e.g. if you call `model.save()` and client side validation fails, that attribute will contain the error object).

rmisio: I didn't add in all the validations for the Profile that we probably could have, but a lot of those validations become much more obvious once we start coding up the real form that drives the data. I basically did enough to give a good example.

I think anything that could fail on the server that the client can check, we should, as it saves an unnecessary server request and the latency / overhead involved with that.

js/start.js
```diff
@@ -51,15 +48,55 @@ app.simpleMessageModal.remove = () => {
...
+
+// get the server config
+$.get(app.getServerUrl('ob/config')).done((data) => {
+  app.profile = new Profile({ id: data.guid });
 
-app.loadingModal.close();
+  // todo: for now busting cache on fetch pending
+  // issue where my server is not running the latest
+  // code which solves this.
+  app.profile.fetch({ cache: false })
```
rmisio: This is mostly temporary code until we get our actual start / server connect flow into place.

js/templates/formError.html
```diff
@@ -0,0 +1,5 @@
+<ul class="unstyled errorList">
```
rmisio: The aforementioned `formError` template. The markup and css are most certainly tweakable.

styles/components/_formErrors.scss
```diff
@@ -0,0 +1,21 @@
+.errorList {
```
rmisio: I'm sure this could be improved. I imagine adjusting the markup in the `formError` template to use some the structure classes you created would make a bunch this css unnecessary. Feel free to adjust or point me in the right direction (I haven't spent much time playing with your classes as of yet).

styles/modules/_testProfile.scss
```diff
@@ -0,0 +1,67 @@
+.testProfilePage {
```
rmisio: This is just our profile test / dummy page, so for expediency's sake, I threw most css best practices out the window.

js/models/Profile.js
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  set(key, val, options) {
+    // Handle both `"key", value` and `{key: value}` -style arguments.
+    let attrs;
+    let opts = options;
+
+    if (typeof key === 'object') {
+      attrs = key;
+      opts = val;
```
jjeffryes: Wouldn't this overwrite any options set in opt?

js/models/Profile.js
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    if (attrs.social) {
+      if (!_.isEqual(attrs.social, this._social)) {
+        this.attributes.social = this._social;
+
+        // if the object contents are not equal, we'll clone a new object so
+        // the change is event is fired.
+        this._social = JSON.parse(JSON.stringify(attrs.social)); // deep clone
+        attrs.social = JSON.parse(JSON.stringify(attrs.social));
```
jjeffryes: Can the JSON.parse(JSON.stringify(attrs.social)) be done once here instead of twice?

js/models/Profile.js
```diff
@@ -0,0 +1,158 @@
...
+    if (typeof key === 'object') {
+      attrs = key;
+      opts = val;
+    } else {
+      (attrs = {})[key] = val;
+    }
+
+    if (attrs.social) {
+      if (!_.isEqual(attrs.social, this._social)) {
+        this.attributes.social = this._social;
```
jjeffryes: Where is attributes.social used?

js/models/Profile.js
```diff
@@ -0,0 +1,158 @@
...
+
+        if (is.not.string(item.type)) {
+          addError(`social[${index}].type`, 'Please provide a type.');
+        } else if (this.socialTypes.indexOf(item.type) === -1) {
+          addError(`social[${index}].type`, 'Type must be one of the required types.');
+        }
+
+        // if there are dupes of the same type, give an error to all
+        // dupes after the first one
+        if (groupedByType[item.type].length > 1 &&
```
jjeffryes: I think we're planning to allow more than one of type Other.

js/models/Profile.js
```diff
@@ -0,0 +1,158 @@
...
+    // to the server
+    if (method !== 'read') {
+      delete options.attrs.followerCount;
+      delete options.attrs.followingCount;
+      delete options.attrs.listingCount;
+      delete options.attrs.lastModified;
+    }
+
+    if (method === 'read') {
+      options.url = app.getServerUrl(`ipns/${model.id}/profile`);
```
jjeffryes: Would this read the user's own profile from ipfs?

js/models/Profile.js
rmisio: This is actually code taken right from backbone source. What it's doing is allowing the function to handle 2 different signartures 1.) `set(keyString, val, options)` 2.) `set(attrObject, options`).

So... if key is passed in as an object, then options would be passed in as the second arg (val) and the line you're looking at is just setting the opts to that.

js/models/Profile.js
rmisio: No, it's critical that the two variables reference completely different objects (on a deep level) and that wouldn't happen if we only "deep cloned" once. They'd both reference the same clone.

js/models/Profile.js
rmisio: Backbone uses it internally. One of it's uses is to compare against to determine if an attribute changed and consequentially if change events should be fired. It should usually not be directly updated (or very very judiciously if so). It was necessary to be updated in order for change event s to properly fire on the nested attribute.

js/models/Profile.js
rmisio: Yeah... good catch. We'll need to update the validation to allow that specific case.

js/models/Profile.js
rmisio: Yes, that's exactly what we want.

js/models/Profile.js
rmisio: FWIW, I still want to see if there's a nested attribute plugin that could help us. Unfortunately, the majority haven't been updated in a long time which gives me cause for concern. There is one though that's actively maintained that I still wanted to try out.

If not, then this code is fine, but I would want to abstract it so we wouldn't have to repeat it for every nested attribute.

js/models/Profile.js
rmisio: I take that back... it would read the profile of whatever ID we're passing into the model.

styles/components/_formErrors.scss
jjeffryes: We can wait for Mike to provide a design. Mostly you'd put some container styles on it.

styles/components/_formErrors.scss
```diff
@@ -0,0 +1,21 @@
...
+    margin-bottom: 5px;
+  }
+
+  margin-bottom: 10px;
+}
+
+label {
+  &.error {
+    input, textarea {
+      border: 1px solid red;
```
jjeffryes: This might get weird with the way radio buttons have the label inside the button. But we can fix that later after everything's merged. We'll need to get a design for errors from @morebrownies.

PR #19, opened 2016-07-27
Settings model + nested attributes

rmisio:
This change-set brings in an initial implementation of the Settings model. It also builds our an abstract way of handling nested model attributes.


js/views/TestProfilePage.js
```diff
@@ -9,6 +9,7 @@ export default class extends BaseVw {
     this.options = options;
 
     this.listenTo(app.profile, 'change', this.render, this);
+    this.listenTo(app.profile.get('social'), 'update', this.render, this);
```
rmisio: There are some bugs on this page having to do with adding and removing social accounts where some unsaved form data is blown away. I don't want to spend too much time on an example page, so I'm letting those be for now. The main purpose of the page was to show an example of validation as well as working with a model with nested attributes.

js/models/ShippingAddress.js
```diff
@@ -0,0 +1,16 @@
...
+  defaults() {
+    return {
+      Name: '',
+      Company: '',
+      AddressLineOne: '',
+      AddressLineTwo: '',
+      City: '',
+      State: '',
+      Country: '',
+      PostalCode: '',
```
jjeffryes: These keys are going to change. That's not an error, but we should discuss what we're going to use for the v2 address fields.

rmisio: I have a bug in the server to lowercase the keyname to be consistent with other APIs.

Yeah, we could discuss v2 address fields although I think some others should be included today (maybe at least wolf and cp?).

jjeffryes: I think Washington has some research on it Chris was going to implement. There's a shared doc with it someplace.

PR #20, opened 2016-07-29
Various tweaks

rmisio:
Multiple, mostly unrelated things:

1.) Removed the singular simpleMessageModal instance that was to be shared by the app. After some thought, I realized there are likely scenarios where more than one simpleMessage will be on the screen at a time. For example, your page displays a message, then a profile save API call which had been fired off a minute ago fails and we want to display a message informing the user of that, but we still want the other message to remain stacked below until the user has a chance to dismiss it.

2.) Added "close on remove" functionality to the base modal. It's off by default, but any modals that want to provide that functionality can just pass in true for that option in the constructor.

3.) By default, I'm having modals remove themselves when the router will route. This is an option that can be turned off if it's not appropriate for your modal instance. While removing a modal you launched in the remove() method of your view (or registering it as a child) works for most cases, there are edge cases where this is not sufficient (e.g. non-views are launching modals). Putting in this functionality should shore up those edge-cases.

4.) Some tweaks to the nested attributes functionality and the profile example page.

### DIFF
--- a/js/templates/testProfile.html
+++ b/js/templates/testProfile.html
@@ -23,10 +23,10 @@
   <label>
     Social Accounts:
   </label>
-  <div class="js-socialContainer">
+  <div>
   <% if (ob.social && ob.social.length) { %>
      <% ob.social.forEach((socialAcct, index) => { %>
-      <div class="socialAccount js-socialAccount">
+      <div class="socialAccount">
         <% if (ob.errors[`social[${index}].type`]) print(ob.formErrorTmpl({ errors: ob.errors[`social[${index}].type`] })) %>  
        <label>
           Type:
--- a/js/templates/testProfile.html
+++ b/js/templates/testProfile.html
@@ -23,29 +23,29 @@
   <label>
     Social Accounts:
   </label>
+  <div class="js-socialContainer">
   <% if (ob.social && ob.social.length) { %>
-       <div> <% // wrapping div needeed for JS to properly calc index %>
-       <% ob.social.forEach((socialAcct, index) => { %>
-         <div class="socialAccount">
-           <% if (ob.errors[`social[${index}].type`]) print(ob.formErrorTmpl({ errors: ob.errors[`social[${index}].type`] })) %>  
-           <label>
-              Type:
-             <select name="<%= `social[${index}].type` %>">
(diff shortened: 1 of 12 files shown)


js/templates/testProfile.html
```diff
@@ -23,33 +23,33 @@
   <label>
     Social Accounts:
   </label>
+  <div>
```
rmisio: These are mostly white-space changes. I did change this quite a bit, but ended up reverting it via copy and paste.

js/views/TestProfilePage.js
```diff
@@ -45,8 +69,15 @@ export default class extends BaseVw {
     social.remove(social.at(index));
   }
 
+  onRemoveSocial() {
+    // since the validation is index based and
+    // removing alters the indexes in the collection,
+    // we need to revalidate
```
rmisio: Yeah, the current approach with validation of a nested collection I took involves keeping indexes in the validation error keyname, e.g:

```
{
  social[0].type: ['an error'],
  social[1].type: ['an error', 'another error'],
}
```

Similarly, I'm using an index approach for the field name in the DOM. This is used when we scrape the form in order to extract the data from it to put in the model (similar to the JSON form plugin you showed me), e.g:

```
<div class="socialAccount">
  <input type="text" name="social[0].type" value="boom" />
</div>

<div class="socialAccount">
  <input type="text" name="social[1].type" value="bam" />
</div>
```

This ends up being problematic for a couple of reasons:

1.) Let's say you use jQuery to remove / add new items. This will leave the indexes out of sync and when you scrape the DOM, you end up with structure likes this (assuming you removed the first account above):

```
{[undefined, { type: "bam"}]}
```

2.) # 2 is similar to # 1, but with validation. Any time you remove an item from the collection, you'll need to re-validate the model because any errors in the `validationError` object may have incorrect indexes.

Anyhow, I think an approach where we could fix this issue is, instead of relying on index (which is something that changes for a model within a collection), we could use the model's cid as a way to map between the model and it's error / DOM field.

But... I need to let this feature breathe for a bit and then I'll circle back.

jjeffryes: We could assign the items ids of some kind instead. We can revisit it if it becomes an issue later.

PR #21, opened 2016-08-01
Startup flow

rmisio:
Re-worked the startup flow to determine if onboarding is needed and if so do a dummy onboarding for now (save the settings and profile models with default values).

### DIFF
--- a/js/router.js
+++ b/js/router.js
@@ -8,19 +8,12 @@ import TransactionsPage from './views/TransactionsPage';
 import TemplateOnly from './views/TemplateOnly';
 import TestModalsPage from './views/TestModalsPage';
 import TestProfilePage from './views/TestProfilePage';
+import Profile from './models/Profile';
 
 export default class ObRouter extends Router {
   constructor(options = {}) {
     super(options);
     this.options = options;
-
-    // Temporarily requiring a dummy 'users' collection. Eventually, we'll
-    // be getting this data via fetches to the server
-    if (!options.usersCl) {
-      throw new Error('Need me some dummy users!');
-    }
-
-    this.usersCl = options.usersCl;
 
     const routes = [
       [/^@([^\/]+)[\/]?([^\/]*)[\/]?([^\/]*)[\/]?([^\/]*)$/, 'userViaHandle'],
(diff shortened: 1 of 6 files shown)


js/router.js
```diff
@@ -121,17 +85,19 @@ export default class ObRouter extends Router {
       pageOpts.layer = args[1];
     }
 
-    this.getUser(guid).done((user) => {
+    const profile = new Profile({ id: guid });
+
+    profile.fetch().done(() => {
```
rmisio: I also eliminated the dummy user data collection, so the user page route is actually fetching from the server.

js/start.js
```diff
@@ -47,59 +34,192 @@ app.loadingModal = new LoadingModal({
...
+
+const onboardingNeededDeferred = $.Deferred();
+let profileFetch;
+let settingsFetch;
+let onboardProfile = false;
+let onboardSettings = false;
+let profileFailed;
+let settingsFailed;
+
+function isOnboardingNeeded() {
```
rmisio: If either the settings or profile are not set, we will onboard. While in 99% of cases, the lack of a profile would also indicate the lack of settings, there are some edge cases where one may have been saved but the other failed.

Anyhow, the idea is that those models are critical to the app and the user will not make it out of our start-up flow unless those are set.

rmisio:
If it looks like the app is hanging on the loading modal, please check your console. It is likely a server API request is pending, which semi regularly take a long time.

rmisio:
Another issue is that sometimes the initial profile save responds with a IPNS failure. Initial saves should be sent via POST and since the initial save failed, the assumption is that we still need to POST. But, if it's IPNS failure, then I guess it did save to the DB, so when we retry we need to PUT.

It's all very confusing and I'm working with Chris to have the save succeed if it is saved locally regardless of whether there's an IPNS error or not.

Anyhow, if you stumble upon this scenario, when you retry the POST, it will come back with an error that PUT needs to be used. You're only recourse would be to wipe your data -or- temporarily update your source to use a PUT on the save.

I know, very confusing, I'm working on getting the changes we need to make it not confusing. Worst comes to worst, if we can't decouple the IPNS errors from our local ones, we'll just have to break REST convention and just only accept POSTs (at least for any endpoints that publish to IPFS).

PR #22, opened 2016-08-02
Lang support

rmisio:


js/start.js
```diff
@@ -1,18 +1,60 @@
...
+
+  return 'en-US';
+}
+
+const initialLang = getValidLanguage(app.localSettings.get('language'));
+app.localSettings.set('language', initialLang, { supressLangChangeWarning: true });
+app.polyglot = new Polyglot();
+app.polyglot.extend(require(`./languages/${initialLang}.json`));
+
+app.localSettings.on('change:language', () => {
```
rmisio: So the idea is that if we are beyond onboarding, any language changes will kick up a dialog telling the user they need to re-start before the changes will fully take effect. This allows us to simplify things where almost all of our views don't need to do anything to update themselves with a new lang (which in some cases was tricky).

If we're not beyond onboarding, then we will re-render the page-nav (so we need to be mindful to ensure page-nav is re-renderable (which in general is a good practice for all views)). Also, the expectation is the onboarding view will re-render itself based on any language changes that happen via it's own interface.

I did consider simplifying things and just putting up the dialog on all lang changes, but the problem is that we could theoretically have a client that toggles between servers with different languages and in that case we might have a re-start for a server change, followed by a restart for lang change on every switch.

js/data/languages.js
```diff
@@ -0,0 +1,121 @@
...
+  },
+  {
+    name: 'Uzbek (Uzbek)',
+    code: 'uz',
+  },
+  {
+    name: 'Klingon (Klingon)',
+    code: 'tlh',
+  },
+];
```
jjeffryes: We should probably start with just 2 languages, English and a placeholder (like Spanish), and add more as translations are created.

We don't know, for example, if we'll get a Uzbek translation at all, since we have to wait for volunteers to do the translation.

js/languages/en-US.json
```diff
@@ -0,0 +1,447 @@
+{
+  "restartNow": "Restart Now",
+  "restartLater": "Restart Later",
+  "langChangeRestartTitle": "Restart needed for language change",
+  "langChangeRestartMessage": "In order for your language change to fully take effect, you must restart the app. If any saves or publishes are still in progress, please wait until they are done.",    
+  "timezones": {
```
jjeffryes: We can get rid of timezones, we won't be using them for anything.

js/data/languages.js
rmisio: I commented out all except english and polish (since I already was using polish as my placeholder lang).

PR #24, opened 2016-08-05
Add polyglot template Helper from the userPage branch.

jjeffryes:


js/utils/templateHelpers.js
```diff
@@ -1,5 +1,10 @@
-// todo: add in some helpers!
+import app from '../app';
+
+const polyT = function (...args) {
+  return app.polyglot.t(...args);
+};
+
+export {
```
rmisio: Can we directly export the polyT function?

```
export function plyT(...args) {
}
```

In ES6 modules, you can export multiple things individually, for example:

```
export function a() {};

export function b() {};

const variableA = 'meatballs';

export { variableA };
```

The benefit is that then individual items can be imported and only those would go into memory. If I'm not mistaken, if you just export everything in one big object, I think the whole thing gets brought in.

jjeffryes: LOL. That's how I had it before, I changed it to this based on our conversation earlier about switching to this pattern.

js/utils/templateHelpers.js
```diff
@@ -1,10 +1,6 @@
...
 
-const polyT = function (...args) {
-  return app.polyglot.t(...args);
-};
-
-export {
-  polyT,
-};
+export function polyT(...args) {
+  return app.polyglot.t(args);
```
rmisio: I think you need the ellipses for the args your passing to app.polyglot.t as well.

js/utils/templateHelpers.js
rmisio: well, we must have been playing 'whos on first?'. What's new, right?

js/utils/templateHelpers.js
jjeffryes: Rolled back to the wrong state. I'll update it.

PR #25, opened 2016-08-05
Save settings

rmisio:
This PR brings in the functionality to save the Settings > General and Settings > Page tabs to their respective APIs. It also brings in the currency, language and country reference data modules.

To see some client validation in action, set the Page tab website field to an invalid url or the email field to an invalid email.

To see the 'close modal at own risk' warning if closing while a save is in progress, you're probably going to need to throttle your network speed (in the Network tab).


test/data/countries.js
```diff
@@ -0,0 +1,14 @@
+import { expect } from 'chai';
```
rmisio: I could definitely add more tests for this module since it also includes a function to return translations, but it depends on app.polyglot and I'm not quite sure how to stub that in. Maybe something like Sinon can help with that. I suppose I may just be able to add Polyglot to the app module inside of the test... but I'm not sure what the best approach is right now, so I'll have to circle back around to this.

js/data/countries.js
```diff
@@ -0,0 +1,1529 @@
...
+  return getIndexedCountries()[dataName];
+}
+
+function getTranslatedCountries(lang, sort = true) {
+  if (!lang) {
+    throw new Error('Please provide the language the translated couuntries' +
+      ' should be returned in.');
+  }
+
+  let translated = [...countries];
```
jjeffryes: Why not start with the let translated below, since you're immediately reassigning it?

rmisio: because sort, sort's the array in place and I don't want to modify the main list.

js/data/currencies.js
```diff
@@ -0,0 +1,660 @@
...
+
+export default currencies;
+
+function getTranslatedCurrencies(lang, sort = true) {
+  if (!lang) {
+    throw new Error('Please provide the language the translated currencies' +
+      ' should be returned in.');
+  }
+
+  let translated = [...currencies];
```
jjeffryes: Same question here with reassigning translated.

js/models/Profile.js
```diff
@@ -84,17 +54,23 @@ export default class extends BaseModel {
...
+      addError('email', app.polyglot.t('profileModelErrors.provideValidEmail'));
     }
 
     if (attrs.website && is.not.url(attrs.website)) {
-      addError('website', 'Please provide a valid url.');
+      addError('website', app.polyglot.t('profileModelErrors.provideValidURL'));
+    }
+
+    if (attrs.handle && attrs.handle.charAt(0) === '@') {
+      addError('handle', 'The handle should not start with a leading hash.');
```
jjeffryes: Don't forget to translate this phase. You probably want "@" instead of hash in the error.

styles/components/_buttons.scss
```diff
@@ -61,6 +61,25 @@
...
+      position: absolute;
+      background-color: rgba(0,0,0,0.6); // use variablized color when refactoring
+      color: white; // use variablized color when refactoring
+      text-align: center;
+      top: 0;
+      left: 0;
+      right: 0;
+      bottom: 0;
+      font-size: 14px;
+      cursor: default;
```
jjeffryes: Add border-radius: inherit so it'll pick up the border radius of the button. 

Otherwise the corners poke out.

js/models/Profile.js
rmisio: There's a few that I purposely left untranslated if they're ones that will never reasonably make it into the UI. I don't want to clutter up the translations file in that case.

For example, in the settings form, I'm having the view strip out any leading @, so the user will never see this error.

Same goes for the leading hash on the color field.

Anyhow, if you still think it's worth it to translate it, even though it's very likely not make it into the UI, I could do so..?

(I will update the handle error to reference '@' instead of hash.)

js/models/Profile.js
```diff
@@ -118,7 +94,8 @@ export default class extends BaseModel {
       // dupes after the first one
       if (socialAttrs.type !== 'other' && groupedByType[socialAttrs.type].length > 1 &&
         groupedByType[socialAttrs.type].indexOf(socialMd) > 0) {
-        addError(`social[${index}].type`, 'You already have a social account of this type.');
```
rmisio: This one I do need to translate. I thought it wouldn't make it into the UI, but it very well can.

styles/components/_buttons.scss
rmisio: hmm... I'm not overwriting border-radius, so why wouldn't it just use the value set by the button?

(fwiw, this is totally dummy styling. I'm expecting this to be properly revamped when the design for this come in)

js/models/Profile.js
jjeffryes: I think any text that could make it into the UI should be translated. Both just in case it does get shown, and so it's not confusing to translators. 

Some of them do go through the code, they could wonder why some phrases are untranslatable.

styles/components/_buttons.scss
rmisio: ok, i think i see your point. Yeah, i'll confirm, if it is indeed sticking out, I'll fix.

jjeffryes: Setting it to inherit does use the value set by the button. 

But if it's just a placeholder you can leave it as is.

rmisio: Putting it on the pseudo `after` would inherit from the button?

rmisio: ok, i'll just leave it as is for now and we can make it awesome when the designs come in.

jjeffryes: Yup. Border-radius doesn't normally inherit, but if you put border-radius: inherit on the after selector the injected content will inherit it from the wrapping parent element.

js/views/modals/Settings/Settings.js
```diff
@@ -55,20 +64,90 @@ export default class extends BaseModal {
...
+          .render()
+          .open();
+        }
+      })
+      .done(() => {
+        if (this.confirmNavAwayDialog && this.confirmNavAwayDialog.isOpen()) {
+          this.$saveStatus.text(app.polyglot.t('settings.statusSafeToClose'));
+        }
+
+        $('#statusBarContainer').text(app.polyglot.t('settings.statusSaveComplete'));
```
jjeffryes: The statusBarContainer could be cached, since it shows up several times.

js/views/modals/Settings/Settings.js
```diff
@@ -55,20 +64,90 @@ export default class extends BaseModal {
...
+  close(skipWarning) {
+    if (!skipWarning && this.saving) {
+      this.confirmNavAwayDialog = new Dialog({
+        title: app.polyglot.t('settings.confirmNavAwayWarning.title'),
+        message: app.polyglot.t('settings.confirmNavAwayWarning.message'),
+        buttons: [{
+          text: app.polyglot.t('settings.btnYes'),
+          fragment: 'yes',
+        }, {
+          text: app.polyglot.t('settings.btnYes'),
```
jjeffryes: This should be btnNo, currently both buttons say "Yes".

test/data/countries.js
jjeffryes: Shouldn't there be a test for the currencies data file too?

rmisio: Yes, when I solve the issues outline in my comment above.

js/views/modals/Settings/Settings.js
rmisio: I purposely left it like this because this is a very temporary status bar. When I bring in the v1 status bar, we won't even be targeting it this way with a selector.

js/data/countries.js
rmisio: ah... i see your point. Will have a fix up soon.

js/models/Profile.js
rmisio: ok, but it is, IMO adding unneeded bloat to the translation file. An example is there are 2 validations on the social account type: a.) Please provide a type b.) Type must be one of the required types.

Since the UI represents the field as a drop down and it populates the drop down from the values stored in the model, there's no practical way the user will ever see these errors. The only way there would be would be is if, for example, we allowed the user to update this value from some other form and the developer didn't represent it as a drop down, in which case I think it's a dev error that would need to be fixed.

styles/components/_buttons.scss
rmisio: added in the border-radius.

PR #27, opened 2016-08-10
Status bar

rmisio:
Ports over the Status Bar view from v1. To see it in action, save your settings via the Settings modal (you probably want to throttle your connection to see the intermediary states).

### DIFF
--- a/styles/_base.scss
+++ b/styles/_base.scss
@@ -118,6 +118,7 @@ ul, ol {
 }
 
 // index containers and sections
+
 #pageNavContainer {
   position: relative;
   z-index: 2;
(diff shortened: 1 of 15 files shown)


styles/_base.scss
```diff
@@ -146,22 +147,18 @@ ul, ol {
...
-
-  &::after {
-    content: '(temporary Walmart of status bars)';
-    position: absolute;
-    right: 5px;
-    top: 5px;
-  }  
+// turn off animations and transitions with this style
+
+#ov1.minEffects,
```
rmisio: I initially needed this for one of the status bar tests... and then I switched approaches and didn't need it anymore. So, we're not using this CSS as of now, but since I think we will, I left it in. If you don't think we're going to carry over this functionality into v2, let me know and I could pull it out.

jjeffryes: We'll still need it, some Linux machines and old or low powered machines need the effects turned off to run well.

PR #28, opened 2016-08-11
User page - About

jjeffryes:
- builds out the basic userPage view 
- changes the index structure to keep the page content aligned to the tabs when the chat sidebar is open or closed, but allow it to move to the right if the window is too small
- the home tab view and template  
- stubs for store, following, followers views/templates
- styling completed for the store search bar and controls
- stubs for the home page message, follow, and moderator controls, with template logic 
- the userShort template, which will be used with the following, followers, channels and search (the userShort view still needs to be created, the template is used on its own in the userPage view)


index.html
```diff
@@ -12,29 +12,31 @@
...
-            </div>
-            <div>
-              <a class="discSm border clrBr js-navListBtn" style="background-image: url('../imgs/defaultUser.png')"></a>
-            </div>
-            <div>
-              <a class="discSm border clrBr js-navListBtn" style="background-image: url('../imgs/defaultUser.png')"></a>
-            </div>
+        <div id="chatLeftSpacer"></div>
+        <div id="pageContainer"></div>
+        <div id="chatRightSpacer"></div>
```
jjeffryes: The spacers allow some flex magic to work for the positioning of the pageContainer. The left spacer will collapse if the window is too small, allowing the pageContainer element to move to the left. 

If the window isn't too small, the pageContainer will align with the left edge of the tabs.

js/router.js
```diff
@@ -83,6 +86,8 @@ export default class ObRouter extends Router {
     if (tab === 'channel') {
       pageOpts.category = args[0];
       pageOpts.layer = args[1];
+    } else {
+      pageOpts.action = args[0];
```
jjeffryes: This will be used in the future for actions like customizing the page that are triggered from a nav link.

js/router.js
```diff
@@ -121,6 +128,12 @@ export default class ObRouter extends Router {
     });
   }
 
+  ownPage(tab, ...args) {
+    tab = [tab || 'about'];   // eslint-disable-line no-param-reassign
```
jjeffryes: The ownPage route always goes to the user's own page.

js/templates/pageNav.html
```diff
@@ -42,43 +38,45 @@
...
+              </a>
             </div>
-            <div class="listGroup clrS clrBr">
-              <a class="listLink">
+            <div class="listGroup clrP clrBr clrSh2">
+              <a class="listItem" href="#ownPage">
                 <span>My Page</span><span class="clrT2">Cltrl + ?</span>
               </a>
-              <a class="listLink">
+              <a class="listItem" href="#ownPage/about/customize">
```
jjeffryes: An example of a route with an action for the userPage.

js/templates/pageNav.html
```diff
@@ -42,43 +38,45 @@
...
                 <span>Customize Page</span><span class="clrT2">Cltrl + ?</span>
               </a>
-              <a class="listLink">
+              <a class="listItem" href="#ownPage/store/newListing">
```
jjeffryes: We're probably going to replace this with a js hook to open the edit listing modal.

js/templates/testProfile.html
```diff
@@ -1,4 +1,12 @@
 <form>
+  <% if (ob.errors.name) print(ob.formErrorTmpl({ errors: ob.errors.name })) %>
+  <label class="<% if (ob.errors.name) print('error') %>">
+    Name:
+    <input type="test" name="name" value="<%= ob.name %>" />
+  <% if (ob.errors.handle) print(ob.formErrorTmpl({ errors: ob.errors.handle })) %>
+  <label class="<% if (ob.errors.handle) print('error') %>">
+    Handle:
+    <input type="test" name="handle" value="<%= ob.handle %>" />
```
jjeffryes: Fields were added to the test view just to get some data for testing.

js/templates/transactionsPage.html
```diff
@@ -1,9 +1,33 @@
...
+      <div class="txt5b">Stars(x) <%= ob.polyT('userPage.followsYou') %></div>
+    </div>
+    <div class="flexExpand">
+      <div class="flexHRight flexVCent gutterH clrT2">
+        <a class="btn tab clrBr js-tab" data-tab="Home">Purchases</a>
+        <a class="btn tab clrBr js-tab" data-tab="Store">Sales</a>
+        <a class="btn tab clrBr js-tab" data-tab="Following">Cases</a>
+      </div>
+    </div>
+  </div>
```
jjeffryes: The userPage and transactions templates both use the same tab structure. It might be worth breaking it off into it's own template.

js/templates/userPage/userPage.html
```diff
@@ -0,0 +1,58 @@
...
+            <a class="btn tab clrBr js-tab" data-tab="Home">Home</a>
+            <a class="btn tab clrBr js-tab" data-tab="Store">Store</a>
+            <a class="btn tab clrBr js-tab" data-tab="Following">Following</a>
+            <a class="btn tab clrBr js-tab" data-tab="Followers">Followers</a>
+          </div>
+        </div>
+      </div>
+
+</nav>
+<div class="banner" <% if(ob.banner){ %>
```
jjeffryes: The banner has a fixed position behind the rest of the page. This lets it be positioned under the chat sidebar. 

It does mean the banner image doesn't scroll with the rest of the page. @morebrownies how do you feel about that?

js/templates/userPage/userPage.html
```diff
@@ -0,0 +1,58 @@
...
+  style="background-image: url('<%= ob.header %>, url('../imgs/defaultBanner.png')"
+  <% } else { %>
+  style="background-image: url('../imgs/defaultBanner.png')"
```
jjeffryes: I put in a placeholder for the default banner. When @morebrownies has one ready we can swap it in.

js/templates/userShort.html
```diff
@@ -0,0 +1,33 @@
+<div class="contentBox clrBr clrP clrSh2 userShort">
```
jjeffryes: This template will be used anywhere the small box with user info appears, such as follow, following, search results, and channels.

js/views/userPage/UserPage.js
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  moreClick() {
+    this.$moreableBtns.toggleClass('hide');
+  }
+
+  selectTab(targ) {
+    let tabViewName = targ.data('tab');
+    let tabView = this.tabViewCache[tabViewName];
+    const tabViewType = tabViewName; // the original view name is passed in for the Follow view
```
jjeffryes: Since the following and follower tabs appear to be identical other than their titles, the plan here is to pass in a flag for the title and use the same view/template for both.

js/views/userPage/UserPageHome.js
```diff
@@ -0,0 +1,77 @@
...
+      }));
+
+      this.userShort = this.$('.js-userShort');
+
+      // add the userShort sub-template
+      loadTemplate('userShort.html', (u) => {
+        this.userShort.html(u({
+          ...app.profile.toJSON(),
+        }));
+      });
```
jjeffryes: The userShort template is added in here.

main.js
```diff
@@ -14,7 +14,7 @@ function createWindow() {
   mainWindow = new BrowserWindow({
     width: 1200,
     height: 760,
-    minWidth: 1024,
+    minWidth: 1170,
```
jjeffryes: I changed the minimum width to match the smallest size where the chat can be open and still have room for the userPage.

js/templates/pageNav.html
rmisio: yeah, since we decided on a modal, it's completely decoupled from the user page.

js/templates/userPage/userPage.html
rmisio: Should it really be doing the if on `ob.banner` and then putting in `ob.header`?

Also, isn't `ob.header` going to be a hash, rather than a full url?

In either case, this is something that was a template helper in v1 and I think we should follow suit in v2. IMHO, it's so much cleaner to call the helper with a few args than have to repeat this type of if / then structure in templates (I've already seen this in some other v2 template).

js/views/PageNav.js
```diff
@@ -4,7 +4,7 @@ import { View } from 'backbone';
 import loadTemplate from '../utils/loadTemplate';
 import app from '../app';
 import $ from 'jquery';
-import SettingsModal from './modals/settings/Settings';
+import SettingsModal from './modals/Settings/Settings';
```
rmisio: I wouldn't think this folder should be capitalized. Maybe we can change it in the file structure? Do you have strong feelings that it should be capitalized?

morebrownies: Do I need my server running to see this?

rmisio:
Can we get a little spacing to the right of the search icon?

image

js/views/PageNav.js
jjeffryes: Re: the Settings folder, it's capitalized in master, when I changed it to lower case (here, and changed the folder name to lower case) Travis threw an error.

jjeffryes: @morebrownies yes, you'll need the server running to see this.

js/templates/userPage/userPage.html
jjeffryes: The ob.header part is a placeholder for when we have the code in place to get images. We can certainly make it a template helper.

rmisio:
So, before as the chat opened/closed, the page would adjust it's size by animating. Now, it's jumping in a herky jerky fashion. Can we put a animated transition back in on that?

js/templates/pageNav.html
```diff
@@ -42,43 +38,45 @@
...
-            <div class="listGroup clrS clrBr">
-              <a class="listLink js-navSettings">
+            <div class="listGroup clrP clrBr clrSh2">
+              <a class="listItem js-navSettings">
                 <span>Settings</span><span class="clrT2">Cltrl + ?</span>
               </a>
             </div>
-            <div class="listGroup clrS clrBr">
-              <a href="#transactions" class="listLink">
+            <div class="listGroup clrP clrBr clrSh2">
```
rmisio: for About and Support, until they're ready, can we remove any href from them? It's a bit confusing to have them go to the transactions page.

js/languages/en-US.json
```diff
@@ -51,13 +73,6 @@
     "provideUsername": "Please provide a username.",
     "duplicateSocialAccount": "You already have a social account of this type."
   },
-  "profileModelErrors": {
```
rmisio: Why are these deleted? These are still needed, since these are ones that can make it into the UI.

js/models/Profile.js
```diff
@@ -6,6 +6,10 @@ import SocialAccounts from '../collections/SocialAccounts';
 export default class extends BaseModel {
   defaults() {
     return {
+      name: `ob ${Math.random().toString(36).slice(2)}`,
```
rmisio: What's going on with this name? Was this just for debugging?

js/templates/userPage/userPageHome.html
```diff
@@ -0,0 +1,81 @@
+<div class="flexRow gutterH">
+  <div class="col4 gutterVMd">
+    <div class="js-userShort">
+      <!-- the userShort template is injected here-->
+    </div>
+    <% if(ob.isModerator) { %>
```
rmisio: If the model has a moderator flag, why do we need isModerator. Is there some business logic to determine that on top of what is stored in the server?

js/templates/userPage/userPageHome.html
```diff
@@ -0,0 +1,81 @@
...
+      </div>
+      <div class="listItem">
+        <div><%= ob.polyT('userPage.website') %></div>
+        <div class="clrT2"><a href="<%= ob.website %>"><%= ob.website %></a></div>
+      </div>
+      <div class="listItem">
+        <div><%= ob.polyT('userPage.email') %></div>
+        <div class="clrT2"><%= ob.email %></div>
+      </div>
+      <div class="listItem">
```
rmisio: I imagine the social accounts declared like this is just placeholder code... because I think we will want to switch it to build off of the nested social collection within Profile model.

In fact, if we want we could even make a separate little view for these guys and then just re-render on the `update` event of the collection. That way if someone had modified their settings while this page was in the background, it would be nice and updated for them without having to refresh the page.

js/views/PageNav.js
```diff
@@ -171,12 +173,15 @@ export default class extends View {
 
   render() {
     loadTemplate('pageNav.html', (t) => {
-      this.$el.html(t());
+      this.$el.html(t({
+        // guid: app.profile.id,
```
rmisio: can we remove the debugging code?

rmisio:
(this is kind of why I wanted to leave the white bg on the chat app for now, so we could easily spot this)

The requirement is that chat and the page nav are always available. In this PR, the modals are now overlaying chat (and making it inaccessible).

styles/modules/modals/_loading.scss
```diff
@@ -14,6 +14,8 @@
 
     .loadingText {
       @include center;
+      width: 100%;
```
rmisio: Why are these two style additions necessary, the mixin already centers the text vertically and horizontally?

styles/components/_layout.scss
```diff
@@ -244,12 +281,30 @@
   margin-left: $pad;
 }
 
+@mixin colHGutter($colPad) {
```
rmisio: We do have a separate mixins file? Is the reasoning for putting this one here that it really would only be used here?

js/languages/en-US.json
jjeffryes: I'm not sure how those got deleted. It might have been merge shenanigans.

js/models/Profile.js
jjeffryes: If you don't provide a name, we should auto-generate a random one for you, the same as in v1.

js/templates/userPage/userPageHome.html
jjeffryes: That's true. This should just use what's in the model.

js/languages/en-US.json
jjeffryes: Weird. In my local branch, this isn't deleted.

js/views/userPage/UserPage.js
```diff
@@ -0,0 +1,113 @@
...
+    this.$tabTitle.text(tabViewName);
+
+    if (tabViewName === 'Followers' || tabViewName === 'Following') tabViewName = 'Follow';
+
+    if (!this.currentTabView || this.currentTabView !== tabView) {
+      this.$('.js-tab').removeClass('clrT active');
+      targ.addClass('clrT active');
+      if (this.currentTabView) this.currentTabView.$el.detach();
+      if (!tabView) {
+        tabView = new this.tabViews[tabViewName]({ tabViewType, ownPage: this.ownPage });
```
rmisio: Could you use createChild here from the baseVw so that the new view is registered as a child and therefore removed when the parent view is removed?

```
tabView = this.createChild(this.tabViews[tabViewName], { tabViewType, ownPage: this.ownPage });
```

js/views/userPage/UserPage.js
```diff
@@ -0,0 +1,113 @@
...
+  constructor(options = {}) {
+    super(options);
+    this.options = options;
+
+    this.tabViewCache = {};
+    this.tabViews = { Home, Store, Follow };
+
+    this.followed = false; // TODO check to see if user is followed by the viewer
+    this.followsYou = true; // TODO check to see if this user follows the viewer
+    this.ownPage = true; // TODO check to see if this is the viewer's page
```
rmisio: It's a matter of saying `this.model.id === app.profile.id`.

js/views/userPage/UserPage.js
```diff
@@ -0,0 +1,113 @@
...
+      }
+      this.$tabContent.append(tabView.$el);
+      this.currentTabView = tabView;
+    }
+  }
+
+  render() {
+    loadTemplate('userPage/userPage.html', (t) => {
+      this.$el.html(t({
+        ...app.profile.toJSON(),
```
rmisio: I don't think that we're going to want to be sending app.profile in here, but rather this.model, which might be the same as app.profile, but only if we're on our own page.

js/views/userPage/UserPageFollow.js
```diff
@@ -0,0 +1,24 @@
...
+      events: {
+      },
+      ...options,
+    });
+  }
+
+  render() {
+    loadTemplate('userPage/userPageFollow.html', (t) => {
+      this.$el.html(t({
+        ...app.profile.toJSON(),
```
rmisio: see comment above.

js/templates/userPage/userPageHome.html
jjeffryes: Right. Once we have the code to put the social data into settings, we can make this a loop that creates whatever ones are present. 

If someone updates their social networks, won't they probably also change other profile data, which would require a re-render anyway?

js/views/userPage/UserPageHome.js
```diff
@@ -0,0 +1,77 @@
...
+import { View } from 'backbone';
+import app from '../../app';
+
+export default class extends View {
+  constructor(options = {}) {
+    super(options);
+    this.options = options;
+
+    this.ownPage = options.ownPage;
+    this.isModerator = true; // TODO set if this user is a moderator
```
rmisio: If we're going to give the subviews this.model, they could get isModerator right from the model.

js/views/userPage/UserPageHome.js
```diff
@@ -0,0 +1,77 @@
...
+    return {
+      'click .js-termsLink': 'termsClick',
+      'click .js-termsClose': 'termsClose',
+      'click .js-addModerator': 'addModeratorClick',
+    };
+  }
+
+  termsClick() {
+    this.$termsDisplay.toggleClass('open');
+    if (this.$termsDisplay.hasClass('open')) this.$termsDisplay[0].scrollIntoViewIfNeeded();
```
rmisio: +1 `scrollIntoViewIfNeeded` is a really cool method the browser folks have given us.

js/views/userPage/UserPageHome.js
```diff
@@ -0,0 +1,77 @@
...
+      this.currentModerator = true;
+      this.$addModeratorLbl.addClass('hide');
+      this.$removeModeratorLbl.removeClass('hide');
+    }
+  }
+
+  render() {
+    loadTemplate('userPage/userPageHome.html', (t) => {
+      this.$el.html(t({
+        ...app.profile.toJSON(),
```
rmisio: same as comment above, wrong model.

styles/modules/modals/_loading.scss
jjeffryes: They're not. They were necessary for a while when the page structure was different, I forgot to take them out when I changed it.

js/views/userPage/UserPageStore.js
```diff
@@ -0,0 +1,24 @@
...
+      events: {
+      },
+      ...options,
+    });
+  }
+
+  render() {
+    loadTemplate('userPage/userPageStore.html', (t) => {
+      this.$el.html(t({
+        ...app.profile.toJSON(),
```
rmisio: same thing as above.

styles/components/_layout.scss
jjeffryes: Yes. This mixin should never be used anywhere else, which is why it's in this file.

js/models/Profile.js
rmisio: Ok, sounds good. There is also an underscore unique id generator you might be interested in:
http://underscorejs.org/#uniqueId

What's here is probably fine, so I'm just sharing with you mostly as an FYI.

js/templates/transactionsPage.html
rmisio: Yeah, I'm seeing a lot of the same code in the views as well to manage the tabs. Sounds like a TabbedPage View and template might be something good to make.

Interestingly enough, we're re-using the same JS from the Settings modal which has a completely different look than this. It may be that the modals and pages share a common TabbedPageView and just implement different templates.

js/templates/userPage/userPageHome.html
rmisio: Yeah, that's true and if the re-render is not super smooth (i.e. too much state is lost), then we may want to instead just put a message somewhere on the page that says a refresh is needed. My reasoning is that the Save could happen after they've closed the modal and they may be interacting with the page and a forced re-render might be to jarring.

That's what github does:
image

js/languages/en-US.json
jjeffryes: If you look in the raw file, they haven't been deleted.

js/templates/pageNav.html
```diff
@@ -42,43 +38,45 @@
...
+              </a>
             </div>
-            <div class="listGroup clrS clrBr">
-              <a class="listLink">
+            <div class="listGroup clrP clrBr clrSh2">
+              <a class="listItem" href="#ownPage">
                 <span>My Page</span><span class="clrT2">Cltrl + ?</span>
               </a>
-              <a class="listLink">
+              <a class="listItem"><!--TODO add js action for Page Customization-->
```
rmisio: Let's circle back and talk about this one. I know it may sound like I'm talking about both sides of my mouth here, but I don't think we want a JS action for this, otherwise we'd have to have both the page-nav and router manage opening the user page, which IMO is not worth it at all.

If the idea is just to keep customize out of history, I think from the pageNav we could probably link to 'ownPage/customize' and just have the user route re-route to '<guid>/about' while giving a 'customize' flag to the userPage view.

For clicks of the Customize CTA within the userPage, we don't even need to go through the router. There we could just go into the customize flow via JS since we wouldn't want to lose other page state (like the tab they're on). That's how it's done in v1 and it seems to work.

Sorry for any confusion.

PR #29, opened 2016-08-11
Own page route

rmisio:
(DO NOT MERGE INTO MASTER!!!!!!!!! If you find this suitable, please merge it into the userPage branch, since this branch was created off of that)

Anyhow, this re-works the ownPage route so that it will just redirect to the user route, albeit with the user owns guid. Any and all other path components will carry over, so as we build out the user pages routes, we never need to also adjust the ownPage route.


rmisio:
closed in liu of: https://github.com/OpenBazaar/openbazaar-desktop/pull/30

PR #32, opened 2016-08-12
Publish status

rmisio:
This code branch wires in a Status Bar message to indicate publishing state based on the 'puiblishing' socket info relayed by the server. It also shows a confirmation dialog if you attempt to close the app during publishing (or if the last publish failed).

To see the status messages in action, save the Page tab of the Settings modal.

*** one gotcha is that the close-confirm-during-publish dialog will not show up on OSX when you click the X icon. This is because that is only hiding our app rather than closing it. Since the app is still active, the publishing process can continue. I confirmed with @morebrownies that this behavior is OK. FWIW, if on OSX you close the app in other ways (e.g. right-click on app icon and close, then you will get the confirm dialog).


js/start.js
```diff
@@ -253,27 +256,167 @@ function onboardIfNeeded() {
...
-  // update the app.settings model which will in turn update our local
-  // settings model. You shouldn't be directly updating the language in our local
-  // settings model.
-  app.settings.on('change:language', (settingsMd, lang) => {
-    app.localSettings.save('language', getValidLanguage(lang));
-  });
+ // let's start our flow - do we need onboarding?,
+ // fetching app wide models...
+function start() {
+  fetchConfig().done((data) => {
```
rmisio: Start.js is, IMO, getting a little too bulky. When I implement the real start-up / connection flow, I'll look into separating some things out.

js/start.js
```diff
@@ -253,27 +256,167 @@ function onboardIfNeeded() {
...
+  } else {
+    location.reload();
+  }
+});
+
+app.apiSocket.on('close', () => {
+  if (lostSocketConnectionDialog) return;
+
+  lostSocketConnectionDialog = new Dialog({
+    title: 'Socket connection failure',
```
rmisio: Didn't translate because this dialog is very temporary. Once we have the connection modal in place, I think we would just want to show that upon socket failures (at least that's what we do in v1).

PR #34, opened 2016-08-18
Settings address

rmisio:
This PR adds in the Address tab to settings.

@morebrownies I found that overlaying every delete button on a delete / add address looked kind of weird. I also found that we want to prevent them from updating the form, since we clear it on success, if they started to update it, they would just be interrupted. So I just overlay the whole tab during saves:

image

Obviously, we could improve on that styling! Also, I only show the tab overlay once they move their mouse. This prevents it from flickering on / off very quickly on super fast saves (which most are. In fact I think you'll have a hard time even getting the overlay to show - I had to muck with the server to delay the response)


morebrownies:
Cool thanks for the update. That seems like a good work around.  I'll check
it out once I get my server going.

js/languages/en-US.json
```diff
@@ -55,13 +56,25 @@
...
     "textColor": "Text Color",
+    "addresses": {
+      "name": "Name",
+      "company": "Company",
+      "street": "Street",
+      "apt": "Apt/Unit/Suite",
+      "city": "City",
+      "state": "State/Province/Region",
+      "country": "Country",
+      "postalCode": "Postal Code",
```
jjeffryes: This should probably be Zip/Postal Code, so American users aren't confused.

js/templates/modals/settings/addressesForm.html
```diff
@@ -0,0 +1,83 @@
+<form class="box border padKids padStack clrS clrBr js-addAddressForm">
+  <div class="flexRow">
+    <div class="col3">
+      <label for="settingsAddressName"><%= ob.polyT('settings.addresses.name') %></label>
+    </div>
+    <div class="col9">
+      <% if (ob.errors.name) print(ob.formErrorTmpl({ errors: ob.errors.name })) %>
+      <input type="text" class="clrBr" name="name" for="settingsAddressName" value="<%= ob.name %>" />
```
jjeffryes: Why do the inputs have for attributes? That's not a valid attribute for an Input element, should it be an Id?

js/templates/modals/settings/addressesForm.html
```diff
@@ -0,0 +1,83 @@
...
+      <input type="text" class="clrBr" name="postalCode" for="settingsAddressPostalCode" value="<%= ob.postalCode %>" />
+    </div>
+  </div>
+  <div class="flexRow">
+    <div class="col3">
+      <label for="settingsAddressCountry"><%= ob.polyT('settings.addresses.country') %></label>
+    </div>
+    <div class="col9">
+      <% if (ob.errors.country) print(ob.formErrorTmpl({ errors: ob.errors.country })) %>
+      <input type="text" class="clrBr" name="country" for="settingsAddressCountry" value="<%= ob.country %>" />
```
jjeffryes: Shouldn't country be a dropdown, like in General?

js/templates/modals/settings/addressesForm.html
jjeffryes: Each input should have a clrP class, so it has a primary color background.

js/templates/modals/settings/addressesForm.html
```diff
@@ -0,0 +1,83 @@
+<form class="box border padKids padStack clrS clrBr js-addAddressForm">
```
jjeffryes: Looks like some of the styling is off. The tabColumn element should have clrP and a border on the right (with clrBr), settingsTabFormWrapper should have clrS and the form should have clrP.

js/templates/modals/settings/addressesList.html
```diff
@@ -0,0 +1,23 @@
...
+        addressIndex += 1;
+    %>
+        <div class="addressBox border clrS clrBr">
+          <%= `${address.name}${address.company ? '<br />' : ''}` %>
+          <%= `${address.company}${address.addressLineOne ? '<br />' : ''}` %>
+          <%= `${address.addressLineOne}${address.addressLineTwo ? ` ${address.addressLineTwo}` : ''}${address.city ? '<br />' : ''}` %>
+          <%= `${address.city}${address.state ? `, ${address.state}` : ''}${address.postalCode ? ` ${address.postalCode}` : ''}${address.country ? '<br />' : ''}` %>
+          <%= `${address.country}${address.addressNotes ? '<br />' : ''}` %>
+          <% if (address.addressNotes) print(`<p class="notes">${address.addressNotes}</p>`) %>
+          <a class="btn clrP clrBr js-delete btnDelete" data-address-index="<%= addressIndex %>">Delete</a>
```
jjeffryes: It doesn't look like the btnDelete class is used, can it be removed?

styles/_mixins.scss
```diff
@@ -22,3 +22,25 @@
   pointer-events: none;
   opacity: 0.2;
 }
+
+// todo: refactor this style once we have designs
+@mixin processingOverlay {
+  position: relative;
```
jjeffryes: The overlay doesn't always cover the element it's overlaying, height: 100% might fix that.

image

styles/_mixins.scss
```diff
@@ -22,3 +22,25 @@
...
+
+  &::after {
+    content: 'Processing...';
+    position: absolute;
+    background-color: rgba(0,0,0,0.6);
+    color: white;
+    display: flex;
```
jjeffryes: Why is this display flex?

styles/_mixins.scss
jjeffryes: Though if this does have height 100%, the content message may not always be visible, depending on where the scrollbar is.

js/languages/en-US.json
rmisio: I was just going off of the designs. @morebrownies what do you think of adding in 'Zip' to the label?

js/templates/modals/settings/addressesForm.html
rmisio: I think when the first two tabs were created they were done off of now-outdated designs. I just copied the markup over from them and didn't retrofit the stylings with the thought that @morebrownies could do such tweaks.

styles/_mixins.scss
rmisio: yeah, just spent an hour on this and I just can't get that darn overlay to extend. If you want to play around and see if you can come up with a way, let me know.

Otherwise, I think another option is to disable the scroll bar. Since in css there's no way to disable a scroll bar without hiding it (which would mean a page jump), I'd need to use some JS to get the scrollbar width and apply a margin to compensate for the hidden scroll bar. But... don't think that would be too difficult.

morebrownies:
I'll catch up on the design questions this afternoon.

jjeffryes:
I'll merge this, another PR takes care of the style issues. We can change the zip/postal code text later if we want to.

PR #35, opened 2016-08-18
Build and Deployment Travis 

hoffmabc:


Gruntfile.js
```diff
@@ -0,0 +1,58 @@
+const path = require('path');
```
rmisio: for consistency with the rest of the app, can we use ES6 syntax for imports?

```
import path from 'path';
```

hoffmabc: NO NO NO

rmisio: i'm curious if this will work, because i'm not sure if our grunt file is being transpiled by Babel. If it doesn't work, don't worry about it. The gruntfile could be an exception.

rmisio:
@hoffmabc gosh, how about a commit that doesn't break Travis!

hoffmabc:
OK @jjeffryes this bad boy is ready to merge.

PR #37, opened 2016-08-23
Changes the settings form colors to match the current design.

jjeffryes:
@morebrownies I updated the settings form to match the colors in Zeplin, take a look and make sure this is right.

@rmisio if these are correct, can you make the same changes in the address PR?


rmisio:
Was this one merged with the other settings merges today and we can close this?

If not, then there are some conflicts to fix.

jjeffryes:
All of these changes have already been merged in.

PR #40, opened 2016-08-24
Make modal height flexible

jjeffryes:
Makes the modal heights change with the window height.

### DIFF
--- a/js/templates/modals/settings/addresses.html
+++ b/js/templates/modals/settings/addresses.html
@@ -1,7 +1,7 @@
 <div class="bar padMd clrBr borderBottom">
   <h2 class="h3 txUnb"><%= ob.polyT('settings.addressesTab') %></h2>
 </div>
-<div class="scrollBox settingsTabFormWrapper js-addressesWrap clrS">
+<div class="settingsTabFormWrapper js-addressesWrap clrS">
   <div class="padMd settingsTabFormWrapperInner">
     <div class="js-listContainer"></div>
     <div class="js-formContainer"></div>
--- a/js/templates/modals/settings/general.html
+++ b/js/templates/modals/settings/general.html
@@ -1,7 +1,7 @@
 <div class="bar padMd clrBr borderBottom">
   <h2 class="h3 txUnb"><%= ob.polyT('settings.generalTab') %></h2>
 </div>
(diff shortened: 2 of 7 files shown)


rmisio:
There's a couple issues with this approach:

1.)  I don't think all modal should have this type of flexible height. For example, here's one that shouldn't:

image

I'm thinking maybe we make the default the old way and then make the flexible heigh opt in via a class (it could be vice versa as well).

2.) I would think even on flexible height modals, we'd still want some max height. Otherwise if someone has a really large monitor, the modal is always going to take up 90% (or whatever) of the screen space and there would potentially be a lot of white space on the modal.

=====

Another thing I wanted to point out is with this approach, if the modal content is larger than the viewport could support (large modal and/or small monitor), we're not going to get the scrollbars on the overlay (like Mike's fiddle example showed). Instead, each modal will still need to be able to have an inner scroll bar to account for this case. For example:

image

=====

I think we need to think through this approach. Because the idea was that this was supposed to be an easy bridge between fixed height modals and the real solution Mike wants. But, it's fraught with problem. I think we need to think about whether it's better to think more about this approach and patch the patch, revert back to the previous way (fixed height modals) or bite the bullet and take the scope hit involved with transitioning to the real way Mike wants.

@morebrownies what do you think? Maybe you want to play around with this branch and see what you think?

jjeffryes:
This has been redone to align modals with the "modalTop" class to the top of the window, below the navBar, and set their height to expand with their content. The settings modal has a minimum height, we may want to move that to apply to all modalTop modals. 

The maximum height is set to fill the area under the navBar, minus a margin at the top and bottom. If the modal content is bigger than that, an internal scrollbar will appear in the modal.

This PR doesn't change the settings save button, it's still a single button for all the tabs, but it is now positioned at the bottom of the content, so it will need to be scrolled to if the content is larger than the modal maximum height.

jjeffryes:
@morebrownies take a look at this when you have a minute, and see what you think.

morebrownies:
@jjeffryes Is there a specific modal I can trigger to see the height change?

jjeffryes:
@morebrownies open the settings modal, and switch between the general, page and address tabs. Each one has a different height.

PR #41, opened 2016-08-24
Npm build script

rmisio:
Created an npm build script (runnable via `npm run build`) which will among other little tweaks output the transpiled JS files to the `dist` folder.


jjeffryes:
@hoffmabc can you test this and see if it works with your builds?

PR #43, opened 2016-08-26
Packaging server daemon + client

hoffmabc:


main.js
```diff
@@ -1,11 +1,359 @@
-import { app, BrowserWindow, ipcMain } from 'electron';
+import { app, BrowserWindow, ipcMain, Menu, Tray, remote, ps } from 'electron';
+let os = require('os');
```
rmisio: Yeah, you're really going to hate our linter... but it means well :)

hoffmabc:
Nope it didn't flag it, but I can change those as well. WHY NOT?

main.js
```diff
@@ -1,10 +1,9 @@
-import { app, BrowserWindow, ipcMain, Menu, Tray } from 'electron';
+import { electron, app, BrowserWindow, ipcMain, Menu, Tray } from 'electron';
```
rmisio: if this is what you had before:

```
const electron = require('electron');
```

I'm pretty sure you want this here, since electron is the default export of that module:

```
import electron, { app, BrowserWindow, ipcMain, Menu, Tray } from 'electron';
```

But, if the way you have works (it doesn't bomb when you reference electron in the code), then it's ok.

rmisio:
There's some build failures, but I don't think it's tests or linting that's failing, so I imagine we could ignore those. Is that right?

main.js
```diff
@@ -1,11 +1,372 @@
...
+    sub.on('close', closecallback);
+    sub.unref();
+  } else {
+    mainWindow.webContents.executeJavaScript("console.log('Unable " +
+      "to find openbazaard')");
+  }
+};
+startLocalServer();
+
+// let killLocalServer = function () {
```
rmisio: I'm wondering if we should be commenting out unused pieces of code or omitting them altogether (we could always get them from v1 later, right).

But, I guess it's not so much, so I think it's OK for now. I would just want to limit it should it ever get to be a situation where the file is more commented out stuff than not.

hoffmabc: I'll just go ahead and fix this before we merge. I'm in no hurry.

rmisio:
@hoffmabc I foolishly merge this without testing it (only reviewed the code) and it's broken:

image

Anyhow, I've reverted master. I don't think there's a way to revive a closed PR, so you'll need to make a new one. Sorry.

hoffmabc:
Ok no worries yeah it's not ready.

PR #46, opened 2016-09-01
Follow

jjeffryes:
This PR adds the follow functionality to the User Page.
- the UsersShort collection and UserShort model can be used for any simple list of users consisting of just guids. 
-- If an array of guids is passed into the UsersShort instead of an array of objects, it will parse them into objects. 
-- The UsersShort collection requires a type and a guid, the URL is set by both. This allows it to call the user's server or ipns based on whether the guid belongs to the user, and to call any endpoint to get the list of guids. 

- in start.js, the ownFollowing and ownFollowers collections are fetched (using UsersShorts). They are stored in app.js, and made available to the entire application.
-- the socket in start.js listens for follow and unfollow messages, and updates the ownFollowing and ownFollowers collections when they happen.

- app.js has a followUnfollow function that can be called by any view to follow or unfollow a guid.

- The UserPage dynamically shows the followed state of the user if they are not the viewing user.
-- the UserPage has a "this user follows you" message if the node being viewed followers the viewer.
-- the UserPage follow button will show "follow" or "unfollow" based on the current state of whether the viewer is following that node.
-- the view listens to ownFollowing and ownFollowers, and updates the follow button and the follows you messages when they change.
-- clicking the button will follow or unfollow the node based on whether it is currently being followed.

- The Follow view is used for both the Following and Followers tabs
-- The Follow view will re-render when the collection it is using updates (this will probably be replaced by something more robust when lazy loading is added).
-- currently the view loads all the users in the collection. This will be updated with lazy loading in a future PR.

- The UserShort view is used for each user 'card'. 
-- the view renders with no data and a loading message so it can occupy a spot in the DOM, and then loads the profile ID it was given. When the profile loads, the view re-renders with that data. If the load fails, it re-renders with an error.
-- the view shows a follow/unfollow button if it is not the viewer's own card.
-- the follow/unfollow button updates it's state by listening to the ownFollowing collection. This isn't used in this PR, since all views where it would apply remove or rerender the userShort views when the ownFollowing collection updates. It will be used in views like search and channels.
-- a button to add/remove the user as a moderator has been stubbed in
-- if the user's name or handle is clicked on, the app will navigate to their page
-- the rating stars have been stubbed in


js/app.js
```diff
@@ -12,4 +15,40 @@ export default {
...
+        this.ownFollowing.add({ guid });
+      } else {
+        this.ownFollowing.remove(guid); // remove via id
+      }
+    })
+    .fail((data) => {
+      const followFailedDialog = new Dialog({   // eslint-disable-line no-unused-vars
+        title: this.polyglot.t('errors.badResult'),
+        message: data.responseJSON.reason,
+        dismissOnOverlayClick: true,
```
rmisio: These last three are all true by default, so no need to pass them in.

js/app.js
```diff
@@ -12,4 +15,40 @@ export default {
   getSocketUrl() {
     return 'ws://localhost:8080/ws';
   },
+
+  followUnfollow(guid, type = 'follow') {
```
rmisio: I know we tentatively decided to put this function in app, but after some thought, I think this would be better in it's own utility module (be it some "follow" specific one or a more generic "request" one where other non-model / collection server calls can go).

App was mainly designed to be something to throw instances on that are needed app wide (e.g. the router), more or less a glorified 'window'. I want to avoid it becoming what it turned into in v1, where it seemed to become a dumping ground for anything that could be needed in many parts of the app. Most of that stuff would be better served in a more semantically named utility module.

Another reason it's dangerous to put stuff in app is because app is used so prevalently that it's easy to get into circular dependency issues.

I know it's a very blurry line on what should go in here and I'm not always 100% clear myself, but I tend to err on keeping it very lean.

Just my two cents.

js/app.js
```diff
@@ -12,4 +15,40 @@ export default {
...
+    .done(() => {
+      // if the call succeeds, add or remove the guid from the collection
+      if (type === 'follow') {
+        this.ownFollowing.add({ guid });
+      } else {
+        this.ownFollowing.remove(guid); // remove via id
+      }
+    })
+    .fail((data) => {
+      const followFailedDialog = new Dialog({   // eslint-disable-line no-unused-vars
```
rmisio: if you just remove the variable assignment (`const followFailedDialog`), you won't need the eslint override.

js/app.js
```diff
@@ -12,4 +15,40 @@ export default {
...
+        title: this.polyglot.t('errors.badResult'),
+        message: data.responseJSON.reason,
+        dismissOnOverlayClick: true,
+        dismissOnEscPress: true,
+        showCloseButton: true,
+      })
+        .render()
```
rmisio: You don't need a new line before that render.

js/collections/UsersShort.js
```diff
@@ -0,0 +1,31 @@
...
+module.exports = Collection.extend({
+  /* we have to use the older style for this collection, the ES6 style creates a bug where models
+  cannot be removed using their ids */
+
+  initialize: function (models, options) {  // eslint-disable-line object-shorthand
+    if (!options.type) {
+      throw new Error('You must provide a type to the collection');
+    }
+
+    this.url = app.getServerUrl(options.guid === app.profile.id || !options.guid ?
```
rmisio: Can we declare this as a 'url' function so it's evaluated every time the url is needed as opposed to on object instantiation?

js/collections/UsersShort.js
```diff
@@ -0,0 +1,31 @@
...
+      `ob/${options.type}` : `ipns/${options.guid}/${options.type}`);
+  },
+
+  model: UserShort,
+
+  parse: function (response) {   // eslint-disable-line object-shorthand
+    response = response.map((guid) => {   // eslint-disable-line no-param-reassign
```
rmisio: If instead of saying `response = ...`, you just returned the results of the map call, then you wouldn't need the 'no-param-reassign' override.

js/router.js
```diff
@@ -78,7 +78,7 @@ export default class ObRouter extends Router {
   }
 
   user(guid, tab, ...args) {
-    tab = tab || 'store'; // eslint-disable-line no-param-reassign
+    tab = tab || 'Store'; // eslint-disable-line no-param-reassign
```
rmisio: The tab value is taken from the route and we don't want it capitalized there (e.g. we don't want @themes/Store), so I don't think we should be capitalizing the tab here.

I know the user view needs it capitalized because it needs to map to a certain data structure in the view, but that is very much an implementation detail of the user view. Can we just have the userView capitalize it when it gets it?

js/start.js
```diff
@@ -104,18 +105,26 @@ function fetchConfig() {
...
 let settingsFailed;
+let ownFollowersFailed;
+let ownFollowingFailed;
 
 function isOnboardingNeeded() {
   profileFetch = !profileFetch || profileFailed ?
     app.profile.fetch() : profileFetch;
   settingsFetch = !settingsFetch || settingsFailed ?
     app.settings.fetch() : settingsFetch;
+  ownFollowersFetch = !ownFollowersFetch || ownFollowersFailed ?
```
rmisio: My Bad!

I totally wrongly advised you on where to put this. This shouldn't be here, because obtaining the followers has absolutely nothing to do with determining if onboarding is needed.

Would you mind reverting this and breaking the following fetches into their own function? It could be named something "follow" specific or some generic "fetchStartupData". It should only be called if it's determined onboarding is not needed, maybe augment the 'onboardIfNeeded' function, e.g:

```
function onboardIfNeeded() {
  isOnboardingNeeded().done((onboardingNeeded) => {
    if (onboardingNeeded) {
      // let's go onboard
      onboard().done(() => onboardIfNeededDeferred.resolve());
    } else {
        fetchStartupData().done(() => onboardIfNeededDeferred.resolve());  // updated this line, rest is as-is from start.js      
    }
  });

  return onboardIfNeededDeferred.promise();
}
```

Really sorry for the misdirection.

js/start.js
```diff
@@ -362,32 +378,40 @@ function setPublishingStatus(msg) {
...
+        });
+
+        unpublishedContent = false;
+
+        publishingStatusMsgRemoveTimer = setTimeout(() => {
+          publishingStatusMsg.remove();
+          publishingStatusMsg = null;
+        }, 2000);
+      }
+    } else if (e.jsonData.notification) {
```
rmisio: These other socket events are publish specific and this block is part of the `// manage publishing sockets` section. Can we move the follow block into it's own message handler (`app.apiSocket.on('message', (e) => {`...) and put it next to the other follow functionality (e.g. new function you create above)? I think the idea is to keep specific bits of functionality together, otherwise things get confusing if we interweave different blocks together.

I am hoping to break up start.js a bit into some sub modules and keeping the functionality as sand boxed as possible would be helpful when I get to that.

js/views/userPage/UserPage.js
```diff
@@ -5,18 +5,43 @@ import app from '../../app';
...
     this.tabViewCache = {};
-    this.tabViews = { Home, Store, Follow };
+    this.tabViews = { Home, Store, Follow, Reputation };
 
-    this.followed = false; // TODO check to see if user is followed by the viewer
-    this.followsYou = true; // TODO check to see if this user follows the viewer
     this.ownPage = this.model.id === app.profile.id;
+
+    if (!this.ownPage) {
+      this.followed = app.ownFollowing.where({ guid: this.model.id }).length > 0;
```
rmisio: Since the userShort model has it's idAttribute set to the guid field, it's indexed by that field, so `app.ownFollowing.get(<guid>)' would be much more efficient than where which has to go through the entire list until it finds the user you're looking for.

Can you also make the switch to the other place you're using 'where' in this way.

js/views/userPage/UserPage.js
```diff
@@ -65,25 +78,34 @@ export default class extends BaseVw {
...
 
   selectTab(targ) {
-    let tabViewName = targ.data('tab');
-    let tabView = this.tabViewCache[tabViewName];
-    const tabViewType = tabViewName; // the original view name is passed in for the Follow view
-
-    this.$tabTitle.text(tabViewName);
+    // if an invalid targ is passed in, set it to Store
+    if (!this.tabViews[targ] && targ !== 'Following' && targ !== 'Followers') {
+      targ = 'Store'; // eslint-disable-line no-param-reassign
```
rmisio: I know it's a pain, but we can't keep just disabling a rule. What the rule wants you to do in this case, is to assign it to a function level var and then modify the var, e.g:

```
let tabTarg = targ;  // modify tabTarg at will
```

Otherwise, we should just disable this rule in its entirety if we're just going to override it everywhere. I'm a little hesitant to do that, since the rule is valued by some and apparently mutating function arguments is a bad practice that can lead to unwanted consequences. (but i have read at least one blog where the dude disabled it).

But, let's choose a path here.

js/views/userPage/UserPageFollow.js
```diff
@@ -1,23 +1,56 @@
...
-    loadTemplate('userPage/userPageFollow.html', (t) => {
-      this.$el.html(t({
-        ...this.model.toJSON(),
-      }));
-    });
+  className() {
+    return 'userPageFollow flexRow';
+  }
 
+  folRender() {
```
rmisio: Why do we have a 'folRender' function and then an empty 'render' function? Can't this just be in render?

js/app.js
jjeffryes: I'm moving it to a followUtils.js file in utils, I think there could be other functions that can go in there that will be useful.

js/app.js
jjeffryes: My usual style with chained calls was to put them all on new lines so it's easy to see what's happening. 

Do we want to always put them on the same line instead?

js/app.js
rmisio: Isn't putting 'Utils' in the filename a bit redundant since it's in the util folder?

js/app.js
rmisio: I think it's a personal preference thing. I guess it's fine.

js/collections/UsersShort.js
```diff
@@ -0,0 +1,31 @@
...
+import { Collection } from 'backbone';
+import UserShort from '../models/UserShort';
+import app from '../app';
+
+
+module.exports = Collection.extend({
+  /* we have to use the older style for this collection, the ES6 style creates a bug where models
+  cannot be removed using their ids */
+
+  initialize: function (models, options) {  // eslint-disable-line object-shorthand
```
rmisio: I think you could use function short-hand here (and therefore get rid of the lint override). I think it was the ES6 class behavior that was messing up the collection. I don't think using function short-hand on the object being passed to 'extend' would cause problems.

js/views/userPage/UserPage.js
```diff
@@ -65,25 +78,34 @@ export default class extends BaseVw {
...
-    if (tabViewName === 'Followers' || tabViewName === 'Following') tabViewName = 'Follow';
+    let tabView = this.tabViewCache[targ];
+    const tabOptions = { ownPage: this.ownPage, model: this.model };
 
     if (!this.currentTabView || this.currentTabView !== tabView) {
+      this.$tabTitle.text(targ);
+      // add tab to history
+      app.router.navigate(`${this.model.id}/${targ}`, {
+        trigger: false,
+        replace: false,
```
rmisio: I believe trigger and replace are false by default, so no need to pass them in.

Although, this brings up a UX question (maybe for Mike), do we want each tab to be a separate item in history?

js/app.js
jjeffryes: I find having multiple files named the same thing but in different folders confusing.

rmisio: I've just never seen for example a string util file called 'util/stringUtils.js' or DOM utilities called 'util/domUtils.js'.

But, if you feel strongly about it, I could totally live with it.

jjeffryes: It's not a great name, I admit, I couldn't think of something better to describe "does things related to following, but isn't called follow.js".

js/views/userPage/UserPageFollow.js
jjeffryes: This is a leftover, originally I had a template that was rendered and the folRender added to it.

js/start.js
```diff
@@ -394,6 +425,18 @@ app.apiSocket.on('message', (e) => {
 
 let unpublishedConfirm;
 
+app.apiSocket.on('message', (e) => {
+  if (e.jsonData) {
+    if (e.jsonData.notification) {
+      if (e.jsonData.notification.follow) {
+        app.ownFollowers.add({ guid: e.jsonData.notification.follow });
```
rmisio: When someone is followed, in the DB, they are added to the top off the list. I would think we want to mimic that. I think Collection.add() will add the model to the end of the list, so we probably want to utilize Collection.unshift() here.

js/utils/follow.js
```diff
@@ -0,0 +1,49 @@
+import $ from 'jquery';
+import app from '../app';
+import Dialog from '../views/modals/Dialog';
+
+export default {
```
rmisio: We can also export each function individually as opposed to exporting one big object with nested functions. It's arguably a cleaner syntax that saves you a level of indentation. But, I suppose it's a personal preference thing as well.

js/utils/follow.js
```diff
@@ -0,0 +1,49 @@
...
+    return $.ajax({
+      type: 'POST',
+      url: app.getServerUrl(`ob/${type}`),
+      data: JSON.stringify({ id: guid }),
+      dataType: 'json',
+    })
+      .done(() => {
+        // if the call succeeds, add or remove the guid from the collection
+        if (type === 'follow') {
+          app.ownFollowing.add({ guid });
```
rmisio: Same thing as my comment about the ownFollowers collection. If I'm not mistaken, we want the list ordered with the most recent users we followed at the top. If that's correct, we could use Collection.unshift() here.

js/views/userShort.js
```diff
@@ -0,0 +1,109 @@
...
+import Profile from '../models/Profile';
+
+export default class extends BaseVw {
+  constructor(options = {}) {
+    super(options);
+    this.options = options;
+    this.guid = options.guid || this.model.get('guid');
+    this.ownGuid = this.guid === app.profile.id;
+    this.profileArgs = {}; // create blank for placeholder render
+    this.followedByYou = app.ownFollowing.get(this.guid) !== undefined;
```
rmisio: You are using a followUtil method for this determination in other places. For consistency, should we use it here?

js/utils/follow.js
```diff
@@ -0,0 +1,49 @@
+import $ from 'jquery';
+import app from '../app';
+import Dialog from '../views/modals/Dialog';
+
+export default {
+
+  followedByYou(guid) {
+    return app.ownFollowing.get(guid) !== undefined;
```
rmisio: If you wanted, this could be written more concisely as `!!app.ownFollowing.get(guid)`.

js/views/userPage/UserPageFollow.js
```diff
@@ -1,23 +1,104 @@
...
-      this.$el.html(t({
-        ...this.model.toJSON(),
-      }));
-    });
+  updateViewerFollowers() {
+    /* if the viewer follows/unfollows this user, add them to the followers list without a fetch */
+    if (this.followType === 'Followers' && !this.options.ownPage) {
+      // if the viewer has followed add them
+      if (followUtils.followedByYou(this.model.id)) {
+        this.followCol.add({ guid: app.profile.id });
```
rmisio: Once again, if I'm not mistaking about most recent followers being at the top of the list, then we want Collection.unshift() here.

js/views/userPage/UserPageFollow.js
```diff
@@ -1,23 +1,104 @@
...
+  className() {
+    return 'userPageFollow flexRow';
+  }
+
+  render() {
+    this.$el.empty();
+    // TODO: add pagination if the collection is Followers
+    if (this.followCol.length) {
+      this.followCol.forEach((follow) => {
+        const user = this.createChild(userShort, {
```
rmisio: Some cleanup is necessary here. If on every render you are creating whole new user views (and there might be hundreds or thousands), then you need to store the current ones in a data structure... and on a new render, iterate through and call remove() on the old ones.

Yes, by using 'createChild()', when the tabView is removed, the children will be... but if while on this tab, you re-render multiple time and the list is large, you'll have a whole lot of unused views in memory. It's just safer to clean them up when you know they're not needed anymore.

js/views/userShort.js
```diff
@@ -0,0 +1,109 @@
...
+    if (guid === app.profile.id) {
+      // don't fetch our this user's own profile, since we have it already
+      profileFetch = $.Deferred().resolve();
+      profile = app.profile;
+    } else {
+      profile = new Profile({ id: guid });
+      profileFetch = profile.fetch();
+    }
+
+    profileFetch.done(() => {
```
rmisio: Can we make this profileFetch an instance level variable (`this.profileFetch`) and in remove() abort it? Also, in the done and fail handlers, can we make those not execute if the view has been removed (i.e. make the first line `if (this.isRemoved()) return;`)?

We may have a whole lot of user views on the page and there may be some queued up / in progress fetches. If the user navigates away from the page (which will ultimately cause this view's remove() to be called) we want to cancel those fetches. Otherwise, any fetches on our new page will be queued behind these. Also, odd things can happen if the handlers execute when we're on a new page.

js/views/userShort.js
jjeffryes: My thought was if we need to change how checking followedByYou works, it'd be better to have it centralized. I forgot to do that here, I'll update it.

js/views/userPage/UserPageFollow.js
jjeffryes: I was saving that for the pagination work, since that will be dynamically adding and removing the user cards. I should have been more clear this part is a placeholder.

js/views/userShort.js
jjeffryes: Good point, I'll add those.

js/views/userPage/UserPageFollow.js
rmisio: ok, let's not forget that because as it is now we have a non-insignificant memory leak. It's a really quick one to fix and have off our heads, but I guess we could defer as long as we don't forget. And.. until then, any type of scale metrics we'd do on this guy are not so accurate.

js/views/userPage/UserPageFollow.js
```diff
@@ -1,23 +1,106 @@
...
-
+    this.userCache = [];
+    // TODO: add pagination if the collection is Followers
+    if (this.followCol.length) {
+      this.followCol.forEach((follow) => {
+        const user = this.createChild(userShort, {
+          model: follow,
+        });
+        this.userCache.push(user);
+        this.$el.append(user.render().$el);
```
rmisio: Another optimization we could work in at some point is that we could add these items to a [document fragment](https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment) and then at the end add the fragment to the DOM. That way (in the case this render happens when this view is already attached to the DOM) it would be one DOM update instead of one for each user view.

js/views/userPage/UserPageFollow.js
```diff
@@ -1,23 +1,106 @@
...
+    return 'userPageFollow flexRow';
   }
 
   render() {
-    loadTemplate('userPage/userPageFollow.html', (t) => {
-      this.$el.html(t({
-        ...this.model.toJSON(),
-      }));
+    this.$el.empty();
+    this.userCache.forEach((user) => {
```
rmisio: Absolutely, 100% no need to change this, but just wanted to point out that this could be shortened to:

```
this.userCache.forEach(user => user.remove());
```

js/views/userShort.js
```diff
@@ -12,31 +12,29 @@ export default class extends BaseVw {
...
+    this.loading = true;
   }
 
   loadUser(guid = this.guid) {
     let profile;
-    let profileFetch;
+    this.fetched = true;
 
-    this.loading = true;
+    if (this.isRemoved()) return;
```
rmisio: This line should be instead both the done() and fail() handlers of the profileFetch.

PR #49, opened 2016-09-05
Listing modal

rmisio:
This PR implements the listing modal with the General sections, which (along with a dummy image) allows the creation of a listing.

Once a listing is created, you can go to a dummy listing view by entering the following route in the address bar:

`listing/<guid>/<slug>`

The dummy listing view will also allow you to open the listings modal in Edit mode.


rmisio:
There are some CSS issue I wanted to call out / request help with:

1.) When a long amount of text is output inside of the forms, it causes the right column to expand into the left column. This happens both on the medium-editor generated div on Description in the Listings modal and on the p element (not medium-editor driven) that represents the Address Notes on the Address tab in settings:

image

image

2.) The styling of the tab links does not match the designs.

image

The main reason for this is a specificity issue where the css is not structured properly. I want those tab links to have the $text1 color, but if I put the cltT on the parent, they end up getting the $text2 color because of a `:not` selector in the clrT declaration:

```
.clrT {
  color: $text;

  a:not(#{$excludedLinks}) {
    color: $text2;
  }

  input[type="radio"]:checked + label:before {
    background: $text;
  }
}
```

If I put the clrT directly on each tab link element, then because of the :not selector having a higher specificity, the clrT on the body element overrides the directly placed clrT and I'm still left with a $text2 color.

Same situation if I want to set a color on the "active" link. Any clr* class I place directly on the the active element will be overridden.

Another issue is that I'm not sure any of the existing text colors match the color in the design. We would need to adjust the colors in the theme or add another color. (can I get a clrT34?? :) )

3.) I'm unable to get the inner scroll bar on the Edit listing modal to work right. I tentatively sized it's element to a hard-coded 700px. The better approach would be for it to automatically take up the available space, while respecting the max-height set on .modalContent, but any variation of percentage width I put on it (be it directly and/or on its parents) doesn't seem to come up with the right height. It seems the percentages are relative to it's content, rather than the height of .modalContent.


---------------

Could you take a look at these issues?

styles/_theme.scss
```diff
@@ -125,3 +125,19 @@ $excludedLinks: ".btn, .iconBtn, .listItem";
 .clrSh2  {
```
rmisio: One suggestion I wanted to make for customizing the theme would be to use template to generate the styles, for example:

```
// theme template
.clrP {
  background-color: <%= ob.clrPrimary %>;

  input[type="radio"] + label:before {
    background: <%= ob.clrSecondary %>;
  }

  input[type="radio"]:checked + label:before {
    box-shadow: inset 0 0 0 3px <%= ob.clrSecondary %>;
  }
}

...
```

```
// when a theme change is needed
$('#themeStyle').remove(); 
$('<style id="themeStyle" />').text(themeTmpl.render(profile.toJSON()))
  .appendTo($('head')); // we should cache the $('head') selector

// I'm assuming the theme color would come from some user's profile model, but we can adjust if this is not the case.

File: js/models/listing/Metadata.js [hunk 2]
```diff
@@ -0,0 +1,47 @@
...
+  // todo: validate the listing type is one of the available types
+  validate(attrs) {
+    const errObj = {};
+    const addError = (fieldName, error) => {
+      errObj[fieldName] = errObj[fieldName] || [];
+      errObj[fieldName].push(error);
+    };
+
+    if (this.contractTypes.indexOf(attrs.contractType) === -1) {
+      addError('contractType', 'The contract type is not one of the available types.');
```
rmisio: Same as other PRs, any non-translated model errors are ones that (short of a developer error) would never be seen by the user.

jjeffryes:
Add flex-shrink: 0  to the .tabColumn class to fix the issue of the content column growing.

styles/_theme.scss
jjeffryes: That's a good idea. I think it would be cleaner than how it was done in v1.

rmisio:
closing in liu of https://github.com/OpenBazaar/openbazaar-desktop/pull/54

PR #50, opened 2016-09-06
Overwrite the Revert

hoffmabc:
Ok @rmisio this is the fix so you can undo the revert. I tested.


hoffmabc:
@rmisio @jjeffryes not sure how you want to handle this, but this fix should solve the exception that was being caused when you merged Rob.

PR #54, opened 2016-09-08
Listing modal

rmisio:
This PR implements the listing modal with the General section and Photo upload section.

Once a listing is created, you can go to a dummy listing view by entering the following route in the address bar:

listing/\<guid\>/\<slug\>

The dummy listing view will also allow you to open the listings modal in Edit mode.


js/languages/en-US.json
```diff
@@ -107,6 +155,18 @@
     "provideName": "Please provide a name.",
     "provideCountry": "Please provide a country."
   },
+  "listingInnerModelErrors": {
+    "provideSlug": "Please provide a slug."
+  },
+  "itemModelErrors": {
+    "provideTitle": "Please provide a title.",
+    "badConditionType": "Please provide the price amount as a number.",
+    "descriptionTooLong": "The price must be greater than 0.",
```
jjeffryes: Is this the right phrase?

js/views/modals/EditListing.js
```diff
@@ -0,0 +1,478 @@
...
+  onClickCancelPhotoUploads() {
+    this.inProgressPhotoUploads.forEach(photoUpload => photoUpload.abort());
+  }
+
+  onChangePrice(e) {
+    let updatedVal = $(e.target).val().trim();
+    const valAsNumber = Number(updatedVal);
+
+    if (!isNaN(valAsNumber)) {
+      updatedVal = valAsNumber.toFixed(2);
```
jjeffryes: If your currency is BTC, you need to be able to enter prices as small as 0.0005, this is truncating the price to zero if it's lower than 0.01.

js/models/listing/Metadata.js
```diff
@@ -0,0 +1,52 @@
...
+  get contractTypes() {
+    return [
+      'PHYSICAL_GOOD',
+      'DIGITAL_GOOD',
+      'SERVICE',
+      'CROWD_FUND',
+    ];
+  }
+
+  // todo: validate the listing type is one of the available types
```
jjeffryes: is this todo taken care of below?

rmisio: no, that one is for the contractType. Still need to do listingType validation at some point. I don't think the listing type is even in the designs at this point.

js/views/modals/EditListing.js
rmisio: ok, so I'll adjust so BTC rounds to 4 places after the decimal.

js/templates/modals/editListing/editListing.html
```diff
@@ -0,0 +1,187 @@
...
+              </div>            
+            </div>
+            <div class="flexRow gutterH">
+              <div class="col6">
+                <label for="editListingPrice" class="required"><%= ob.polyT('editListing.price') %></label>
+                  <% if (ob.errors['listing.item.price']) print(ob.formErrorTmpl({ errors: ob.errors['listing.item.price'] })) %>
+                <div class="inputSelect">
+                  <%
+                    let listingPrice = ob.listing.item.price;
+                    listingPrice = typeof listingPrice === 'number' ? listingPrice.toFixed(2) : listingPrice;
```
jjeffryes: Don't forget to update for BTC having more decimals here too.

js/views/PageNav.js
```diff
@@ -171,6 +179,16 @@ export default class extends View {
     this.togglePopMenu();
   }
 
+  navCreateListingClick() {
+    const listingModel = new Listing();
+
+    launchEditListingModal({
+      mode: 'create',
```
jjeffryes: Is mode being used?

js/views/modals/EditListing.js
```diff
@@ -0,0 +1,529 @@
...
+
+      new SimpleMessage({
+        title: app.polyglot.t('editListing.errors.tooManyPhotosTitle'),
+        message: app.polyglot.t('editListing.errors.tooManyPhotosBody'),
+      })
+      .render()
+      .open();
+    }
+
+    if (!photoFiles.length) return;
```
jjeffryes: Very minor, but this could be before checking the current length plus the new length, since if the new length is zero you don't need to find out if the combined value is more than the max.

rmisio: but if they exceed the max length, I strip out any about the max, which may be all of them. That's why I need the check after the potential stripping.

Eventually, I could update the UI, so the Add button is disabled if they are at the max. But, if they are at 9 and click add, they could still select more than 1, in which case I'll need to strip out all but 1.

jjeffryes: Ah, right, the slice might reduce it to zero.

PR #62, opened 2016-09-29
Upload avatar and header in settings

jjeffryes:
- adds avatar and header upload controls to the settings modal
- the controls use the cropit.js library
- each one can be scaled and rotated
- the avatar and header API save the hashes to the profile, but this takes some time, so the hashes are added manually to the profile model when they are returned by the respective APIs.


js/start.js
```diff
@@ -307,6 +306,12 @@ function start() {
     app.ownFollowers = new Followers(null, { type: 'followers' });
 
     onboardIfNeeded().done(() => {
+      // add listener to pageNav after the profile is available
+      app.pageNav.listenTo(app.profile, 'change', () => {
```
rmisio: We can't get away with re-rendering here because the profile might be updated at random times (AJAX request completes at random time) and if the user is typing in the address field, it'll lose focus and whatever they are typing will be cleared. Also, the page url is lost on re-render. I don't think we'll be re-rendering this view too much.

I think the better approach here is to bind a change event specifically to the avatar field and then set the image url with jQuery.

Also, this event and handler should be bound in the PageNav view and not here.

js/templates/userPage/userPage.html
```diff
@@ -3,7 +3,12 @@
       <div class="flexVCent pageTabs">
         <!--consider turning tabs into a reusable view-->
         <div>
-          <div class="userIcon disc clrBr2 clrSh1" style="background-image: url('../imgs/defaultUser.png')"></div>
+          <div class="userIcon disc clrBr2 clrSh1"
+            <% if(ob.avatarHash){ %>
+               styel="background-image: url(<%= ob.getServerUrl(`ipfs/${ob.avatarHash}`) %>), url('../imgs/defaultBanner.png')"
```
rmisio: Shouldn't the fallback be defaultUser?

js/templates/userShort.html
```diff
@@ -1,29 +1,31 @@
...
+    <% if(ob.headerHash){ %>
+      style="background-image: url(<%= ob.getServerUrl(`ipfs/${ob.headerHash}`) %>), url('../imgs/defaultHeader.png')"
     <% } else { %>
       style="background-image: url('../imgs/defaultHeader.png')"
     <% } %>>
     <div class="userIcon disc clrBr2 clrSh1"
-      <% if(ob.avatar){ %>
-       style="background-image: url('<%= ob.avatar %>, url('../imgs/defaultUser.png')"
+      <% if(ob.avatarHash){ %>
+        style="background-image: url(<%= ob.getServerUrl(`ipfs/${ob.avatarHash}`) %>), url('../imgs/defaultBanner.png')"
```
rmisio: Should the fallback be defaultUser?

js/templates/userShort.html
```diff
@@ -1,29 +1,31 @@
...
-          <a class="iconBtn clrP clrBr js-mod <% if(ob.ownMod){ %>active<% } %>">
-            <i class="ion-briefcase"></i>
+    <% if(!ob.hideControls){ %>
+      <div class="userControls">
+        <% if(!ob.ownGuid) { %>
+          <a class="iconBtn clrP clrBr js-follow <% if(ob.followedByYou){ %>active<% } %>">
+            <i class="ion-person-stalker"></i>
           </a>
+          <% if(ob.moderator) { %>
+            <a class="iconBtn clrP clrBr js-mod <% if(ob.ownMod){ %>active<% } %>">
```
rmisio: lol, i got to find a linter for our templates so we can have some spaces after an 'if' and before the curly brace!

rmisio:
Should we also disable the zoom slider when there is no image there? It seems odd that you can still move it.

image

js/views/modals/EditListing.js
```diff
@@ -208,7 +251,34 @@ export default class extends BaseModal {
 
         canvas.width = imgW;
         canvas.height = imgH;
-        canvas.getContext('2d').drawImage(newImage, 0, 0, imgW, imgH);
+
+        switch (orientation) {
```
rmisio: What is the purpose of this code? When I comment out this code and I rotate an image on my computer and then upload, the rotation is respected. What am i missing?

image

Also, this code seems quite complex, would [this](http://stackoverflow.com/a/27882870/632806) not be simpler?

js/views/modals/Settings/Page.js
```diff
@@ -37,27 +92,46 @@ export default class extends baseVw {
...
-
-      save.done(() => this.trigger('saveComplete'))
-        .fail((...args) =>
-          this.trigger('saveComplete', false, true,
-            args[0] && args[0].responseJSON && args[0].responseJSON.reason || ''));
     }
 
-    // render so errrors are shown / cleared
-    this.render();
+    if (this.avatarChanged && this.avatarCropper.cropit('imageSrc')) {
```
rmisio: Can you please wrap the `if (this.avatarChanged...`, `if (this.headerChanged..` & `$.whenAll(save..` blocks in an `if (save) {` so that they're only conditionally executed if client side validation passes?

If client side validation fails, we shouldn't be kicking off any image uploads until the user fixes the errors.

rmisio:
I touched on this before, but I think we need to adjust the sizing of how we display the header image on the user page. I don't think we could have it be 100% width because, unless they happen to upload an image larger than the largest monitor in the world, there's always going to be distorting:

image

Ideally, it would stretch up until it's natural dimensions, but I'm not sure if this is possible with pure CSS.

js/views/modals/Settings/Page.js
```diff
@@ -37,27 +92,46 @@ export default class extends baseVw {
...
     const save = this.profile.save();
+    let saveAvatar;
+    let saveHeader;
 
     this.trigger('saving');
 
     if (!save) {
       // client side validation failed
       this.trigger('saveComplete', true);
+      this.render();
```
jjeffryes: The render needs to happen after any uploads of the avatar and header, otherwise the new images are replaced by the old ones.

js/views/modals/Settings/Page.js
```diff
@@ -37,27 +92,46 @@ export default class extends baseVw {
...
+        this.profile.set('headerHash', headerData.hash);
+        app.profile.set('headerHash', headerData.hash);
+      });
+    }
 
-    if ($firstErr.length) $firstErr[0].scrollIntoViewIfNeeded();
+    $.whenAll(save, saveAvatar, saveHeader)
+      .done(() => {
+        this.trigger('saveComplete');
+        this.render();
```
jjeffryes: If the user uploads an avatar or header, the view can't render until the hashes for the new images are returned, otherwise the images are set to their previous values, which is incorrect.

So if we need to render to show errors, that has to happen after the hashes are returned. 

Unless there's another way around that, we could put an overlay to block the form while the API requests are pending, like we do in the address section.

rmisio: Ok, this is how I would do it and it would solve this issue.

This is pseudo code, hasn't been tested, but the gist is that in render you aren't exclusively setting the cropIt image src based off the model. If it had already been set, you are setting it to the existing version, otherwise the model, otherwise nothing.

So, at the top of render (before loadTemplate) put this:

```
    let currAvatar;

    if (this.avatarCropper) {
      currAvatar = this.avatarCropper.cropit('imageSrc');
    }
```

then change the logic where you set the imageSrc to:

```
        if (currAvatar || this.profile.get('avatarHash')) {
          const imageSrc = currAvatar ||
            app.getServerUrl(`ipfs/${this.profile.get('avatarHash')}`);

          this.avatarCropper.cropit('imageSrc', imageSrc);
        } 
```

Now with this code in place, you are free to render() to hide/show errors (how it was before) and you won't lose the existing image state that's already in crop it.

Also, when the uploads complete, there's no need to re-render, because the images are already there. In fact, you can't re-render because you can't just re-render at a random time when the user might be typing in a field.

js/views/modals/Settings/Page.js
```diff
@@ -46,28 +123,164 @@ export default class extends baseVw {
...
+        })
+        .fail((...args) => {
           this.trigger('saveComplete', false, true,
-            args[0] && args[0].responseJSON && args[0].responseJSON.reason || ''));
+            args[0] && args[0].responseJSON && args[0].responseJSON.reason || '');
+        });
     }
 
-    // render so errrors are shown / cleared
     this.render();
```
jjeffryes: Yeah, just forgot that.

PR #66, opened 2016-10-07
More listing fields

rmisio:
This change-set adds in the Tags, Category, Return Policy and Terms & Conditions fields to the Listing modal.


js/views/modals/EditListing.js
```diff
@@ -514,19 +536,86 @@ export default class extends BaseModal {
...
+        // and think you are selecting that.
+        matcher: () => false,
+      }).on('change', () => {
+        const tags = this.$editListingTags.val();
+        this.innerListing.get('item').set('tags', tags);
+        this.$editListingTagsPlaceholder[tags.length ? 'removeClass' : 'addClass']('emptyOfTags');
+      });
+
+      this.$editListingTagsPlaceholder[
+        this.$editListingTags.val().length ? 'removeClass' : 'addClass'
```
jjeffryes: Nice use of logic inside the array notation here.

styles/components/_select2.scss
```diff
@@ -22,3 +22,72 @@
...
+// Focus styling is with other form focus styling in form.scss.
+.select2TagsWrap {
+  position: relative;
+
+  .select2Tags + .select2-container {
+    &.select2-container--default {
+      .select2-selection--multiple .select2-selection__rendered {
+        padding-bottom: $padSm;
+        border-radius: 3px;
+      }
```
jjeffryes: The .select2-selection--multiple element is getting a height of 4em, that makes it shorter than the UL and LI elements, which throws off other measurements. 

If you gave it a height of auto, that would cause it to contain the UL and LI, instead of being shorter.

```
.select2-selection--multiple {
        height: auto;

        .select2-selection__rendered {
          padding-bottom: $padSm;
          border-radius: 3px;
        }
      }
```

The height of the control is jumping up when a tag is added, and then going back down when all tags are removed, this is because the input is shorter than the tags. That could be fixed by giving it a line-height to make it the same height (inside .select2-container--default). 

```
.select2-search--inline .select2-search__field {
        line-height: 42px;
      }
```

styles/components/_select2.scss
```diff
@@ -22,3 +22,72 @@
...
+    }
+
+    &.select2-container--focus {
+      .select2-selection {
+        border: 1px solid transparent;
+      }
+
+      & + .tagsPlaceholder {
+        display: none;
```
jjeffryes: With the changes above, the placeholder would need top: 0 and line-height: 54px to center it in the control.

styles/modules/modals/_editListing.scss
```diff
@@ -60,6 +60,18 @@
...
+      .tagsSection {
+        .select2 {
+          margin-bottom: $padSm;
+        }
+      }
+
+      .categoriesSection {
+        .select2 {
+          margin-bottom: $padSm;
+        }
```
jjeffryes: If you make the changes to the select2 tags above, the hrs after the control will be positioned right, and you won't need the extra padding here.

js/templates/modals/editListing/editListing.html
```diff
@@ -153,16 +154,44 @@ <h1 class="tx3 tx500 floL"><%= ob.polyT('editListing.sectionNames.photos') %></h
           </ul>
         </section>
         <hr class="dashed margR margL" />
-        <section class="js-scrollToSection" style="height: 350px;background-color: orange">
+        <section class="js-scrollToSection pad tx3">
+          <div class="overflowAuto">
+            <h1 class="tx3 tx500 floL"><%= ob.polyT('editListing.sectionNames.attachments') %></h1>
```
rmisio: Can we cross that bridge when we come to it? Mike is updating the designs and I'd rather wait for that. In either case, I think we'll need a PR to retrofit existing areas that use the current font.

styles/components/_select2.scss
rmisio: In my new branch, for Shipping, I've added a height 100% to a certain element that has fixed this issue. So, this will be fixed in a future PR.

PR #67, opened 2016-10-11
Fixes some layout issues.

jjeffryes:
This will fix issues with the white control background not matching the border:
image

And with the height of the wrapping elements being smaller than the UL, which throws off any elements that come after the select2 control.

### DIFF
--- a/styles/components/_select2.scss
+++ b/styles/components/_select2.scss
@@ -30,9 +30,18 @@
 
   .select2Tags + .select2-container {
     &.select2-container--default {
+      .select2-selection--multiple {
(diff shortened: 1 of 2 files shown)


styles/components/_select2.scss
```diff
@@ -84,10 +93,13 @@
     // This class must be set via JS
     &.emptyOfTags {
       display: block;
-    }    
+      top: 0;
+      line-height: 4rem;
```
jjeffryes: The line height is set to 4rem here, because that's what the original select2 css sets elements in the control to.

PR #69, opened 2016-10-14
Notofont

jjeffryes:
Changes the font to Noto. 

We will need to add additional Noto fonts for languages that aren't included in the core Noto font, along with code to switch to those fonts, when we add additional languages. 

For example, there is a Noto Sans Thai font that has Thai characters not in the main Noto font. When we add a Thai language option, selecting it will also need to set it as the font to use.

This PR also fixes a typo in the user page.


rmisio:
Looks like we're going to need to adjust our text sizes with this new font. The listing modal looks a bit off (doesn't show so well in this screenshot). Looks like if $tx3 were dropped from 1.7rem to 1.6rem, it would more accurately match the designs. That, or possibly the listing modal should use $tx4, which would get it to 1.5 rem.

image

PR #70, opened 2016-10-14
Adjust avatar header upload

jjeffryes:
- makes changes to match the new data structure for avatars and headers
- adds app.hiRez as a value set in start.js for whether the computer has a high dpi screen or not
- adds logic to the display the right images size for avatars and headers based on the hiRez value


js/start.js
```diff
@@ -54,6 +54,11 @@ app.localSettings.on('change:language', (localSettings, lang) => {
   .open();
 });
 
+app.hiRez = (
```
rmisio: 1.) Why is the matchMedia part necessary? Wouldn't chromium always give us the devicePixelRatio?

2.) Instead of attaching this to app, can we make this into a utility? Either make an appropriately named module or (if it's a one-off) add it to utils/index.js.

js/views/userPage/UserPage.js
```diff
@@ -125,6 +125,7 @@ export default class extends baseVw {
         followed: this.followedByYou,
         followsYou: this.followsYou,
         ownPage: this.ownPage,
+        hiRez: app.hiRez,
```
rmisio: Since I imagine this will be used by many, many template, can we please add it as a template helper rather than having to pass it in all the time?

js/views/modals/Settings/Page.js
```diff
@@ -206,20 +206,21 @@ export default class extends baseVw {
...
+      // if the avatar or header exist, don't count the first load as a change
+      this.avatarLoadedOnRender = Boolean(this.profile.get('avatarHashes').original);
+      this.headerLoadedOnRender = Boolean(this.profile.get('headerHashes').original);
 
       setTimeout(() => {
         this.avatarCropper.cropit({
           $preview: avatarPrev,
           $fileInput: avatarInpt,
-          exportZoom: 4.7,
+          // exportZoom: 4.7,
```
rmisio: should we be leaving this commented out value in?

js/models/Profile.js
```diff
@@ -7,10 +7,10 @@ export default class extends BaseModel {
   defaults() {
     return {
       about: '',
-      avatarHash: '',
+      avatarHashes: { tiny: '', small: '', medium: '', large: '', original: '' },
```
rmisio: hmm... in general, for nested objects or arrays of objects, I've been using nested models and collections. It offers a few benefits, for example, a nice concise place to put validations, being able to have more granular change events (e.g. app.profile.get('avatarHash).listenTo('change:medium'), etc.

I guess this is a simple enough structure that we may be able to get away with it being a flat object... but for anything more complex, I think we'd want to take advantage of the full API of a nested Model.

(for the listing, I will probably use a nested model. If I discover some more advantages to that approach, I may retrofit this guy)

js/start.js
jjeffryes: 1. the research I did said that you need to do both to catch all devices, so this makes it more reliable for someone running it on a tablet or something else weird. 

2. good idea.

js/views/userPage/UserPage.js
jjeffryes: Good call.

js/models/Profile.js
jjeffryes: Sounds good. I can add the nested model.

js/utils/templateHelpers.js
```diff
@@ -1,7 +1,22 @@
 import app from '../app';
+import { getHiRez } from './responsive';
```
rmisio: Can we combine these into one statement?

```
import { a, b, c } from './module';
```

js/utils/templateHelpers.js
```diff
@@ -1,7 +1,22 @@
...
+import { getLargeWidth } from './responsive';
+import { getSmallHeight } from './responsive';
 
 export function polyT(...args) {
   return app.polyglot.t(...args);
 }
 
 export const getServerUrl = app.getServerUrl;
+
+export function hiRez() {
```
rmisio: This can be written more concisely as `export { getHiRez as hiRez };`... but why the name change? Without the name change, it could just be `export { getHiRez };`.

jjeffryes: No particular reason for the name change, just incomplete refactoring from earlier code. I'll change it.

js/views/modals/Settings/Page.js
```diff
@@ -232,7 +232,10 @@ export default class extends baseVw {
               new SimpleMessage({
```
rmisio: If you load an avatar or header that is too small, this message appears on every render. It wasn't doing this before, it was just showing when you loaded the image, which I think is the correct behavior.

With it showing on every render, every time you save the modal (whether you updated the header or now), the message will show. Also, if you close and re-open the model, it will show.

jjeffryes: This is because cropit is no longer upscaling the image, it's saving it at the size of the original (with the cropping applied). 

I saw the issue yesterday, I kind of thought it was good to leave it in as a nag, but you're right, it should only warn you when you do the original upload and/or save (if the cropped image is too small after cropping).

rmisio:
@jjeffryes Would you mind fixing the merge conflicts?

js/utils/responsive.js
```diff
@@ -0,0 +1,14 @@
+// functions for determining size and resolution of the window
+
+export function getHiRez() {
```
rmisio: Shouldn't these really be isHiRez, isLargeWidth?

rmisio:
With the current code, I'm unable to move the zoom handle for either image.

js/utils/responsive.js
jjeffryes: That would fit our naming convention better, I'll update.

rmisio:
Ok, I see what's happening... if the image is below the recommended size, I can't adjust the zoom. Hmm... at the very least I would think we need to tweak the UX on that, because it just appeared broken to me. Maybe we could give it a disabled state so it's clear it's disabled?

jjeffryes:
If you can't move the zoom handle, it should mean zooming would make the image too low rez. 

We should probably put a warning or some other indicator if that's the case. 

Try using a larger image, if the zoom doesn't start working again, it's a bug.

EDIT: cross posting!

rmisio:
I don't know if it's a perfect solution, but visually just adding a disabled class to the zoom input seems to look ok:

image

But, you'll also need to ensure that after a render the disabled state is still there when necessary.

js/views/modals/Settings/Page.js
```diff
@@ -28,39 +28,50 @@ export default class extends baseVw {
...
-      this.avatarChanged = true;
-    }
+    this.avatarRotate(1);
   }
 
-  headerLeftClick() {
+  headerRotate(direction) {
     if (this.headerCropper.cropit('imageSrc')) {
-      this.headerCropper.cropit('rotateCCW');
+      const loadedSize = this.headerCropper.cropit('imageSize');
```
rmisio: The block to determine whether an avatar/header is too small is repeated. Could we break that determination off into a separate function? It's possible we'll need to determine this elsewhere in the view and I wouldn't want to repeat the block again.

e.g:

```
  headerTooSmall() {
    const loadedSize = this.headerCropper.cropit('imageSize');

    return (loadedSize.height < this.headerMinWidth ||
        loadedSize.width < this.headerMinHeight);
  }
```

rmisio:
There's a bit of an inconsistency here. If the user has no image, the 'image too small to zoom' message overlays the zoom handle:

image

Whereas,once you upload an image that is too small, the message replaces the handle:

image

Probably the latter is better, but can we choose one and make it consistent?

jjeffryes:
That's an oversight, the message shouldn't appear if there is no image loaded. I'll fix it.

rmisio:
The avatar too small alert dialog shows both when you load an image that is too small and subsequently when you save the profile. Could it please just show when you load the image? It seems redundant to show it both times.

If you keep re-saving, then it won't show anymore. But, the first time you save an image that is too small, you're going to get the double alert.

PR #71, opened 2016-10-14
Listing shipping options

rmisio:
This change-set implements the Shipping section within the Listing modal.


rmisio:
@jjeffryes If you were feeling brave (as you know, select2 css is a bit heinous!), I have a select2 styling issue that I wanted to look at. It's something I mentioned before, but basically the drop down is having it's width set dynamically via JS by the select2 plugin. For some reason, it's setting it a few pixels too wide. Perhaps, some aspect of our CSS on the main field is causing the issue?

image

Anyhow, if you see some reasonably easy way to fix it, I'm all ears.

Otherwise, the couple of some-what icky ideas I have are:

option 1.) Have our JS dynamically set the width to the right size. This is a bit tricky because we need to ensure our JS happens after theirs. Also, ideally, we'd want to do it in a way where there is no noticeable flicker (e.g. drop-down opens to one width and a split second later sizes down).

option 2.) We can't just adjust the width of the top field, because the drop down width is calculated off of that every time it's opened. But, we could possible add an `::after` pseudo content to the field that is a few pixels wide with a white background and black border and position it in a way that it makes the top field look slightly wider to line up with the drop down. Sounds like this would probably work and perhaps the only down-side is that it adds another piece to handle in the themes style sheet.

rmisio:
@jjeffryes I updated this PR to work with the latest server.

Still need to look at the CSS PR you submitted to this branch. Will do so tomorrow.

jjeffryes:
image

The error messages throwing the input fields out of alignment is kind of unpleasant looking. We should probably think of a better way to handle that. I have a few ideas. cc: @morebrownies

jjeffryes:
If a user adds a completely blank service by accident, it seems like it would be better to just remove it rather than throw an error.

rmisio:
@morebrownies @jjeffryes 
re: error message alignment. I saw that and meant to fix. I think the easiest thing is just to bottom align the fields:

image

rmisio:
> If a user adds a completely blank service by accident, it seems like it would be better to just remove it rather than throw an error.

Well, at least one service is required, so I couldn't just remove that one. If there's blank ones beyond the first one, I could remove it, but it would mean ensuring all fields on that service are blank. My concern is future proofing. What if we add a drop down field to a service? That field would always have a value and I wouldn't know (without monitoring change events) whether that field was set by the user or not.

It seems like a reasonable thing for me that if a user adds a blank service (or few), they should have to clean up after themselves (we're talking one click per service).

js/views/modals/editListing/ShippingOption.js
```diff
@@ -0,0 +1,216 @@
...
+    this.listenTo(view, 'click-remove', e => {
+      this.services.remove(
+        this.services.at(this.serviceViews.indexOf(e.view)));
+    });
+
+    return view;
+  }
+
+  get $headline() {
+    return this._$headline || this.$('h1');
```
jjeffryes: Remember to set this._$headline.

js/views/modals/editListing/ShippingOption.js
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  get $shipDestinationDropdown() {
+    return this._$shipDestinationDropdown || this.$(`#shipDestinationsDropdown_${this.model.cid}`);
+  }
```
jjeffryes: Set this._$shipDestinationDropdown

js/views/modals/editListing/ShippingOption.js
```diff
@@ -0,0 +1,216 @@
...
+
+  get $serviceSection() {
+    return this._$serviceSection || this.$('.js-serviceSection');
```
jjeffryes: Set this._$serviceSection.

js/views/modals/editListing/EditListing.js
```diff
@@ -494,6 +582,33 @@ export default class extends BaseModal {
     return this._$editListingTermsAndConditions || this.$('#editListingTermsAndConditions');
```
jjeffryes: Set this._$editListingTermsAndConditions.

js/views/modals/editListing/EditListing.js
```diff
@@ -494,6 +582,33 @@ export default class extends BaseModal {
...
   }
 
+  get $sectionShipping() {
+    return this._$sectionShipping || this.$('.js-sectionShipping');
```
jjeffryes: Set this._$sectionShipping

styles/components/_type.scss
```diff
@@ -77,10 +69,14 @@ p {
   text-align: center;
 }
 
-.txUnb {
+#ov1 .txUnb {
```
jjeffryes: The return of the dreaded override class. ;)

styles/modules/modals/_editListing.scss
```diff
@@ -7,6 +7,36 @@
...
+
+      .closeIcon {
+        width: 24px;
+        height: 24px;
+        position: absolute;
+        top: 0;
+        left: 0;
+      }      
+
+      button {
```
jjeffryes: Even though it's unlikely there will be another button element inside the tile, since the button already has the .btn class, can this just use that class instead?

styles/modules/modals/_editListing.scss
```diff
@@ -35,26 +56,59 @@
...
+        form {
+          label {
+            font-weight: 400;          
+          }
+
+          .shippingOptionButtonBar {
+            margin-top: 25px;
+          }
+
+          .servicesWrap {
```
jjeffryes: You could take out some of this by using flexColRows and gutterV instead, since that's designed to give a vertical stack with no padding on the sides.

js/templates/modals/editListing/editListing.html
```diff
@@ -65,21 +65,10 @@ <h1 class="h3 txUnb padMd js-listingHeading"><% print(ob.createMode ? ob.polyT('
                 <label for="editListingPrice" class="required"><%= ob.polyT('editListing.price') %></label>
                   <% if (ob.errors['listing.item.price']) print(ob.formErrorTmpl({ errors: ob.errors['listing.item.price'] })) %>
                 <div class="inputSelect">
-                  <%
-                    const decimalPlaces = ob.listing.metadata.pricingCurrency === 'BTC' ? 8 : 2;
-                    let listingPrice = ob.listing.item.price;
-                    listingPrice = typeof listingPrice === 'number' ? listingPrice.toFixed(decimalPlaces) : listingPrice;
-                  %>
-                  <input type="text" class="clrBr clrP js-price" name="listing.item.price" id="editListingPrice" value="<%= listingPrice%>" data-var-type="number">
+                  <input type="text" class="clrBr clrP js-price" name="listing.item.price" id="editListingPrice" value="<% if (typeof ob.listing.item.price === 'number') print(ob.formatPrice(ob.listing.item.price, ob.listing.metadata.pricingCurrency === 'BTC')) %>" data-var-type="number">
```
rmisio: To determine whether we want to add the decimal places in. If it's not set, it will be `undefined` and in that case, we don't want to output anything for the value attribute of this input (i.e. we want the field to be empty).

If it's a number, then we want to format it by adding the appropriate decimal places (e.g. 123 => 123.00).

js/views/modals/editListing/EditListing.js
```diff
@@ -55,12 +62,39 @@ export default class extends BaseModal {
...
+    this.shippingOptionViews = [];
 
     loadTemplate('modals/editListing/uploadPhoto.html',
       uploadT => (this.uploadPhotoT = uploadT));
 
     this.listenTo(this.images, 'add', this.onAddImage);
     this.listenTo(this.images, 'remove', this.onRemoveImage);
+
+    this.listenTo(this.shippingOptions, 'add', (shipOptMd) => {
+      if (this.shippingOptions.length) {
```
rmisio: I guess not - will update.

styles/components/_type.scss
rmisio: you best recognize!

styles/modules/modals/_editListing.scss
rmisio: Yes, will do. Also, as an FYI, the parent of this guy is .tile which is something I'll re-use when I end up coding up attachments:

image

That's something you should be able to re-use for the listing detail view (without the close icon). I'll probably make a template for it and we could put that template and the tile class in a location less specific to the edit modal.

But, i think we could hold off on doing that until the server actually implements attachments.

styles/modules/modals/_editListing.scss
rmisio: I'll put it another way... I welcome PRs :)

jjeffryes: lol. It's not a huge deal, but the vertical column classes would make things more efficient, and may be useful in some other ways. I may take a look at it later and submit something.

rmisio:
Not ready for merge yet.

js/templates/modals/editListing/editListing.html
```diff
@@ -33,26 +33,19 @@ <h1 class="h3 txUnb padMd js-listingHeading"><% print(ob.createMode ? ob.polyT('
...
                 <% if (ob.errors['listing.metadata.contractType']) print(ob.formErrorTmpl({ errors: ob.errors['listing.metadata.contractType'] })) %>
-                <select id="editformat" name="listing.metadata.contractType" class="clrBr clrP">
+                <select id="editContractType" name="listing.metadata.contractType" class="clrBr clrP marginTopAuto">
                   <% ob.contractTypes.forEach((contractType) => { %>
                   <option value="<%= contractType.code %>" <% if (contractType.code === ob.listing.metadata.contractType) print('selected') %>><%= contractType.name %></option>
                   <% }); %>
                 </select>
               </div>
-              <div class="col6">
+              <div class="col6 simpleFlexCol">
```
rmisio: I'm using this class to help with bottom aligning fields so they look nicer when there's errors on fields that are in columns. I tried using the `flexCol` and `flexKidsExpand` classes for this, but they were causing the Select2 fields to not be 100% wide.

Anyhow, if it's important enough to you and you want to play around with it, feel free to submit a PR.

rmisio:
Ok... if it passes you're review, this PR is ready to merge.

jjeffryes:
If you start adding a shipping option, then switch to a digital good or service, the half-entered shipping option just greys out, which is a little weird. Not necessarily something that has to change.

js/utils/currency.js
```diff
@@ -38,3 +38,22 @@ export function integerToDecimal(amount, isBtc = false) {
...
+    throw new Error('Please provide a price as a number');
+  }
+
+  if (isNaN(price)) {
+    throw new Error('Please provide a price that is not NaN');
+  }
+
+  const decimalPlaces = isBtc ? 8 : 2;
+
+  return price.toFixed(decimalPlaces);
```
jjeffryes: In the future it seems like this would be a good place to add internationalization of the number formats for prices.

rmisio: Brilliant idea! Please see [here](https://github.com/OpenBazaar/openbazaar-desktop/pull/82/files#diff-a33774cec9d5ed9ee7be4dcfc0cc95bdR76) from the [store PR](https://github.com/OpenBazaar/openbazaar-desktop/pull/82).

PR #82, opened 2016-10-26
Store

rmisio:
Initial version of the store. Includes:
- displaying a grid of your listings 
- open a stubbed listingDetail modal by clicking on the title of a listing
- from the stubbed listingDetail, open the edit listing modal
- load a listing (via the store) by entering the url in the address bar

When it's peer approved, we could go ahead an merge this PR. I'll make the rest of the store in a subsequent PR (or few). It's useful to get this guy in because it provides some useful stubbed functionality for the listing detail modal.

Please note: This PR includes the code from the current listingShipping PR. Once that guy is merged, the diff report on this PR will be leaner.


js/utils/currency.js
```diff
@@ -38,3 +40,79 @@ export function integerToDecimal(amount, isBtc = false) {
...
+    throw new Error('Please provide a price that is not NaN');
+  }
+
+  const decimalPlaces = isBtc ? 8 : 2;
+
+  return price.toFixed(decimalPlaces);
+}
+
+// todo: unit test me!!!
+export function formatCurrency(amount, currency, locale = app.settings.get('language')) {
```
rmisio: @jjeffryes this will probably be useful to you on the listingDetail modal.

js/router.js
```diff
@@ -77,19 +76,65 @@ export default class ObRouter extends Router {
...
+      'followers',
+    ];
+  }
+
+
+  /**
+   * Based on the route arguments, determine whether we
+   * have a valid user route.
+   */
+  isValidUserRoute(guid, state, ...deepRouteParts) {
```
rmisio: I made some changes in the user page route so that it's more explicit on what are valid urls. This is something a while back I talked about with @morebrownies and he seemed to agree with me that, in general, if a user enters an unrecognized url, we take them to a Page Not Found page, instead of trying to guess where they want to go. I think it's much cleaner in that it takes out ambiguity of what a route represents.

Anyhow, before it was like this:

\<guid|handle\>/crap
\<guid|handle\>/crap/hippo

would all redirect to `<guid|handle>/store`.

Similarly:

\<guid|handle\>/home/poo
\<guid|handle\>/following/shoo

would redirect to `<guid|handle>/home` & `<guid|handle>/following` respectively.

Now, any unrecognized urls will redirect to a Page Not Found page (which obviously can use a better design):

image

The one exception to this is entering just the \<guid|handle\> in the address bar, which will redirect you to `
<guid|handle>/store`.

morebrownies:
Thanks @rmisio 

I added the "Page not found" design to my to do list. I'll clean it up later.

rmisio:
closing in liu of #84

PR #84, opened 2016-11-03
Store filters

rmisio:
This is an unfinished version of a user's Store tab. However, it does provide a lot of foundation code for the Listing Detail view dev request, so please merge this before starting on that.

Functionality it includes:
- displaying a grid of your listings
- open a stubbed listingDetail modal by clicking on a listing card
- from the stubbed listingDetail, open the edit listing modal
- load a listing (via the store) by entering the url in the address bar
- hovering over a listing card displays an overlay with Edit and Delete buttons (currently the Edit button works)


js/views/userPage/UserPage.js
```diff
@@ -48,9 +50,12 @@ export default class extends baseVw {
...
     }
 
-    this.listenTo(this.model, 'change', () => this.render());
-    this.listenTo(this.model.get('avatarHashes'), 'change', () => this.render());
-    this.listenTo(this.model.get('headerHashes'), 'change', () => this.render());
+    // todo: This is way too elaborate of a view to re-render (at least at an unexpected / random
+    // time). When we need to update UI based off of async changes, we need to either display
+    // some indicator and let the user refresh -or- if the changes are simple enough, make them
+    // with jQuery. In this case, I think jQuery would be simple enough.
+    // this.listenTo(this.model.get('avatarHashes'), 'change', () => this.render());
```
rmisio: @jjeffryes Had to comment out this code. It does unpleasant things to the store and would to the other tabs as we start to build them out.

styles/components/_type.scss
```diff
@@ -41,6 +41,10 @@ h6, .h6, .tx6, .btn.tx6 {
   font-size: $txBg;
 }
 
+.txTiny {
```
jjeffryes: Can we stick with the current naming convention for Tiny and make this txTn?

In other places we're using the convention Tn, Sm, Md, Lg, Hg. 

A few places have Bg for Big instead of Lg for Large, that'll have to be cleaned up.

styles/modules/_userPage.scss
```diff
@@ -78,5 +60,152 @@
...
+                display: none;
+              }
+            }
+          }
+        }
+      }
+
+      .storeListings {
+        flex-shrink: 1;
+        width: 100%;
```
jjeffryes: Instead of flex-shrink 1 and width" 100%, you want this to be flex-grow 1, that will make it grow to fill the remaining space.

styles/modules/_userPage.scss
```diff
@@ -78,5 +60,152 @@
...
+
+        .listingsWrap {
+          flex-wrap: wrap;
+
+          &:empty {
+            & ~ .noResults {
```
jjeffryes: I noticed the html for no results is commented out. Is that temporary?

styles/modules/_userPage.scss
```diff
@@ -78,5 +60,152 @@
...
+
+          .col {
+            margin: 0 $padMd $padMd 0;
+            width: 225px;
+            height: 283px;
+            border-style: solid;
+            border-width: 1px;
+            font-size: $tx5b;
+            cursor: pointer;
+            position: relative;
```
jjeffryes: The .col base class already has position: relative, you can take it off here.

styles/modules/_userPage.scss
```diff
@@ -78,5 +60,152 @@
...
+          }
+        }
+
+        .noResults {
+          display: none;
+        }
+
+        .listingsHeaderRow {
+          .listingsCount {
+            flex-basis: 100%;
```
jjeffryes: This probably isn't the best way to do this.

It would be better to give this flex-grow: 1, and then wrap the next sibling div in a div, and remove the floR class from it.

The reason the sibling needed floR to collapse it is you can't have an element with display:flex as a direct child of a flex element, or it mucks up how the space is distributed.

js/templates/modals/editListing/editListing.html
```diff
@@ -86,16 +92,16 @@ <h1 class="h3 txUnb padMd js-listingHeading"><% print(ob.createMode ? ob.polyT('
                 <label>Mature Content (nsfw, adult, 18+, etc)</label>
                 <% if (ob.errors['listing.item.nsfw']) print(ob.formErrorTmpl({ errors: ob.errors['listing.item.nsfw'] })) %>
                 <div class="btnStrip marginTopAuto">
-                  <div class="btn clrBr">
+                  <div class="btnRadio clrBr">
                     <input type="radio"
                            name="listing.item.nsfw"
                            value="true"
-                           id="editListingNSFWInputTrue"
+                           id="moo"
```
rmisio: Just debugging something. I believe I fixed this in the more-store branch I'm working on now. Anyhow, I'll fix here so clicking on the checkbox label works correctly.

js/views/PageNav.js
```diff
@@ -207,13 +209,21 @@ export default class extends View {
 
   render() {
     let avatarHash = '';
-    if (isHiRez() && app.profile && app.profile.avatarHashes.small) {
-      avatarHash = app.profile.avatarHashes.small;
-    } else if (app.profile && app.profile.avatarHashes.tiny) {
-      avatarHash = app.profile.avatarHashes.tiny;
+
+    if (app.profile && isHiRez()) {
```
rmisio: This is fixed in the more-store branch, which is coming later. The reason it's not here is because at the time I wrote this, the branch with isHiRez function was not merge into Master yet.

js/utils/index.js
```diff
@@ -75,3 +76,41 @@ export function setDeepValue(obj, path, value) {
...
+
+// http://stackoverflow.com/a/2686098/632806
+// todo: unit test
+export function abbrNum(_number, _decPlaces = 1) {
+  // 2 decimal places => 100, 3 => 1000, etc
+  const decPlaces = Math.pow(10, _decPlaces);
+  let number = _number;
+
+  // Enumerate number abbreviations
+  const abbrev = ['k', 'm', 'b', 't'];
```
rmisio: I think bringing a lib is a bit overkill just for this. Anyhow, I've just made a commit (not yet pushed) where it will build the string off of a Polyglot translation, so our translators can localize the 'k', 'm'... part and I'm using the core Intl.NumberFormat to internationalize the actual number.

js/utils/string.js
```diff
@@ -0,0 +1,7 @@
+export function capitalize(str) {
+  if (typeof str !== 'string') {
+    throw new Error('Please provide a str as a String.');
+  }
+
+  return str.charAt(0).toUpperCase() + str.slice(1);
```
rmisio: At this point our string utils consists of a single capitalize function, so it's not looking like we need a full blow lib. I want to be mindful of bringing in libs since it impacts download time of the binary.

Anyhow, thanks for pointing out that lib. I'll keep it in mind and if we see that we are in need of more than a small number of functions that the lib provides, we could look into bringing it in.

js/views/userPage/Store.js
```diff
@@ -0,0 +1,404 @@
...
+    // make sure they're not still typing
+    if (this.searchKeyUpTimer) {
+      clearTimeout(this.searchKeyUpTimer);
+    }
+
+    this.searchKeyUpTimer = setTimeout(() =>
+      (this.search($(e.target).val())), 150);
+  }
+
+  onRequest(cl, xhr) {
```
rmisio: No, but the xhr is used, so I need to declare the cl param. These are provided by backbone in the 'request' event callback.

js/views/userPage/Store.js
```diff
@@ -0,0 +1,404 @@
...
+    this.fetch = xhr;
```
rmisio: Yes, because this is a callback that I'm setting to be called (via backbone) on the collection's 'request' event:

`this.listenTo(this.collection, 'request', this.onRequest);`

I just happen to also need to execute that code path if the view is loaded with a request already in progress (in which case the request event already fired before we bound our callback to it). So... I'm just mimicking what Backbone sends to that callback.

js/views/userPage/Store.js
```diff
@@ -0,0 +1,404 @@
...
+    this.setSortFunction(col);
+    col.sort();
+
+    if (!$.contains(this.$listingsContainer[0], this.storeListings.el)) {
+      this.$listingsContainer.empty()
+        .append(this.storeListings.el);
+    }
+
+    const listingCountContent =
+      `<span class="txB"><span>${col.length}</span> listing` +
```
rmisio: I'll be doing translations in one big sweep in the more-store branch later. On that PR, feel free to go all out with any untranslated strings. I did make a note on this string so I don't miss it.

Re: extra span - no clue why, I've removed it.

js/views/userPage/Store.js
```diff
@@ -0,0 +1,404 @@
...
+    this.$shipsToSelect.select2({
+      dropdownParent: this.$('.js-shipsToSelectDropdownContainer'),
+      // dropdownPosition : 'below',
+    });
+
+    if (!this.rendered && this.options.listing) {
+      // if first render, show a listing if it was
+      // passed in as a view option
+      this.showListing(this.options.listing);
+      this.rendered = true;
```
rmisio: Good catch, my `if` wasn't properly structured. I've fixed it.

styles/modules/_userPage.scss
rmisio: I have a note on my list to revisit the noResults output, which I'll do before I submit the more-store branch as a PR. A few things ended up changing that will likely require me to refactor how I go about indicating no results.

styles/modules/_userPage.scss
rmisio: Well, this falls into the category of, "I welcome PRs" (lol). It took me a good amount of tinkering to get it working correctly and, since it is working fine, I'd rather focus my efforts elsewhere.

PR #88, opened 2016-11-15
More store

rmisio:
This builds on #84 with more store features, notably:

- Ability to delete a listing, both from the listing card and the listing detail pages.
- A free shipping banner is added to the listing detail change and if the relevant listing / settings data changes, it is added / removed.
- As listing data changes, a pop in message is displayed on the store and listing detail modals recommending the user should refresh their view.
- Viewing store listing in "list" view. The user's preference for this is stored in localStorage.

Notable things left to do:
- Ability to view listings at 100% window width (initiated by toggling the 960px / 100% button at the top of the store). @jjeffryes will handle this since it will require some rearchitecting of existing foundational markup.
- Loading the listings in batches, so a.) the browser is not inundated with way too many image requests b.) painting the aggregate listings doesn't take too long (browser is locked during painting).
- A couple dozen + of relatively minor loose little odds and ends your truly has been keeping a list of.
- Wire in translations.
- Write unit tests.


js/startup/listingData.js
```diff
@@ -0,0 +1,304 @@
...
+// The main function is:
+// createDummyListings(count = 300)
+//
+// Which is used to bulk create listings. However, the server is
+// kicking off a publish on each listing created and if you send
+// too many, your computer might explode (my cpu somehow reached
+// 600% + when I called it with 150). So, I recommend you call
+// it with 5 listings and then not call again until publishing
+// has completed.
+
```
rmisio: This may be useful to you as well if you want to create some dummy listings. It will include some placeholder images, categories and tags.

styles/modules/_userPage.scss
```diff
@@ -115,96 +120,43 @@
         flex-shrink: 1;
         width: 100%;
 
-        .listingsWrap {
-          flex-wrap: wrap;
+        .listingsGrid {
+          // &:empty {
```
rmisio: No results handling will happen in a subsequent branch.

js/views/modals/listingDetail/Listing.js
```diff
@@ -56,27 +82,100 @@ export default class extends BaseModal {
...
+      }).always(() => {
+        if (!this.isRemoved()) {
+          this.$deleteListing.removeClass('processing');
+        }
+      });
+    }
+  }
+
+  showDataChangedMessage() {
+    if (this.dataChangePopIn || (this.dataChangePopIn && this.dataChangePopIn.isRemoved())) {
```
jjeffryes: If this.dataChangePopIn is true, the second condition isn't checked, but if it's false, then the second condition will also be false, won't it?

js/views/userPage/Store.js
```diff
@@ -32,15 +35,85 @@ export default class extends BaseVw {
...
+        this.dataChangePopIn = null;
+      });
+
+      this.$popInMessages.append(this.dataChangePopIn.render().el);
+    }
+  }
+
+  showShippingChangedMessage() {
+    if (this.shippingChangePopIn ||
+      (this.shippingChangePopIn && this.shippingChangePopIn.isRemoved())) {
```
jjeffryes: Similar comment here. When would this.shippingChangePopIn be false, but (this.shippingChangePopIn && this.shippingChangePopIn.isRemoved()) be true?

styles/components/_listingCard.scss
```diff
@@ -0,0 +1,155 @@
...
+      bottom: 0;
+    }
+  }  
+
+  .editOverlay {
+    .btn {
+      z-index: 1;
+      border-left-width: 0;
+
+      &:first-of-type {
```
jjeffryes: first-of-type! Very nice.

styles/components/_listingCard.scss
```diff
@@ -0,0 +1,155 @@
...
+    .listViewContent {
+      display: block;
+    }
+
+    .gridViewContent {
+      display: none;
+    }
+
+    // This isn't ideal, but there's inconistent spacing / widths of cols, plus there's variable
+    // content width on the price (e.g. some currencies may have quite large numbers)
```
jjeffryes: I can take a look here later, and see if there's another way to do it.

styles/modules/_userPage.scss
```diff
@@ -62,6 +62,11 @@
     }
 
     .userPageStore {
+      // Setting this to establish a z-index stacking context so
+      // that the pop-in messages are properly stacked.
+      // http://stackoverflow.com/a/16148341/632806
+      isolation: isolate;
```
jjeffryes: Another piece of css I've never seen in the wild before!

js/models/listing/index.js
```diff
@@ -0,0 +1,71 @@
...
+          .forEach(service => {
+            if (service.get('price') === 0) {
+              freeShipping = freeShipping.concat(shipOpt.get('regions'));
+            }
+          });
+      });
+  }
+
+  // countries may have dupes, but it's no bother for this purpose
+  return !!_.intersection(freeShipping, countries).length;
```
rmisio: Yup, I included this in the aforementioned [Trello card](https://trello.com/c/c64Z2vF5/215-integrate-shipping-regions-into-the-ui).

js/startup/listingDelete.js
```diff
@@ -0,0 +1,56 @@
...
+    model.get('title')
+);
+
+export default function () {
+  const statusMessages = {};
+
+  listingEvents.on('destroying', (model, opts) => {
+    let statusMessage = statusMessages[opts.slug];
+
+    if (statusMessage) return;
```
rmisio: Yeah, good catch. I forgot about that part. Should be fixed in a push coming soon.

js/views/modals/listingDetail/Listing.js
rmisio: good catch... will push fix soon.

PR #92, opened 2016-11-18
Completes TODO to update avatar and header on user page.

jjeffryes:


js/views/userPage/UserPage.js
```diff
@@ -91,6 +88,28 @@ export default class extends baseVw {
     this.$moreableBtns.toggleClass('hide');
   }
 
+  updateAvatar() {
+    const avatarHashes = this.model.get('avatarHashes').toJSON();
```
rmisio: No need to change this as it's arguably stylistic, but as an FYI... `avatarHashes` is a nested model, so you could also get its attributes using the standard model API (as opposed to having to flatten the object via toJSON()), e.g:

```
const headerHashes = this.model.get('headerHashes');
const headerHash = isHiRez() ? headerHashes.get('small') : headerHashes.get('tiny');
```

PR #97, opened 2016-11-23
Even more store

rmisio:
This branch brings in:
- Loading store items in batches so the browser is not inundated with image requests and having to paint a large amount of items to the DOM.
- Wires in translations.
- A plethora of loose little odds and ends, tweakaroos, etc...


js/utils/currency.js
```diff
@@ -115,3 +124,92 @@ export function formatCurrency(amount, currency, locale = app.settings.get('lang
 
   return formattedCurrency;
 }
+
+let exchangeRates = {};
+
+export function fetchExchangeRates(options = {}) {
```
rmisio: @jjeffryes You mentioned that you intended on fetching the exchange rates at order time to ensure the most accurate price. A couple of thoughts on that:

1.) Currently, we're fetching the exchange rates every 15 seconds (which is also how often the server fetches them), so a fetch-on-demand is at best going to get you rates that are 14.9999 seconds fresher than the cached ones available via getExchangeRate(). And, since our interval of fetching will not be in sync with the servers, it is likely the cached ones are even fresher than that and a re-fetch may return you the exact same values as the cached one.

So, I'm not sure how critical it is to fetch-on-demand.

2.) If you do decide to carry on and fetch-on-demand, I would recommend you do so via fetchExchangeRates() (and then convertCurrency() and/or convertAndFormatCurrency()). That will fetch all the exchange rates, however, that's probably more efficient than calling the api for individual rates, since if you need to convert between two fiat values, you would need to make two API calls which, with all things considered, will be heavier than calling the single bulkier one.

test/utils/currency.js
```diff
@@ -148,4 +150,89 @@ describe('the currency utility module', () => {
         .equal('฿523.12000000');
     });
   });
+
+  describe('has functions that involve converting between currencies', () => {
+    let ajax;
+
+    before(function () {
+      ajax = sinon.stub($, 'ajax', () => {
```
rmisio: Yeah, cool stuff here... using Sinon to stub ajax, so when I call fetchExchangeRates(), it doesn't blow up without a server, it calls the stubbed function which I have set to return some dummy data.

js/views/userPage/Store.js
```diff
@@ -244,14 +201,64 @@ export default class extends BaseVw {
...
+        this.dataChangePopIn = null;
+      });
+
+      this.$popInMessages.append(this.dataChangePopIn.render().el);
+    }
+  }
+
+  showShippingChangedMessage() {
+    if (this.shippingChangePopIn ||
+      (this.shippingChangePopIn && this.shippingChangePopIn.isRemoved())) {
```
jjeffryes: If this.shippingChangePopIn is false, wouldn't (this.shippingChangePopIn && this.shippingChangePopIn.isRemoved()) also always be false, meaning the second condition is never evaluated?

js/utils/currency.js
```diff
@@ -115,3 +128,105 @@ export function formatCurrency(amount, currency, locale = app.settings.get('lang
...
+  }
+
+  if (!exchangeRates[toCur]) {
+    throw new NoExchangeRateDataError(`We do not have exchange rate data for ${toCur}.`);
+  }
+
+  const fromRate = fromCur === 'BTC' ? 1 : getExchangeRate(fromCur);
+  const toRate = toCur === 'BTC' ? 1 : getExchangeRate(toCur);
+
+  return (amount / fromRate) * toRate;
```
rmisio: When I check the amounts using the app and Google's currency conversion calculator, things check out.

Anyhow, right now our api is returning the following rates:
"ISK": 83179.69,
"USD": 739.07,

So, using the formula in the code, $5 USD would result in 562.73 ISK, which I believe is right.

js/views/StatusMessage.js
```diff
@@ -11,6 +12,28 @@ export default class extends baseVw {
     this.listenTo(this.model, 'change', this.render);
   }
 
+  events() {
+    return {
+      'click [class^="js-"], [class*=" js-"]': 'onClick',
```
rmisio: The second one is prefaced with a space, so they are targeting two different things. For example, if the class attribute were `class="js-alpha sugar-time js-beta foo-bar-js-moo"`, the first one would target `js-alpha` whereas the second one would target `js-beta`.

If we just used the second one without the leading space, then it would correctly target the first one, but it would also target `foo-bar-js-moo`, which is not what we want.

js/utils/currency.js
jjeffryes: Ah. My fault. I assumed the rate would be the ratio of 1 fiat to bitcoin, not the value of one bitcoin in the fiat price. 

That shows me to not do math before running the code.

js/views/StatusMessage.js
jjeffryes: I see. I missed the space.

PR #108, opened 2016-11-30
Fix Linux scaling

### LINKED ISSUE #79: Add Zoom Handling for Linux
On startup, use electron.screen.getPrimaryDisplay().scaleFactor to adjust the scaling for Linux.

Something like the below should work:

if (platform === "linux") {
  var scaleFactor = require('electron').screen.getPrimaryDisplay().scaleFactor || 1;
  $("body").css("zoom", 1 / scaleFactor);
}

hoffmabc:
Fixes #79

### DIFF
--- a/js/start.js
+++ b/js/start.js
@@ -1,4 +1,4 @@
-import { ipcRenderer } from 'electron';
+import { ipcRenderer, screen } from 'electron';
 import $ from 'jquery';
 import Backbone from 'backbone';
 import Polyglot from 'node-polyglot';
@@ -22,6 +22,8 @@ import './utils/listingData';
 
 app.localSettings = new LocalSettings({ id: 1 });
 app.localSettings.fetch().fail(() => app.localSettings.save());
+
+const platform = process.platform;
 
 // initialize language functionality
 function getValidLanguage(lang) {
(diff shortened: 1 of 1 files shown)


js/start.js
```diff
@@ -74,6 +76,15 @@ app.loadingModal = new LoadingModal({
   removeOnRoute: false,
 }).render().open();
 
+// fix zoom issue on Linux hiDPI
+if (platform === 'linux') {
+  let scaleFactor = screen.getPrimaryDisplay().scaleFactor;
+  if (scaleFactor === 0) {
+    scaleFactor = 1;
+  }
+  $('body').css('zoom', 1 / scaleFactor);
```
rmisio: Could we please use getBody from the utils/selectors.js file? It will cache that element so we only need to go to the DOM once for it.

hoffmabc:
Ok @rmisio how's this look?

js/start.js
```diff
@@ -18,11 +18,13 @@ import Followers from './collections/Followers';
 import listingDeleteHandler from './startup/listingDelete';
 import { fetchExchangeRates } from './utils/currency';
 import './utils/exchangeRateSyncer';
-import './utils/listingData';
+import { getBody } from './utils/listingData';
```
rmisio: getBody() is in utils/selectors.js. Did this code work when you ran it, or are you cowboy dev'in? lol

Issue #94, opened 2016-11-21
Advanced Settings


jjeffryes:
The existing settings modal should have an advanced tab with settings that are technical and/or less frequently used.

## Functional Requirements
The following settings should be available for the user:
**Appearance**
- [ ] Radio-style buttons to select the window controls style (Mac/Windows).*
- [ ] Radio-style buttons to turn advanced visual effects on or off.*
- [ ] **Transactions**
- [ ] Radio-style buttons to select whether additional data is saved to the wallet for each purchase.*
- [ ] Radio-style buttons to select a low, medium, or high default fee.*

**Server**
- [ ] Text displaying the current firewall type. 
- [ ] A button to open the connection management modal.
- [ ] A button to show a list of the connected peers. The list should show each GUID, clicking on one should navigate to that user's user page.
- [ ] A button to turn off the server.
- [ ] If the server is the local default server installed with the bundled app, a button to turn on the server.

**SMTP Integration**
- [ ] Radio-style button to turn SMTP notifications on or off.
- [ ] Text input for SMTP server address.**
- [ ] Text input for SMTP user name.
- [ ] Text input for SMTP password.
- [ ] Text input for SMTP email address. **
- [ ] Text input for SMTP recipient email address.**
- [ ] If the SMTP notifications are turned on, all the other SMTP fields must be filled out or an error should be thrown when the settings are saved.
- [ ] A button to test the SMTP settings. It will open a connection to the SMTP server using the user name and password in the inputs, and show a success/fail message.

* these settings are saved in local storage.
** these settings should be validated to make sure they have the right format.

## Data
- the current firewall type and SMTP settings should be retrieved from the settings API.
- the SMTP settings should be saved to the settings API.

rmisio:
> the current firewall type and SMTP settings should be retrieved from the settings API.

The settings API is fetched on app startup and need not be fetched again. The data can be retrieved from the model available via `app.settings`.

> these settings are saved in local storage.

Please do so using the localSettings model which is available via `app.localSettings` and, like the Settings model, it is fetched at startup, so you do not need to fetch it.

> these settings should be validated to make sure they have the right format.

Validations should be integrated into their respective models (app.settings or app.localSettings)

>  Radio-style buttons to select the window controls style (Mac/Windows).

This is a matter of wiring in the radio button to the app.localSettings.macStyleWinControls field.

jjeffryes:
[...]

PR #111, opened 2016-12-01
Advanced settings tab issues #94

tenthhuman:
### Complete :+1: :+1: 

- 4 sections with required inputs:
Appearance, Transactions, Server and SMTP Integration

### Excepting :ghost: 

- **Validation** : Not clear how best to do validation of nested models.
- **Local settings** : Splitting some of SMTP integration's settings into local settings
and some into the settings api store. It is unclear why and the best
way to do this.
- **API format** : `./js/utils/settingsAPIAdaptor.js` contains an explanation of the
difference between the API format and the model format and especially how
3 keys are omitted from the api, being those for Appearance, Transactions
and Server settings. Requesting an API extension to permit storing these keys
is proposed.

### Other Issues :strawberry: 

- **Newness** : There is still much I do not know about how data flows through this application.
- **Nesting** : There was a strange bug discovered where, the code that attempts to facilitate nesting models,
also ends up deleting the keys of nested models. Existing code, like avatarHashes depends on this,
and the best way to really implement nested models in a maintainable architecture is something it works
to get some clarity around. 
The existing model nesting code appears to be a "partial solution" to the
issue of nesting models. Possible forward paths involve having separate APIs for all models and "denormalizing" models ( ShippingAddress, avatar ) so there is no nesting. This is the least preferred option. 
Another option is to build ( or use ) a complete solution for nesting Models in backbone. One that brings clarity to all things like syncing, change events, and so on. This is the preferred option


JustinDrake:
Looks like you took this pull request seriously @tenthhuman. 👍 Welcome to OpenBazaar! Are you on the OpenBazaar Slack? http://slack.openbazaar.org

jjeffryes:
@JustinDrake Justin meet Cris. He's the newest member of the OB1 team.

js/templates/modals/settings/advanced/transactions.html
```diff
@@ -0,0 +1,68 @@
+<form class="box padKids padStack clrP clrBr js-transactions">
+  <div class="flexRow">
+    <h3><%= ob.polyT('settings.transactions') %></h3>
+  </div>
+  <div class="flexRow">
+    <div class="col3">
+      <label><%= ob.polyT('settings.saveMetadata') %></label>
+    </div>
+    <div class="col9">
+      <div class="btnStrip">
```
jjeffryes: Just noting the btnStrip isn't needed here.

js/templates/modals/settings/advanced/transactions.html
```diff
@@ -0,0 +1,68 @@
...
+        </div>
+      </div>
+    </div>
+  </div>
+  <div class="flexRow">
+    <div class="col3">
+      <label><%= ob.polyT('settings.defaultFee') %></label>
+    </div>
+    <div class="col9">
+      <div class="btnStrip">
```
jjeffryes: Just noting the btnStrip isn't needed here.

js/views/modals/Settings/Advanced/Appearance.js
```diff
@@ -0,0 +1,37 @@
+import app from '../../../../app';
+import loadTemplate from '../../../../utils/loadTemplate';
+import baseVw from '../../../baseVw';
+
+export default class extends baseVw {
+  constructor( options = { } ) {
+    super( options );
+
+    this.options = options;
+    this.settings = app.settings.clone( );
```
jjeffryes: To handle the appearance, you should get both the parameters from local settings, using:

```
app.localSettings.get('macStyleWinControls')
app.localSettings.get('visualEffect') //this doesn't exist yet
```

When the form is saved, you should set them both using localSettings. 

The pageNav view has a listener that will make the style changes when the local settings for macStyleWinControls is changed.

We'll add something similar later for visualEffects, the class to deactivate the visual effects doesn't exist yet.

js/views/modals/Settings/Advanced/Appearance.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  render( ) {
+    this.settings = app.settings.clone( );
+    const mySettings = this.settings.toModelFormatJSON().appearanceSettings;
+
+    if ( mySettings.windowControlStyle == 'mac' )
+      app.localSettings.set( { 'macStyleWinControls' : true } );
+    else
+      app.localSettings.set( { 'macStyleWinControls' : false } );
```
jjeffryes: We won't need this section, since the controls style is already set.

js/views/modals/Settings/Advanced/Settings.js
```diff
@@ -0,0 +1,108 @@
...
+        .fail((...args) => {
+          this.trigger('saveComplete', false, true,
+            args[0] && args[0].responseJSON && args[0].responseJSON.reason || '');
+          this.render();
+        });
+    }
+  }
+
+  render() {
+    loadTemplate('modals/settings/advanced/settings.html', (t) => {
```
jjeffryes: It would be better to load the pieces of the form with nested loadTemplates, so the DOM doesn't need to be traversed and modified after it's rendered. See the userPage home view for an example.

js/models/Settings.js
```diff
@@ -26,10 +30,40 @@ export default class extends BaseModel {
   nested() {
     return {
       shippingAddresses: ShippingAddresses,
-      smtpSettings: SMTPSettings,
```
rmisio: Why was this renamed to 'smtpIntegrationSettings'? It should match what the server calls it which is 'smtpSettings', that way when the model is fetched the right nested instance will be set based off the data returned from the server.

js/models/Settings.js
```diff
@@ -26,10 +30,40 @@ export default class extends BaseModel {
...
+      appearanceSettings: AppearanceSettings,
+      transactionSettings: TransactionSettings,
```
rmisio: I'm not seeing transactionSettings in the Settings API:
https://github.com/OpenBazaar/openbazaar-go/blob/master/repo/settings.go

Is that something that still has to be done on the back-end? If so, please keep in mind that it only needs to be a nested model is the server has it as nested data. If the server has the properties with the TransactionSettings model as non-nested attributes, then we should follow suit and integrate them directly into the Settings model (no nesting).

js/models/Settings.js
```diff
@@ -26,10 +30,40 @@ export default class extends BaseModel {
   nested() {
     return {
       shippingAddresses: ShippingAddresses,
-      smtpSettings: SMTPSettings,
+      appearanceSettings: AppearanceSettings,
```
rmisio: As we discussed, these are local settings for the localSettings model and I don't see a need for them to be nested.

js/templates/modals/settings/advanced/smtpintegration.html
```diff
@@ -0,0 +1,120 @@
...
+  <div class="flexRow">
+    <div class="col3">
+      <label for="smtpServerAddress">
+        <%= ob.polyT('settings.smtpServerAddress') %>
+      </label>
+    </div>
+    <div class="col9">
+      <input class="clrBr" 
+        required
+        pattern=".*"
```
rmisio: Re: validation. This is done at the template level. It's a matter of:

1.) Pass in you model errors when rendering the template:

```
      this.$el.html(t({
        errors: this.model.validationError || {},
      });
```

Any nested attributes will be included in the validationError object, but they'll be flattened, for example:

```
{
  email: 'bull-shiznit-email',
  nestedModel: {
    email: 'crap-on-a-stick-yo',
  }
}
```

would result in the following error object:

```
{
  email: ['The email is not valid'],  // errors are returned in an array because a field could fail multiple          validations
  nestedModel.email: ['The email is not valid'],
}
```

2.) If a field has an error, render the `formErrorTmpl` (it's by default going to available in your template) above it's corresponding field, e,g:

```
        <% if (ob.errors.email) print(ob.formErrorTmpl({ errors: ob.errors.email })) %>
        <input type="text" name="email" value="<%= ob.email %>">

        <section>  /// nested model
          <% if (ob.errors['nestedModel.email']) print(ob.formErrorTmpl({ errors: ob.errors['nestedModel.email'] })) %>
          <input type="text" name="nestedModel.email" value="<%= ob.nestedModel.email %>">
       </section>
```

3.) Make sure that at the end of you're save() method you render() (this way any new errors will be displayed and any old ones will be cleared), e.g:

```
    this.render();
    // next two lines will scroll to the first error
    const $firstErr = this.$('.errorList:first'); 
    if ($firstErr.length) $firstErr[0].scrollIntoViewIfNeeded();
```

Please see Page.js and / or EditListing.js (and their templates) as an example (FWIW, I would start with Page.js, since EditListing.js is much more complex (very deeply nested)).

js/views/modals/Settings/Advanced/Appearance.js
```diff
@@ -0,0 +1,37 @@
+import app from '../../../../app';
```
rmisio: Is it really necessary to make separate views for all these sections? I'm all for modularity, but unless you think these sections will be re-used somewhere else in the app (which I really doubt), I think it's very much overkill here. I would make this all one Advanced view (look at Page.js as an example).

When you toJSON() the settings or localSettings model, it will include and flatten any nested attributes, so they could easily be handled in one view. Similary, getFormData() will properly handle nested attributes for you as long as you use the proper naming convention (namely giving them a flattened name attribute, e.g. `name="nestedAttribute.attributeName"`.

js/views/modals/Settings/Advanced/Settings.js
```diff
@@ -0,0 +1,108 @@
...
+    super({
+      className: 'settingsAdvanced',
+      ...options
+    });
+
+    this.options = options;
+    this.settings = app.settings.clone( );
+
+
+    self.app = app;
```
rmisio: What is the purpose of this line?

js/views/modals/Settings/Advanced/Settings.js
```diff
@@ -0,0 +1,108 @@
...
+
+    this.listenTo(this.settings, 'sync', () => app.settings.set(this.settings.toModelFormatJSON()));
+    this.listenTo(this.settings, 'save', () => app.settings.set(this.settings.toModelFormatJSON()));
```
rmisio: There is no 'save' event on models.

We have made a decision to only update global models (like app.settings) upon 'sync'. The idea is that we only want them updated once the data has successfully been updated on the server. We wouldn't want to update them before that, because if the call fails on the server, we'd have to revert, which could be a bit hairy since, in theory, there can be more than one pending Save request.

js/views/modals/Settings/Advanced/Server.js
```diff
@@ -0,0 +1,30 @@
+import app from '../../../../app';
```
rmisio: I have some comments for these individual views, but I'm holding them back because, as I mentioned above, I strongly recommend they be consolidated into one Advanced view.

js/views/modals/Settings/Advanced/Settings.js
rmisio: I don't think that would work for nested views... but, hopefully those will go away :)

js/models/AppearanceSettings.js
```diff
@@ -0,0 +1,10 @@
+import BaseModel from './BaseModel';
+
+export default class extends BaseModel {
+  defaults() {
+    return {
+      showAdvancedVisualEffects : true,
+      windowControlStyle : 'mac'
```
rmisio: @jjeffryes ok, I see you touched on a lot of my points above in some comments below.

js/views/modals/Settings/Advanced/Appearance.js
rmisio: In either case, setting a model should almost never happen in render. It should really happen in save() and can be as simple as passing the results of getFormData() into Model.save().

It's a little more complex if you happen to have a form that's driven by multiple models. In that case, you'd probably want to add an attribute to each form field indicating which model it belongs to. Then you could call getFormData() with different selectors (one for each form) and save each form individually.

It's also a little more complex because in your view's save, you'd need to manage a save call for each model and the fact that one could succeed whereas the other could fail.

rmisio:
@tenthhuman You have 316 lint errors. Would you please fix them? (warnings are ok)

```
npm run lint
```

js/models/SMTPIntegrationSettings.js
```diff
@@ -0,0 +1,43 @@
...
+      errObj[fieldName].push(error);
+    };
+
+    if (attrs.smtpNotifications) {
+      if ( attrs.smtpServerAddress.trim().length == 0 )
+        addError( 'smtpServerAddress', app.polyT( 'smtpIntegrationModelErrors.smtpServerAddress' ) );
+      if ( attrs.smtpUserName.trim().length == 0 )
+        addError( 'smtpUserName', app.polyT( 'smtpIntegrationModelErrors.smtpUserName' ) );
+      if ( attrs.smtpPassword.trim().length == 0 )
+        addError( 'smtpPassword', app.polyT( 'smtpIntegrationModelErrors.smtpPassword' ) );
```
rmisio: None of these translation appear to be in the translations file.

js/templates/modals/settings/advanced/server.html
```diff
@@ -0,0 +1,45 @@
...
+    </div>
+  </div>
+  <div class="flexRow">
+    <div class="col3">
+      <label>
+        <%= ob.polyT('settings.onOff') %>
+      </label>
+    </div>
+    <div class="col9">
+      <div class="btnStrip flex clrP clrBr">
```
morebrownies: If not, we could ditch the Turn Server On button.

jjeffryes: Yes, you can have the client open without being connected to a server, that's how you connect to remote servers.

morebrownies:
I must be looking at the wrong screen. Can you post a screenshot?

jjeffryes:
@morebrownies here you go.

image

.gitignore
```diff
@@ -1,5 +1,5 @@
 node_modules
-npm-debug.log
+npm-debug.log*
```
dosaygo-coder-0: I get numbered ones.

js/views/modals/Settings/Advanced/Settings.js
dosaygo-coder-0: @jjeffryes I don't know what you mean by nested loadTemplates. Could you give an example?

@rmisio are you indicating a preference to have a single advanced settings view rather than split into separate views for each section?

js/views/modals/Settings/Advanced/Settings.js
tenthhuman: @rmisio debug instrumentation. So I could check what was going on with app settings in console.

js/views/modals/Settings/Advanced/Settings.js
tenthhuman: @jjeffryes I don't know what you mean by nested loadTemplates. Could you give an example?

.gitignore
tenthhuman: I get numbered ones sometimes. Similar to `npm-debug.log2384723`

js/models/BaseModel.js
```diff
@@ -213,7 +227,11 @@ export default class extends Model {
   }
 
   clone() {
-    const clone = new this.constructor(this.toJSON(), { guid: this.guid });
+    let clone;
```
tenthhuman: @rmisio this key deletion thing was happening when I started to use nested models. `toModelFormatJSON` is part of the monkey patch to get around it. Without the monkey patching, the advanced settings model gets serialized to `{}`. Either the nested models code was incomplete or I do not know how to use it correctly. More likely the latter, tho my feeling is it's the former. I see that it would have worked for me to trigger the PR earlier in the process -- rather than trying to monkey patch my way through. @jjeffryes and @rmisio are being very helpful, I really appreciate that and wasn't necessarily counting on it. So next time, I think, I'll trigger a PR earlier if I meet something that isn't so clear.

js/models/SMTPIntegrationSettings.js
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  validate(attrs) {
+    const errObj = {};
+    const addError = (fieldName, error) => {
+      errObj[fieldName] = errObj[fieldName] || [];
+      errObj[fieldName].push(error);
+    };
+
+    if (attrs.smtpNotifications) {
```
tenthhuman: @rmisio okay. I think the spec was to only save the fields if notifications are on. Tho maybe that doesn't make so much sense as what you just said. Agree validating anyway makes sense.

js/models/SMTPIntegrationSettings.js
```diff
@@ -0,0 +1,43 @@
...
+      if ( attrs.smtpServerAddress.trim().length == 0 )
```
tenthhuman: @rmisio @jjeffryes of course. Also, I have not run the linter and I intend to.

js/models/SMTPIntegrationSettings.js
```diff
@@ -0,0 +1,43 @@
...
+        addError( 'smtpServerAddress', app.polyT( 'smtpIntegrationModelErrors.smtpServerAddress' ) );
```
tenthhuman: Okay

js/models/SMTPIntegrationSettings.js
```diff
@@ -0,0 +1,43 @@
...
+      if ( attrs.smtpUserName.trim().length == 0 )
+        addError( 'smtpUserName', app.polyT( 'smtpIntegrationModelErrors.smtpUserName' ) );
+      if ( attrs.smtpPassword.trim().length == 0 )
+        addError( 'smtpPassword', app.polyT( 'smtpIntegrationModelErrors.smtpPassword' ) );
+      if ( attrs.smtpFromEmail.trim().length == 0 )
+        addError( 'smtpFromEmail', app.polyT( 'smtpIntegrationModelErrors.smtpFromEmail' ) );
```
tenthhuman: Okay

js/models/SMTPIntegrationSettings.js
```diff
@@ -0,0 +1,43 @@
...
+      if ( attrs.smtpToEmail.trim().length == 0 )
+        addError( 'smtpToEmail', app.polyT( 'smtpIntegrationModelErrors.smtpToEmail' ) );
+    }
+
+    if (Object.keys(errObj).length) {
+      console.warn( errObj );
```
tenthhuman: Okay

js/models/Settings.js
tenthhuman: I don't think there are currently any transaction related keys in the settings API. I'm intending to file an issue with a few additional keys required for the settings API.

js/models/Settings.js
```diff
@@ -26,10 +30,40 @@ export default class extends BaseModel {
...
+      transactionSettings: TransactionSettings,
+      serverSettings: ServerSettings,
+      smtpIntegrationSettings: SMTPIntegrationSettings
     };
   }
 
+  parse( response ) {
+    const data = this.fromAPIFormatJSON( response );
+    // make it immutable to highlight unintended modification errors
```
tenthhuman: It was instrumentation and to try to monkey patch the key deletion that was happening with nested models.

js/models/ShippingAddress.js
```diff
@@ -2,7 +2,7 @@ import BaseModel from './BaseModel';
 import app from '../app';
 import is from 'is_js';
 
-export default class extends BaseModel {
+export default class ShippingAddress extends BaseModel {
```
tenthhuman: It was clear. It was instrumentation. I was seeing if Backbone would read and report the model class name so I could see what NestedClass type name was. Turns out Backbone doesn't keep the names.

js/templates/modals/settings/advanced/appearance.html
```diff
@@ -0,0 +1,59 @@
+<form class="box padKids padStack clrP clrBr js-appearance">
+  <div class="flexRow">
+    <h3><%= ob.polyT('settings.appearance') %></h3>
+  </div>
+  <div class="flexRow">
+    <div class="col3">
+      <label><%= ob.polyT('settings.advancedVisualEffects') %></label>
+    </div>
+    <div class="col9">
+      <div class="btnStrip">
```
tenthhuman: Okay

js/templates/modals/settings/advanced/appearance.html
```diff
@@ -0,0 +1,59 @@
...
+        </div>
+      </div>
+    </div>
+  </div>
+  <div class="flexRow">
+    <div class="col3">
+      <label><%= ob.polyT('settings.windowControlStyle') %></label>
+    </div>
+    <div class="col9">
+      <div class="btnStrip">
```
tenthhuman: Okay

js/templates/modals/settings/advanced/server.html
```diff
@@ -0,0 +1,45 @@
...
+  <div class="flexRow">
+    <h3><%= ob.polyT('settings.server') %></h3>
+  </div>
+  <div class="flexRow">
+    <div class="col3">
+      <label for="currentFirewallType">
+        <%= ob.polyT('settings.currentFirewallType') %>
+      </label>
+    </div>
+    <div class="col9">
```
tenthhuman: Okay. I think semantics-wise it makes sense to be a readonly form slot if it comes from the same model as the other slots in the form. Also, I get that semantics of rendered HTML is not a main concern here. In any case, I wasn't that clear on what the firewall requirement meant. And now I see mentioned above maybe a firewall slot is being removed.

js/templates/modals/settings/advanced/server.html
```diff
@@ -0,0 +1,45 @@
...
+    </div>
+  </div>
+  <div class="flexRow">
+    <div class="col3">
+      <label>
+        <%= ob.polyT('settings.connections') %>
+      </label>
+    </div>
+    <div class="col9">
+      <div class="btnStrip flex clrP clrBr">
```
tenthhuman: Okay

js/templates/modals/settings/advanced/smtpintegration.html
```diff
@@ -0,0 +1,120 @@
+<form class="box padKids padStack clrP clrBr js-smtpintegration">
+  <div class="flexRow">
+    <h3><%= ob.polyT('settings.smtpIntegration') %></h3>
+  </div>
+  <div class="flexRow">
+    <div class="col3">
+      <label><%= ob.polyT('settings.smtpNotifications') %></label>
+    </div>
+    <div class="col9">
+      <div class="btnStrip">
```
tenthhuman: Okay

js/templates/modals/settings/advanced/smtpintegration.html
tenthhuman: Okay. Thanks a lot @rmisio for the in depth pointers, really appreciate.

js/templates/modals/settings/advanced/transactions.html
tenthhuman: Okay

js/templates/modals/settings/advanced/transactions.html
tenthhuman: Okay

js/utils/settingsAPIAdaptor.js
```diff
@@ -0,0 +1,201 @@
...
+  transactionSettings,
+  serverSettings
+
+  Of these 1 key, 'smtpIntegrationSettings', is substantially similar
+  to an existing key, 'smtpSettings'. 
+
+  So the proposal for the adaptor is:
+
+  - convert smtpIntegrationSettings to smtpSettings format.
+  - convert showNotifications to showNotificaton
```
tenthhuman: I totally agree an adapter is not what is wanted here as well. The adapter was just a way to get the functionality working and demonstrate that maybe some parts of the API could change, and that some parts of the API and requirements were not the same. I wasn't going to assume that "smtpSettings" was lore, when the spec read "SMTP Integration" settings. Seems tho that it also would work if I sought clarification from someone on the name earlier on.

js/utils/settingsAPIAdaptor.js
```diff
@@ -0,0 +1,201 @@
...
+
+  And the proposal for the requests API changes are:
+
+  request spelling correction to showNotifications
+  request 3 new keys be added, namely:
+
+  appearanceSettings,
```
tenthhuman: Okay

js/views/modals/Settings/Advanced/Appearance.js
tenthhuman: Okay. It seems the consensus from @jjeffryes and @rmisio is aggregate the views. And this approach with separating the views ran into some issues with nested models and so on. Seriously considering just making 1 view. It's a good point about the flattened names. Thanks @rmisio.

js/views/modals/Settings/Advanced/Appearance.js
tenthhuman: @rmisio agree that setting in the render was not something that works that well. I just wanted to get the functionality of the save working quickly to see it all come together, and didn't consider putting it somewhere else for that.

js/views/modals/Settings/Advanced/Settings.js
tenthhuman: Okay. I was feeling annoying when I was clicking save and not seeing the view update, even tho the settings were being sent to the server and saved there. Tho I think since it is a localhost API the lack of update must have been because something else wasn't being triggered correctly to render them. This 'save' event was more monkey patching attempt to get everything working together. I didn't actually believe there was a 'save' event, I was experimenting.

js/models/AppearanceSettings.js
tenthhuman: So I'll use app.localSettings for this.

js/models/BaseModel.js
```diff
@@ -92,7 +92,13 @@ export default class extends Model {
     this.lastSyncedAttrs = {};
 
     this.on('sync', () => {
-      this.lastSyncedAttrs = JSON.parse(JSON.stringify(this.toJSON()));
+      //FIXME : this is a hackish solution
+      //since the superclass has knowledge of a derived class's
+      //more specific methods
+      if ( this.toModelFormatJSON instanceof Function )
```
tenthhuman: Agree

js/models/BaseModel.js
```diff
@@ -116,8 +122,16 @@ export default class extends Model {
         const nestedInstance = this.attributes[nestedKey];
 
         if (nestedInstance) {
-          if (nestedData) nestedInstance.set(nestedData);
-          delete attrs[nestedKey];
+          try {
```
tenthhuman: Improperly set how? AFAIK nested is just a getter returning an object that maps nested key names to class constructors, which is implemented.

I see I used `nested` as a method which is how it is in `Profile.js`. But actually in `listings/*.js` models `nested` is a getter. I think I copied the usage which is a typo and then monkey patched all around that incorrect usage to try to get it to work. I'm going to try and test changing both `Profile.js` and advanced settings usage to the getter usage of nested. I'm very hopeful this might be the cause of the weird bug I saw.

js/templates/modals/settings/advanced/server.html
rmisio: Actually, Mike is kind of right. If the local server dies and you are connect to it, the Connection modal will pop up, so you won't be able to get to settings.

The only time you would be able to get to settings when the local server is down is if you are connected to a remote server... but, in that case, why would you want to start the local server? If you want to connect to the local server, you would just use the connection modal (or page nav connection menu) and connect to the default server... which would automatically start the local server.

I really think the Turn Server On button is not necessary.

js/views/modals/Settings/Advanced/Appearance.js
rmisio: Actually, I don't think you would get the values in the way Josh described. Likely, you would just pass the model's toJSON() output as context to your template and then right in your template (`ob. macStyleWinControls`) you would have those values and could set the fields based off of them.

But, yeah, if for some reason in you're view you need to programmatically access the values, then it would be via app.localSettings.get().

.gitignore
rmisio: ah, cool.. then it shall remain!

js/models/BaseModel.js
rmisio: hmm... I think it should work whether it's a getter or a function that returns an object, because the base model is accessing it via _.result() (`const nested = _.result(this, 'nested', []);`).

Yeah, if you find one way is not working, then please let me know... but Profile and the listing models seem to work and they each use different ways.

Anyhow... it's probably best, at least initially, to check in with us (or me for the nested model stuff) before tweaking some of the foundational classes too much. I understand with the timezone differences that might not be super easy.

js/models/SMTPIntegrationSettings.js
rmisio: For whatever code editor you are using, there is almost certainly an es-lint plugin which would highlight issues right there in the editor. Josh and I both use one and I've found it to be super helpful.

js/models/ShippingAddress.js
rmisio: In either case it's a problematic approach because if we ever obfuscated the code during the build process, the class name would change.

js/views/modals/Settings/Advanced/Settings.js
jjeffryes: Yes, like that, but not needed if you switch to one view/template.

If you had one view but multiple templates, you could still do it that way, though I don't think multiple templates are needed with one view in this case.

js/views/modals/Settings/Advanced.js
```diff
@@ -0,0 +1,90 @@
...
+  render() {
+    loadTemplate('modals/settings/advanced.html', (t) => {
+      this.$el.html(t({
+        errors: this.settings.validationError || {},
+        ...this.settings.toJSON(),
+        ...app.localSettings.toJSON()
+      }));
+
+      this.$formFields = this.$('select[name], input[name], textarea[name]').not('[persist="local"]');
+      this.$localFields = this.$('[persist="local"]');
```
jjeffryes: The persist attribute is a neat way to handle local vs. server fields, but I think it would be better as a data attribute, just so it's clear what the purpose is.

package.json
```diff
@@ -45,7 +45,7 @@
     "eslint-config-airbnb": "^9.0.1",
     "eslint-config-defaults": "^9.0.0",
     "eslint-plugin-import": "^1.10.0",
-    "eslint-plugin-jsx-a11y": "^1.5.3",
+    "eslint-plugin-jsx-a11y": "^3.0.1",
```
jjeffryes: Why did this change?

js/models/Settings.js
```diff
@@ -23,13 +23,21 @@ export default class extends BaseModel {
     return app.getServerUrl('ob/settings/');
   }
 
-  nested() {
+  get nested() {
```
rmisio: It shouldn't and doesn't when I try:

image

js/start.js
```diff
@@ -24,6 +24,7 @@ import { getBody } from './utils/selectors';
 app.localSettings = new LocalSettings({ id: 1 });
 app.localSettings.fetch().fail(() => app.localSettings.save());
 
+
```
rmisio: Could we remove this extraneous line please?

js/views/PageNav.js
```diff
@@ -36,9 +36,8 @@ export default class extends View {
 
     $(document).on('click', this.onDocClick.bind(this));
 
-    this.listenTo(app.localSettings, 'change:macStyleWinControls',
-      this.onWinControlsStyleChange);
-    this.setWinControlsStyle(app.localSettings.get('macStyleWinControls') ? 'mac' : 'win');
+    app.localSettings.on( 'change:windowControlStyle', (_,style) => this.setWinControlsStyle(style));
```
rmisio: We should be using `listenTo` here, not `on`. `listenTo` gives you the added benefit that when this view is removed, the handler will be unbound, this preventing our removed view remaining as a ghost view because a bound handler has a reference to it.

js/models/LocalSettings.js
```diff
@@ -16,7 +18,10 @@ export default class extends Model {
 
   defaults() {
     return {
-      macStyleWinControls: remote.process.platform === 'darwin',
+      windowControlStyle: remote.process.platform === "darwin" ? "mac" : "win",
```
tenthhuman: Cool, okay.

js/models/LocalSettings.js
```diff
@@ -16,7 +18,10 @@ export default class extends Model {
...
+      showAdvancedVisualEffects : true,
+      saveTransactionMetadata : true,
+      defaultTransactionFee : 'high',
```
tenthhuman: Cool, okay.

js/models/LocalSettings.js
```diff
@@ -29,11 +34,11 @@ export default class extends Model {
       errObj[fieldName].push(error);
     };
 
-    if (is.not.boolean(attrs.macStyleWinControls)) {
-      addError('macStyleWinControls', 'Please provide a boolean value.');
+    if (!controlStyles.includes( attrs.windowControlyStyle ) ) {
```
tenthhuman: I get you mean windowControlStyle -- yes there is an incorrect `y` after Control. Okay, cool with the lint rule.

js/models/LocalSettings.js
```diff
@@ -29,11 +34,11 @@ export default class extends Model {
...
+      addError('windowControlStyle', 'Please provide a boolean value.');
```
tenthhuman: Correct.

js/views/modals/Settings/Advanced.js
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  get events() {
+    return {
+      'change input[name="windowControlStyle"]': 'changedWindowStyle'
+    };
+  }
+
+  changedWindowStyle( event ) {
+    if ( event.target.id == 'windowControlStyleMac' )
```
rmisio: Do we really want to update the model (and therefore UI) as soon as the user changes the field? The rest of the form is persisted when you click save. It's a bit odd UX for me that localSettings are updated as soon as you change them in the form whereas server persisted settings happen when you click save.

The fact that one saves to localStorage whereas the other to the server, IMO, shouldn't change the UX. It's an implementation detail.

I could live with it if you guys feel otherwise, but it feels weird to me.

js/models/Profile.js
```diff
@@ -31,7 +31,7 @@ export default class extends BaseModel {
     return 'use-sync';
   }
 
-  nested() {
+  get nested() {
```
tenthhuman: For consistency -- this was the only place in the codebase where `<BaseModel>.nested` is not a getter ( until I reverted the Settings.js slot which you feel breaks address validation. 

In fact, this update was from the previous change set on this branch ( before the merge with the branch where I included the changes that emerged as workable ideas in the discussion ). Previously, @rmisio has said that it should work either way since it is accessed through `_.result`.

js/models/SMTPSettings.js
```diff
@@ -1,9 +1,41 @@
...
     };
   }
+
+  validate(attrs) {
+    const errObj = {};
+    const addError = (fieldName, error) => {
+      errObj[fieldName] = errObj[fieldName] || [];
+      errObj[fieldName].push(error);
+    };
+
```
tenthhuman: Correct.

js/models/Settings.js
tenthhuman: Removed. Please re-test.

js/views/modals/Settings/Advanced.js
tenthhuman: Okay, done.

js/views/modals/Settings/Advanced.js
tenthhuman: How does that make the purpose clearer?

package.json
tenthhuman: Please ask the `npm` auteurs.

```bash
>npm i--saved-dev eslint
>sudo npm i -g eslint
``` 

Reverted.

js/templates/modals/settings/advanced.html
```diff
@@ -0,0 +1,298 @@
...
+        </div>
+      </div>
+      <div class="flexRow">
+        <div class="col3">
+          <label>
+            <%= ob.polyT('settings.advancedTab.server.peers') %>
+          </label>
+        </div>
+        <div class="col9">
+          <a class="btn clrP clrBr js-showConnectedPeers flexNoShrink"><%= ob.polyT('settings.advancedTab.server.btnShowPeersAction') %></a>
```
rmisio: For now, this could be a link to the connected peers page: `href="#connected-peers"`

js/templates/modals/settings/advanced.html
```diff
@@ -0,0 +1,298 @@
...
+        </div>
+      </div>
+      <div class="flexRow">
+        <div class="col3">
+          <label>
+            <%= ob.polyT('settings.onOff') %>
+          </label>
+        </div>
+        <div class="col9">
```
rmisio: We've decided to remove this functionality, since the server is automatically turned off / on when you connect to the default connection (coming soon).

js/views/modals/Settings/Advanced.js
```diff
@@ -0,0 +1,90 @@
...
+    const $firstErr = this.$('.errorList:first');
+    if ($firstErr.length) $firstErr[0].scrollIntoViewIfNeeded();
+  }
+
+  saveLocal() {
+    const localData = this.getFormData(this.$localFields);
+    app.localSettings.set( localData );
+  }
+
+  save() {
```
rmisio: As of now this is never being called and localSettings (except Window Style) are not being saved at all.

I also think this method needs to be smarter. For example, I would only call save if both the localSettings and this.settings (server settings) models passed client validation. If either failed client side validation, I would let the user fix the errors first. It would be odd to me to have a local setting fail validation, we display a red error message, but server settings passed client validation and we show the status bar with "Saved!".

js/models/LocalSettings.js
```diff
@@ -16,7 +18,10 @@ export default class extends Model {
 
   defaults() {
     return {
-      macStyleWinControls: remote.process.platform === 'darwin',
+      windowControlStyle: remote.process.platform === 'darwin' ? 'mac': 'win',
+      showAdvancedVisualEffects: true,
+      saveTransactionMetadata: true,
+      defaultTransactionFee: 'high',
```
rmisio: Could we please get a validation for the defaultTransactionFee attribute. I'm assuming the valid values would be 'high', 'medium', 'low'?

js/views/modals/Settings/Advanced.js
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  get events() {
+    return {
+      'change input[name="windowControlStyle"]': 'changedWindowStyle'
+    };
+  }
+
+  changedWindowStyle(event) {
+    app.localSettings.set('windowControlStyle', $(event.target).val());
```
rmisio: Do we really want to update the model (and therefore UI) as soon as the user changes the field? The rest of the form is persisted when you click save. It's a bit odd UX for me that localSettings are updated as soon as you change them in the form whereas server persisted settings happen when you click save.

The fact that one saves to localStorage whereas the other to the server, IMO, shouldn't change the UX. It's an implementation detail.

I could live with it if you guys feel otherwise, but it feels weird to me.

js/views/modals/Settings/Advanced.js
```diff
@@ -0,0 +1,87 @@
...
+    const $firstErr = this.$('.errorList:first');
+    if ($firstErr.length) $firstErr[0].scrollIntoViewIfNeeded();
+  }
+
+  saveLocal() {
+    const localData = this.getFormData(this.$localFields);
+    app.localSettings.set(localData);
+  }
+
+  save() {
```
rmisio: As of now this is never being called and localSettings (except Window Style) are not being saved at all.

I also think this method needs to be smarter. For example, I would only call save if both the localSettings and this.settings (server settings) models passed client validation. If either failed client side validation, I would let the user fix the errors first. It would be odd to me to have a local setting fail validation, we display a red error message, but server settings passed client validation and we show the status bar with "Saved!".

js/views/modals/Settings/Advanced.js
jjeffryes: Normally attributes on an html element pertain to the html and how it should be handled by the browser. Data-attributes are generally used to store data that will be used by JavaScript. 

Since the browser doesn't do anything with the persist attribute, and it doesn't affect css, screenreaders, or anything else that directly reads the markup, I think it would be better as a data attribute.

js/views/modals/Settings/Advanced.js
jjeffryes: That's a good point, it would be more consistent to save the local storage data on save instead of when the button is clicked on.

js/views/modals/Settings/Advanced.js
morebrownies: Changing on the save event sounds best.

js/templates/modals/settings/advanced.html
tenthhuman: Okay

js/templates/modals/settings/advanced.html
tenthhuman: Okay

js/languages/en-US.json
```diff
@@ -263,6 +330,13 @@
     "provideUsername": "Please provide a username.",
     "duplicateSocialAccount": "You already have a social account of this type."
   },
+  "smtpModelErrors": {
```

js/views/modals/Settings/Advanced.js
tenthhuman: `<Advanced>.save` was called.
Validation on the nested models was not called.
I believe this may be related to @rmisio's PR #169 
I fetched this from upstream and merged it, 
and after that I confirmed validation is operating.

js/models/SMTPSettings.js
```diff
@@ -1,9 +1,50 @@
...
+      if (is.empty(attrs.password.trim())) {
+        addError('password', app.polyglot.t('smtpModelErrors.password'));
+      }
+      if (is.not.email(attrs.senderEmail.trim())) {
+        addError('senderEmail', app.polyglot.t('smtpModelErrors.senderEmail'));
+      }
+      if (is.not.email(attrs.recipientEmail.trim())) {
+        addError('recipientEmail', app.polyglot.t('smtpModelErrors.recipientEmail'));
+      }
+    }
```
jjeffryes: Something is still broken in validation, when I try deleting an address, this error is thrown:

```
Uncaught Error: Client side validation failed: {"smtpSettings.serverAddress":["Please provide a valid URL."],"smtpSettings.username":["Please provide a username."],"smtpSettings.password":["Please provide a valid password."],"smtpSettings.senderEmail":["Please provide a valid email."],"smtpSettings.recipientEmail":["Please provide a valid email."]}
```

Saving the addresses shouldn't result in the SMTP settings model validation failing.

**Edit: it looks like this is because I had notifications set to true, with no values for the other fields.** While that shouldn't happen in the wild once this is merged, we should handle that situation instead of just having the client break. 

For the situations where there are multiple models all being saved to one API, how do we want to handle cross-tab errors? If saving an address throws an error in another tab, we should alert the user to that so they can fix it.

js/views/modals/Settings/Advanced.js
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    this.render();
+    const $firstErr = this.$('.errorList:first');
+    if ($firstErr.length) $firstErr[0].scrollIntoViewIfNeeded();
+  }
+
+  saveLocal() {
+    const localData = this.getFormData(this.$localFields);
+    app.localSettings.set(localData);
```
jjeffryes: You need to do app.localSettings.save here too, just setting sets the model but the model isn't saved.

Currently, changing your window control style between mac and win doesn't persist when you refresh the client.

tenthhuman: Okay

js/models/SMTPSettings.js
rmisio: @tenthhuman Hold of any changes related to this. The reason this is happening is not exactly the reason Josh is giving. It's really that Josh has invalid data saved on the server. I'll explain more when I fully review this PR later. I was reviewing it now, but with you changing it, my comments are being blown away, so I'll wait.

js/languages/en-US.json
```diff
@@ -85,6 +85,41 @@
     "generalTab": "General",
     "pageTab": "Page",
     "addressesTab": "Addresses",
+    "advancedTab" : {
+      "_sectionName" : "Advanced",
+      "smtp" : {
+        "_sectionName" : "SMTP Integration",
```
rmisio: Do these names have to prefaced with an _? Why are they like that?

js/languages/en-US.json
```diff
@@ -85,6 +85,41 @@
...
+        "_testAction" : "Test Integration",
+        "notifications" : "Notifications",
+        "username" : "Username",
+        "password" : "Password",
```
rmisio: Our linter does not lint JSON, so it's kind of up to us to be consistent. Could you not have a space before the colon?

js/models/LocalSettings.js
```diff
@@ -16,7 +17,10 @@ export default class extends Model {
 
   defaults() {
     return {
-      macStyleWinControls: remote.process.platform === 'darwin',
+      windowControlStyle: remote.process.platform === 'darwin' ? 'mac' : 'win',
+      showAdvancedVisualEffects: true,
+      saveTransactionMetadata: true,
+      defaultTransactionFee: 'high',
```
rmisio: Can we please validate that any defaultTransactionFee provided is one of the available types (I assume 'high', 'medium', 'low')?

js/models/SMTPSettings.js
rmisio: @jjeffryes This is not happening because of cross-tab errors. We specifically avoid this happening by having each tab work on a clone of the model, so they're not working on the same model and any errors on a form in one tab, do not interfere nor are included with the saving of another tab.

The reason you are getting the error is because you have invalid data saved on the server. This happened because a.) we didn't have the right client validations in place b.) the server is not validating the settings data.

Cris has already fixed A and for B, I've logged a [server bug](https://github.com/OpenBazaar/openbazaar-go/issues/252).

Anyhow, to get you're data corrected, just click Save from the Advanced Tab and then refresh the app.

js/views/modals/Settings/Advanced.js
rmisio: Instead of working on the global app.localSettings, could you please make a clone of it in the constructor and work off of that? Also, please bind a sync handler to the cloned localSettings and update the global one there. Basically, give localSettings the same treatment as app.settings is getting.

The reason is that we don't want invalid data (data that failed client validation) making it into the global localSettings.

js/views/modals/Settings/Advanced.js
```diff
@@ -0,0 +1,103 @@
...
+      this.$el.html(t({
+        errors: this.settings.validationError || {},
+        ...this.settings.toJSON(),
+        ...app.localSettings.toJSON(),
+      }));
+
+      this.$formFields = this.$('select[name], input[name], textarea[name]').
+        not('[data-persistence-location="local"]');
+      this.$localFields = this.$('[data-persistence-location="local"]');
+      this.$el.find(`${siblingsToggleInputs}:checked`).
```
rmisio: You don't have to make this change, but as an FYI, backbone view's give you a shortcut for `this.$el.find`.

`this.$el.find('<selector>')` ==> `this.$('<selector>')`

js/views/modals/Settings/Advanced.js
```diff
@@ -0,0 +1,103 @@
...
+        });
+    }
+
+    this.render();
+    const $firstErr = this.$('.errorList:first');
+    if ($firstErr.length) $firstErr[0].scrollIntoViewIfNeeded();
+  }
+
+  save() {
+    this.reportSave(this.saveLocal() && this.saveServer());
```
rmisio: (Github is driving me bonkers. It's dropping some comments and this was a big one. Anyways, I'll try again)

How you're handling the save of these two models is problematic for a couple of reasons:

1. If one fails client side validation, but the other does not, it can result in a weird state where the UI is displaying a red error message for one model, yet the other one is being persisted. To a degree, this is masked by the fact that the nature of the radio buttons for the localSettings field won't let the user fail client validation, but for future proofing reasons, I think we should only save() both models if they both pass client validation.

2. The way you're passing the save's into reportSave means only one save will be "reported" on whereas they both should be included in the "reporting" process since they will fail / succeed independently of each other.

I would suggest doing the save as follows:

```
    this.localSettings.set(this.getFormData(this.$localFields), { validate: true });
    this.settings.set(this.getFormData(), { validate: true });

    if (this.localSettings.validationError || this.settings.validationError) {
      // client side validation failed on one or both models
      this.trigger('saveComplete', true);
    } else {
      this.trigger('savingToServer');

      // let's save and monitor both save processes
      const localSave = this.localSettings.save();
      const serverSave = this.settings.save();

      $.when(localSave, serverSave)
        .done(() => {
          // both succeeded!
          this.trigger('saveComplete');
        })
        .fail((...args) => {
          // One has failed, the other may have also failed or may
          // fail or may succeed. It doesn't matter, for our purposed one
          // failure is enough for us to consider the "save" to have failed
          this.trigger('saveComplete', false, true,
            args[0] && args[0].responseJSON && args[0].responseJSON.reason || '');
        });
    }

    this.render();
    const $firstErr = this.$('.errorList:first');
    if ($firstErr.length) $firstErr[0].scrollIntoViewIfNeeded();

```

tenthhuman: Okay

js/views/modals/Settings/Advanced.js
tenthhuman: Okay

js/languages/en-US.json
tenthhuman: The _ is to distinguish between words to do with model slots ( names of slots, errors about slots ), and words to do with things which are in the view and are meta to the model slots. Things like actions, or section headings.

js/languages/en-US.json

js/views/modals/Settings/Advanced.js
```diff
@@ -0,0 +1,125 @@
+import app from '../../../app';
+import loadTemplate from '../../../utils/loadTemplate';
+import baseVw from '../../baseVw';
+import $ from 'jquery';
+
+const siblingsToggle = '[data-next-siblings-toggle]';
+const siblingsToggleInputs = `${siblingsToggle} input[type="radio"]`;
+
+function getResponseReason(resp) {
```
rmisio: In the code I gave you, this same thing was accomplished in one line:

```
resp && resp.responseJSON && resp.responseJSON.reason || '';
```

Why break it up into a 9 line function?

js/views/modals/Settings/Advanced.js
```diff
@@ -0,0 +1,125 @@
...
+    };
+  }
+
+  attemptSave({ cloudFail, localFail }) {
+    if (this.localSettings.validationError || this.settings.validationError) {
+      this.trigger('saveComplete', true);
+    } else {
+      this.trigger('savingToServer');
+      const saving = this.startSaving();
+      saving.local.done(() =>
```
rmisio: Not using $.when here (as I had suggested) is probably OK in this case since the first call is saving locally. If, however, these were two server saves, I would strongly recommend $.when or $.whenAll, since that would kick off the multiple requests concurrently rather than sequentially.

js/views/modals/Settings/Advanced.js
```diff
@@ -0,0 +1,125 @@
...
+    const t = event.target;
+    this.applyNextSiblingsToggle(t, t.value);
+  }
+
+  applyNextSiblingsToggle(t, val) {
+    const p = $(t).parents(siblingsToggle);
+    if (val === 'true') {
+      p.addClass('showNextSiblings');
+    } else {
+      p.removeClass('showNextSiblings');
```
jjeffryes: I think it's a bad user experience to be able to save something as valid (to the user's perception, if it saves with no error it's valid data), but when you turn notifications on, it's invalid. The user doesn't know we only check those fields when notifications are turned off.

js/views/modals/Settings/Advanced.js
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  showErrors() {
+    const $firstErr = this.$('.errorList:first');
+    if ($firstErr.length) $firstErr[0].scrollIntoViewIfNeeded();
+  }
+
+  saveLocalAndServer() {
+    const backupLocalSettings = this.localSettings.toJSON();
+    const rollbackLocal = () => { this.localSettings.set(backupLocalSettings); this.localSettings.save(); };
```
tenthhuman: Okay

js/views/modals/Settings/Advanced.js
tenthhuman: Okay

js/views/modals/Settings/Advanced.js
```diff
@@ -0,0 +1,125 @@
...
+
+  setAndValidate() {
+    this.localSettings.set(this.getFormData(this.$localFields), { validate: true });
+    this.settings.set(this.getFormData(), { validate: true });
+  }
+
+  startSaving() {
+    return {
+      local: this.localSettings.save(),
+      cloud: this.settings.save(),
```
tenthhuman: Okay

js/views/modals/Settings/Advanced.js
tenthhuman: Okay

js/views/modals/Settings/Advanced.js
tenthhuman: Okay.

The property getting idiom using `&&` to test the chain or falling back to a default is an idiom that is widely used. 

My feeling is this type of code is not really clear and direct as to its purpose. One reason is because its purpose is to do this:

```js
const deepValue = this.is.a.long.property.path;
if (!deepValue) deepValue = defaultValue;
```

Which is clear and direct about what is intended. 

The one line version using Boolean short-circuit evaluation is an idiom that probably is clear to a person if the person already knows the idiom, or if a person works out what the evaluation would be.

js/models/SMTPSettings.js
```diff
@@ -21,7 +21,20 @@ export default class extends BaseModel {
...
+    // get all keys that have String values
+    const stringValuedKeys = Object.keys(attrs).filter(k => typeof attrs[k] === 'string');
+
+    // trim all string values of whitespace
+    stringValuedKeys.forEach(k => { attrs[k] = attrs[k].trim(); });
+
+    // check if they are all empty
+    const everyInputEmpty = stringValuedKeys.every(k => attrs[k].length === 0);
+
+    if (everyInputEmpty) {
```
jjeffryes: Rob's right, we should do one of the following here:

1. only validate the fields that have a value when notifications are turned off, so blank fields are not marked invalid but fields with data that is incorrect are.

2. go back to only validating when notifications are turned on. That would still allow saving of invalid data when that data isn't used, but that would be better than making the user think they need to fill in all the blank fields when notifications are off.

js/views/modals/Settings/Advanced.js
```diff
@@ -0,0 +1,85 @@
...
+      this.settings.get('smtpSettings').defaults(), { validate: true });
+    this.render();
+  }
+
+  getFormData(subset = this.$formFields) {
+    return super.getFormData(subset);
+  }
+
+  save() {
+    this.localSettings.set(this.getFormData(this.$localFields), { validate: true });
```
rmisio: Could you add this line as the first line in this function?

```
this.trigger('saving');
```

Sorry... forgot to include that when I gave you the code earlier. Without it, the "Settings Saved." status bar message at the bottom of the screen never goes away.

js/models/SMTPSettings.js
tenthhuman: Okay

tenthhuman:
Merged!!! 🎉 🎈 👍 👍 👻 🍓 !!

I'm so happy right now to see this right now.

Thanks @jjeffryes @rmisio and @morebrownies for all your extensive feedback, tireless reviews and patient support contributing to us merging my first PR.

Thanks!! Onward! XD 😂 😜 xx

tenthhuman:
Closes #94

hoffmabc:
Congrats @tenthhuman thanks so much for your patience this was a big one!

PR #116, opened 2016-12-01
Store styles

morebrownies:
**Changes**
- Made the FREE SHIPPING badge a label so that it can be clicked to toggle free shipping
- Changed the listingCard hover event to be less overwhelming
- Changed the listingCard edit/delete buttons to be less overwhelming
- Added a listingCard transition on hover
- Changed the hero image to stretch the full width (background-size: cover + center position)
- Changed the modal/overlay background to inherit the secondary color as the background 
- Cleaned up some spacing issues in a few places
- Cleaned up some typography stuff
- Set the secondary color in a few places it was missing
- Radio button labels are now bold when active
- Updated the user navigation
- Updated the modal/overlay styling a bit
- Added missing shadows to listingCard and storeFilters


hoffmabc:
I think you need to rebase against the latest code since Josh's stuff isn't in there.

hoffmabc:
I went ahead and rebased it for you.

hoffmabc:
@morebrownies there's some weird css stuff on the listing delete button.
image

morebrownies:
Thank you for the rebase. I'll take a look at the css issue.

rmisio:
@hoffmabc was that photo really necessary? I don't think I want to know what item that listing is offering.

hoffmabc:
Just a moment of brevity @rmisio

morebrownies:
Ok I fixed the css issue. This should be ready for review

hoffmabc:
@morebrownies another border issue

image

styles/components/_listingCard.scss
```diff
@@ -31,22 +36,10 @@
     align-items: center;
     justify-content: center;
     z-index: 1; // so it's above the overlay icons
-
-    .overlayPanelInner {
```
hoffmabc: Should this be removed or is it necessary still @rmisio ?

rmisio: I will bring this up with wolf because his removal of it from the html has caused some issue.

hoffmabc: Thinking about this, I may have borked this on the rebase potentially. So it might be my fault.

rmisio:
@morebrownies Could you also style the Deleting and Deleted state of the listing card? You could see them by applying .listingDeleting or .listingDeleted to the .listingCard element.

The styles were very much placeholder styles before, but now, with the removal of the overlay panel, they are even worse:

before:

image

now:

image

Anyhow, they're important for a few reasons:

a.) The delete is an async process and can complete at any time - the user may be happily scrolling further down the list when the delete completes, so we don't want to just remove the item having the page jump abruptly.
b.) Removing a listing may affect filters, e.g. a category may need to be removed from the cat filter. Having this update on the fly is a bit of a can of worms to make the experience smooth, so we've decide to take the approach of popping in a message urging the user to refresh:

image

c.) The delete could be initiated elsewhere, namely the listing detail modal. If it takes a bit, the user could close the modal landing back on the store. We would want a "deleting" state so the user understands the listing is in the process of being deleted.

rmisio:
@morebrownies FYI: The fact that you removed the overlayPannelInner element from the markup is a bit problematic because now when you click on of the icons, e.g. Edit, it not only opens the Edit modal, it also opens the Detail modal. This is because a click is happening on the icon (causing the Edit modal to open) but also bubbling up to the card (causing the Detail modal to open). The overlay panel was preventing the latter from happening.

Anyhow, let's fix the Deleting and Deleted styles first and depending on how that is fixed, I will either put in a patch to fix this issue or advise you on possible approaches (maybe we'll even resurrect the overlay panel, but with a transparent bg...).

styles/_base.scss
```diff
@@ -165,6 +164,7 @@ ul, ol {
         margin: 0 calc((100% - #{$pageWidth + $padLg + $padLg}) / 2) 0 auto;
         padding: 0 $padLg $bar $padLg;
         position: relative;
+        min-height: 1000px;
```
rmisio: @jjeffryes do you think it's a good idea to unilaterally be adding such a min-height to all pages?

styles/components/_listingCard.scss
rmisio: TBD

morebrownies:
I broke out a few of these requests into separate issues. #124 #125 

@rmisio I'll take a closer look at overlayPannelInner in a few.

rmisio:
@morebrownies Ok, but I don't think we could merge this request until #124 is addressed in some way, because without the overlay panel the deleting states get washed out and appear broken. Check out Brians firestorm in Slack (lol).

morebrownies:
I'm going to add overlayPannelInner back, but styling the delete state I'll do later.

js/templates/userPage/store.html
```diff
@@ -40,15 +40,15 @@ <h1 class="txCtr">Loading...</h1>
           <div class="select2Small js-shipsToSelectDropdownContainer"></div>
         </div>
         <div class="flexVCent rowSm">
-          <input type="checkbox" class="margRSm js-filterFreeShipping" <% if (ob.filter.freeShipping) print('checked') %> />
-          <span class="clrE1 clrTOnEmph txTn txCtr upper padTn"><%= ob.polyT('userPage.store.freeShippingBanner') %></span>
+          <input type="checkbox" id="filterFreeShipping" class="margRSm margL0 js-filterFreeShipping" <% if (ob.filter.freeShipping) print('checked') %> />
```
jjeffryes: Since this is in a flex wrapper, it would be better to use a gutterHSm class on the parent than to monkeypatch things (to borrow a phrase from Cris S) with a margin override. 

I do see that if we do that the checkbox still has a left margin from the browser defaults, I've updated master with a fix.

js/templates/userPage/store.html
```diff
@@ -40,15 +40,15 @@ <h1 class="txCtr">Loading...</h1>
...
+          <label class="clrE1 clrTOnEmph txTn txCtr upper padTn borderRSm txB" for="filterFreeShipping"><%= ob.polyT('userPage.store.freeShippingBanner') %></label>
```
jjeffryes: Flagging this to get the labelBox class if you implement that.

styles/_base.scss
jjeffryes: What's the purpose of the min-height? The pagecontainer element always sizes itself to the height of the content in it.

styles/_theme.scss
```diff
@@ -9,6 +9,10 @@
     background-color: $secondary;
   }
 
+  input[type="radio"]:checked + label {
+    font-weight: 700;
+  }
```
jjeffryes: The theme file should only be used for colors. Put input styles other than colors in the form file. If you do it like this, it only applies to radio labels inside a color class, which may not always be the case.

styles/_theme.scss
```diff
@@ -20,6 +24,10 @@
   input[type="radio"] + label:before {
     background-color: $primary;
   }
+  
+  input[type="radio"]:checked + label {
+    font-weight: 700;
+  }
```
jjeffryes: You can remove this one.

styles/_variables.scss
```diff
@@ -9,15 +9,15 @@ $secondary: adjust-color($primary, $lightness: -4%);
 $background: adjust-color($primary, $lightness: -10%);
 $border: adjust-color($primary, $lightness: -14%);
 $border2: adjust-color($primary, $lightness: 100%);
-$focusedBorder: rgb(0, 0, 0);
+$focusedBorder: adjust-color($text, $lightness: 35%);
 $emphasis1: rgb(253,166,57);
 $emphasis2: rgb(174, 138, 88);
 $error: red;
-$overlay: adjust-color($primary, $lightness: -56%, $alpha: -0.35);
+$overlay: adjust-color($secondary, $lightness: 0%, $alpha: -0.1);
```
jjeffryes: $lightness: 0% doesn't do anything, you can write this as adjust-color($secondary, $alpha: -0.1).

styles/components/_buttons.scss
```diff
@@ -21,6 +21,7 @@
   &:hover {
     text-decoration: none;
     cursor: pointer;
+    border: solid 1px $focusedBorder;
```
jjeffryes: This should be border-color, otherwise any buttons with a different border width will get swapped to a 1px border width.

styles/components/_buttons.scss
```diff
@@ -109,6 +114,17 @@
...
 
+.iconBtnSm {
+  //this button has only an icon in it
+  @extend .btn;
+  font-size: $tx5;
+  padding: 0;
+  margin: 0;
+  width: 30px;
+  height: 30px;
+  transition: all 0.15s ease;
```
jjeffryes: Should the transition be on the .btn class, or does only the iconBtnSm class have a transition?

What is supposed to transition? The only thing I saw was the border on hover, if that's the intent it would be better to transition only the border color, so other properties aren't affected by surprise (for instance, if a group of elements was supposed to immediately go to opacity 0, this button would fad slower than the rest of them).

styles/components/_form.scss
```diff
@@ -26,7 +26,7 @@ textarea,
   color: inherit;
   border-radius: 3px;
   font: {
-    family: "Helvetica Neue", Helvetica, Arial, sans-serif;
+    font-family: "Noto Sans", sans-serif;
```
jjeffryes: There's a typo here, it should be "family" not "font-family" (SASS turns this into "font-font-family" if you look in your web dev panel), but font family is set by the base file, so it can just be removed here.

styles/components/_layout.scss
```diff
@@ -304,6 +304,10 @@
   margin-left: $pad;
 }
 
+.margL0 {
+  margin-left: 0;
+}
+
```
jjeffryes: Let's avoid this kind of override class if at all possible.

styles/components/_layout.scss
```diff
@@ -411,3 +415,11 @@
...
   position: relative;
 }
+
+.borderRSm {
+  border-radius: 1px;
+}
+
+.borderRMd {
+  border-radius: 2px;
+}
```
jjeffryes: Let's also avoid border-radius classes, in most cases this should be taken care of by a container class.

In general, we're trying to avoid single-property classes, so we don't end up with hundreds of them.

styles/components/_listingsGrid.scss
```diff
@@ -4,11 +4,22 @@
   .listingCard {
     margin: 0 $padMd $padMd 0;
     font-size: $tx5b;
+    border-radius: 2px;
```
jjeffryes: Use the $corner variable here.

styles/components/_modal.scss
```diff
@@ -12,6 +12,8 @@
     min-width: 300px;
     max-height: calc(100% - 50px);
     overflow: auto;
+    border: solid 1px $border;
+    border-radius: 2px;
```
jjeffryes: Use $corner here for the radius.

styles/modules/_pageNav.scss
```diff
@@ -12,12 +12,13 @@
...
 
     .addressBar {
-      font-size: $tx5b;
+      font-size: $tx5;
       line-height: 1em; //allow dragging on area above and below the text input in the address bar
     }
 
     #AvatarBtn {
       margin: 5px 0;
+      border: solid 2px #fff;
```
jjeffryes: As noted above, remove the color here, otherwise it can't be set by the theme.

js/templates/modals/settings/general.html
```diff
@@ -1,8 +1,8 @@
 <div class="bar padMd clrBr borderBottom">
   <h2 class="h3 txUnb"><%= ob.polyT('settings.generalTab') %></h2>
 </div>
-<div class="padMd tabFormWrapper clrS">
-  <form class="box border clrP clrBr padMdKids padStack settingsGeneralForm">
+<div class="tabFormWrapper clrP">
```
rmisio: tabFormWrapper is being used in _settings.scss. We shouldn't remove it unless that file were adjusted to.

morebrownies:
Thanks for the detailed write up @jjeffryes. I think it will be more efficient for me to close this PR and open issues for things that need to be tweaked in client. I have zero bandwidth to write code that meets our standards.

jjeffryes:
I think this PR could be patched up in a reasonable amount of time, and it would be a good experience. But it's up to you. 

There are definitely a lot of intricate things going on that it takes some time to get used to, we can do a walkthrough some time if you want. That'd probably be good for everyone.

morebrownies:
I'm going to take a different approach. Instead of trying to knock out 20 style issues in one PR, I'm going to open issues in GH first and then write small PRs to address 1 or 2 of them at a time.

jjeffryes:
Roger that.

js/templates/pageNav.html
```diff
@@ -30,15 +30,15 @@
...
+          <a id="AvatarBtn" class="discSm border clrBr2 clrSh1 js-navListBtn navListBtn"
           <% if (ob.avatarHash) { %>
           //this will be updated to the real image parameter
           style="background-image: url(<%= ob.getServerUrl(`ipfs/${ob.avatarHash}`) %>), url('../imgs/defaultAvatar.png')"
           <% } else { %>
             style="background-image: url('../imgs/defaultAvatar.png')"
           <% } %>
           ></a>
-          <nav class="navList listBox clrS clrSh1 foldDown js-navList js-navPopMenu">
+          <nav class="navList listBox clrP clrSh1 clrBr foldDown js-navList js-navPopMenu">
```
morebrownies: I tend to prefer the slide down transition that we have in 1.0. The current transition in 2.0 seems a bit flimsy.

jjeffryes: The v1 is pretty hacky, it hides the menu behind the top bar and moves it up, which creates bugs where it peaks through if we're not super careful.

The clipping path is pretty reliable, it'll just need some tweaking to make it wider to account for the shadow etc.

If that doesn't work for some reason, I'll come up with something else.

jjeffryes: What do you feel is different about how it feels?

morebrownies: It could just be a timing thing. I can play around with it a bit to see if I can improve on it.

PR #118, opened 2016-12-02
Listing validations

rmisio:
Adds in client side validations on the Listing modal for:
- title length
- number of tags
- length of individual tag
- number of categories

### DIFF
--- a/js/models/listing/Item.js
+++ b/js/models/listing/Item.js
@@ -56,11 +56,11 @@ export default class extends BaseModel {
   }
 
   get maxTags() {
-    return 2;
+    return 10;
   }
 
   get maxCategories() {
-    return 3;
+    return 10;
   }
 
   get titleMaxLength() {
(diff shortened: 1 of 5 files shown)


js/models/listing/Item.js
```diff
@@ -90,13 +94,32 @@ export default class extends BaseModel {
...
+    if (attrs.tags) {
+      if (is.array(attrs.tags)) {
+        if (attrs.tags.length > max.tags) {
+          addError('tags',
+            app.polyglot.t('itemModelErrors.tooManyTags', { maxTags: max.tags }));
+        }
+
+        attrs.tags.forEach((tag, index) => {
+          if (is.not.string(tag)) {
+            addError('tags', `Tag at index ${index} is not a string and should be.`);
```
rmisio: As usual, if you're seeing an untranslated model validation, it's because the UI has a validation mechanism in place (e.g. maxLength) that will never allow the user to enter bad data. The validation is in the model for data integrity purposes, in case the model is set programatically.

PR #128, opened 2016-12-02
This closes #125

### LINKED ISSUE #125: Button stacking leads to display issues on hover event
@morebrownies another border issue

image

and here:

image

both message and follow have the issue

jjeffryes:
This will update the storeStyles branch to only remove the right border on buttons in a button strip when they aren't in a hover state.

### DIFF
--- a/styles/_variables.scss
+++ b/styles/_variables.scss
@@ -13,7 +13,7 @@ $focusedBorder: adjust-color($text, $lightness: 35%);
 $emphasis1: rgb(253,166,57);
 $emphasis2: rgb(174, 138, 88);
 $error: red;
-$overlay: adjust-color($secondary, $lightness: 0%, $alpha: -0.1);
+$overlay: adjust-color($secondary, $alpha: -0.1);
 
 // typography
 $baseFontSize: 10px;
--- a/styles/components/_buttons.scss
+++ b/styles/components/_buttons.scss
@@ -136,8 +136,8 @@
     border-radius: 0;
     float: left; //float left will collapse the button and position it correctly
   }
-  & > .btn:not(:hover) {  // on hover use the normal border width
-    border-right-width: 0;
+  & > .btn:not(:last-child):not(:hover) {   // on hover use the normal border width
+    border-right-color: transparent;
   }
   & > .btn:first-child {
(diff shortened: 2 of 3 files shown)


styles/_variables.scss
```diff
@@ -13,7 +13,7 @@ $focusedBorder: adjust-color($text, $lightness: 35%);
 $emphasis1: rgb(253,166,57);
 $emphasis2: rgb(174, 138, 88);
 $error: red;
-$overlay: adjust-color($secondary, $lightness: 0%, $alpha: -0.1);
+$overlay: adjust-color($secondary, $alpha: -0.1);
```
jjeffryes: I accidentally included this fix to one of the comments I made, so free bonus.

styles/components/_buttons.scss
```diff
@@ -21,12 +21,13 @@
...
     text-decoration: none;
     cursor: pointer;
-    border: solid 1px $focusedBorder;
+    border-color: $focusedBorder;
   }
 
   &:focus {
     text-decoration: none;
     outline: 0;
+    border-color: $focusedBorder;
```
jjeffryes: The border should change on focus too, for anyone using a keyboard to tab to the buttons.

rmisio:
This ends up causing the button to jump by a pixel on hover.

@morebrownies @jjeffryes Is that ok?

morebrownies:
It would prefer if the button didn't jump a pixel.

jjeffryes:
Good catch. It shouldn't do that. I'll fix it.

rmisio:
Can we omit the hover border on disabled elements?

image

jjeffryes:
The code is updated to not affect disabled elements, or children of disabled elements. It also is moved to the themes file so it can add a different focus color for different border classes if desired.

Note that this does not affect the screenshot, as that element has a span inside it with an opacity set, rather than being disabled. We should change the view code to add a disabled class to that button when it's not usable.

hoffmabc:
@jjeffryes do you mind leaving a message when closing for historical purposes to understand what happened?

jjeffryes:
My bad there. This is a branch from a PR that was closed, so it's no longer a viable PR.

PR #138, opened 2016-12-02
Fixes #136

### LINKED ISSUE #136: Address bar is using Helvetica instead of Noto Sans
screenshot 2016-12-02 12 09 18

morebrownies:
Fixes #136 + made the font-size slightly larger

### DIFF
--- a/styles/components/_form.scss
+++ b/styles/components/_form.scss
@@ -26,7 +26,7 @@ textarea,
   color: inherit;
   border-radius: 3px;
   font: {
-    family: "Helvetica Neue", Helvetica, Arial, sans-serif;
+    family: "Noto Sans", sans-serif;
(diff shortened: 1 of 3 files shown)


styles/components/_form.scss
```diff
@@ -26,7 +26,7 @@ textarea,
   color: inherit;
   border-radius: 3px;
   font: {
-    family: "Helvetica Neue", Helvetica, Arial, sans-serif;
+    family: "Noto Sans", sans-serif;
```
jjeffryes: You can delete the entire font rule, it is set by the base style.

jjeffryes: You're right, I'm looking at it, it looks like the user agent has higher specificity.

PR #139, opened 2016-12-02
Open anothers listing

rmisio:
- Fixes an issue where other peoples listings were not being fetched with the right url from their store
- Puts icons on the Edit and Delete listing buttons (when deleting from a card). (I'll leave the deeper styling to Mike, I just did the icon part because the JS was not accounting for the icons in determining whether a edit/delete button was clicked vs. the card being clicked and it was therefore wrongly triggering both actions.)

### DIFF
--- a/js/views/ListingCard.js
+++ b/js/views/ListingCard.js
@@ -22,7 +22,7 @@ export default class extends baseVw {
       throw new Error('Please provide a ListingShort model.');
     }
 
-    if (!opts.ownListing) {
+    if (opts.ownListing === undefined) {
       // For search and channels this will need to be included in the API, in
       // which case, we could get the value from the model and this could
       // be completely optional.
--- a/js/views/ListingCard.js
+++ b/js/views/ListingCard.js
@@ -116,8 +116,9 @@ export default class extends baseVw {
     // how likely is it that the listing will change (probably not very likely) and
     // how graceful is the experience if the user goes through the buy flow on
     // an old listing.
-    if (!this.options.ownListing ||
-        (e.target !== this.$btnEdit[0] && e.target !== this.$btnDelete[0])) {
+    if (!this.ownListing ||
+        (e.target !== this.$btnEdit[0] && e.target !== this.$btnDelete[0] &&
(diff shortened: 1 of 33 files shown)


js/views/ListingCard.js
```diff
@@ -22,11 +22,10 @@ export default class extends baseVw {
       throw new Error('Please provide a ListingShort model.');
     }
 
-    if (opts.ownListing === undefined) {
-      // For search and channels this will need to be included in the API, in
-      // which case, we could get the value from the model and this could
-      // be completely optional.
-      throw new Error('Please provide an ownListing indicator.');
+    if (!options.ownerGuid) {
+      throw new Error('Please provide a a guid representing the owner of the listing.');
```
jjeffryes: It's inconsequential, but you have "a" twice there.

styles/_base.scss
```diff
@@ -35,7 +35,7 @@ html {
     weight: 400;
   }
   -webkit-font-smoothing: antialiased;
-  letter-spacing: 0;
+  letter-spacing: 0.2px;
```
jjeffryes: It looks like this is undoing Mike's update?

styles/components/_form.scss
```diff
@@ -26,7 +26,7 @@ textarea,
   color: inherit;
   border-radius: 3px;
   font: {
-    family: "Noto Sans", sans-serif;
+    family: "Helvetica Neue", Helvetica, Arial, sans-serif;
```
jjeffryes: This is undoing Mike's update to the form file.

styles/modules/_pageNav.scss
```diff
@@ -12,7 +12,7 @@
     }
 
     .addressBar {
-      font-size: $tx5;
+      font-size: $tx5b;
```
jjeffryes: This undoes Mike's update.

styles/components/_form.scss
```diff
@@ -130,7 +130,3 @@ label {
     background-color: #DDD;
   }
 }
-
-input[type="radio"]:checked + label {
-  font-weight: 700;
-}
```
jjeffryes: Should it be removing this css? I must have missed it last time

PR #140, opened 2016-12-02
Listing card fixes (fixes #130)

### LINKED ISSUE #130: Listing cards and store filters are missing shadows and border radius
screenshot 2016-12-02 10 56 13

morebrownies:
Fixes #130

### DIFF
--- a/js/templates/userPage/store.html
+++ b/js/templates/userPage/store.html
@@ -27,7 +27,7 @@
 <% } else if (!ob.isFetching) { %>
   <div class="flex">
     <div class="col storeFilters <% if (!ob.listingCount) print('disabled') %>">
-      <div class="clrP clrBr border padMd clrT">
+      <div class="clrP clrBr border padMd clrT clrSh2 searchFilter">
         <div class="txB tx4 row"><%= ob.polyT('userPage.store.shippingFilter.heading') %></div>
         <div class="flexVCent tx5 rowSm">
           <label class="margRSm" for="shipsToSelect"><%= ob.polyT('userPage.store.shippingFilter.shipsTo') %>:</label>
--- a/js/templates/userPage/store.html
+++ b/js/templates/userPage/store.html
@@ -27,7 +27,7 @@
 <% } else if (!ob.isFetching) { %>
   <div class="flex">
     <div class="col storeFilters <% if (!ob.listingCount) print('disabled') %>">
-      <div class="clrP clrBr border padMd clrT contentBox">
+      <div class="clrP clrBr padMd clrT clrSh2 contentBox">
         <div class="txB row"><%= ob.polyT('userPage.store.shippingFilter.heading') %></div>
         <div class="flexVCent tx5 rowSm">
           <label class="margRSm" for="shipsToSelect"><%= ob.polyT('userPage.store.shippingFilter.shipsTo') %>:</label>
--- a/js/templates/userPage/store.html
+++ b/js/templates/userPage/store.html
@@ -27,8 +27,8 @@
 <% } else if (!ob.isFetching) { %>
   <div class="flex">
     <div class="col storeFilters <% if (!ob.listingCount) print('disabled') %>">
-      <div class="clrP clrBr border padMd clrT clrSh2 searchFilter">
-        <div class="txB tx4 row"><%= ob.polyT('userPage.store.shippingFilter.heading') %></div>
+      <div class="clrP clrBr border padMd clrT contentBox">
+        <div class="txB row"><%= ob.polyT('userPage.store.shippingFilter.heading') %></div>
         <div class="flexVCent tx5 rowSm">
           <label class="margRSm" for="shipsToSelect"><%= ob.polyT('userPage.store.shippingFilter.shipsTo') %>:</label>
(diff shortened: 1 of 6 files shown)


morebrownies:
Hold off on this for a minute. I need to add the border radius and shadow to the listing card too, in order to close #130

morebrownies:
Ok should be ready for review

js/templates/userPage/store.html
```diff
@@ -27,7 +27,7 @@ <h1 class="txCtr">Loading...</h1>
 <% } else if (!ob.isFetching) { %>
   <div class="flex">
     <div class="col storeFilters <% if (!ob.listingCount) print('disabled') %>">
-      <div class="clrP clrBr border padMd clrT">
+      <div class="clrP clrBr border padMd clrT clrSh2 searchFilter">
```
jjeffryes: If the searchFilter class only adds the rounded corner, this div should probably get the contentBox class instead. That will add the corner and the border, so you could remove border and searchFilter.

It also sets the text size to tx4, so that class could be removed from the Shipping text inside it.

The contentBox class is intended to be used for all of the boxes in the design that share the same basic properties, I think there are some places we forgot to use it.

If we want all of those elements to stay consistent, and change together if we ever change the design (like to tweak the corner radius), then we should use the contentBox class. If we think one of those elements should stay the same if the others change, then we could use a custom class.

js/views/userPage/CategoryFilter.js
```diff
@@ -23,7 +23,7 @@ export default class extends BaseVw {
   }
 
   className() {
-    return 'clrP clrBr border padMd clrT categoryFilter';
+    return 'clrP clrBr border padMd clrT clrSh2 categoryFilter';
```
jjeffryes: This element could have the contentBox class too.

js/views/ListingCard.js
```diff
@@ -66,7 +66,7 @@ export default class extends baseVw {
   }
 
   className() {
-    return 'listingCard col clrBr clrT clrP';
+    return 'listingCard col clrBr clrT clrP clrSh2';
```
jjeffryes: The listing card could get the contentBox class too.

styles/components/_listingsGrid.scss
```diff
@@ -4,6 +4,7 @@
   .listingCard {
     margin: 0 $padMd $padMd 0;
     font-size: $tx5b;
+    border-radius: $corner;
```
jjeffryes: If the listing card had the content box, we could remove the border-radius here, and remove the border style and width.

styles/modules/_userPage.scss
```diff
@@ -90,9 +90,14 @@
           }
         }
 
+        .searchFilter {
+          border-radius: $corner;
```
jjeffryes: See comments on contentBox.

styles/modules/_userPage.scss
```diff
@@ -90,9 +90,14 @@
...
+        }
+
         .categoryFilter {
           max-height: 250px;
           overflow: auto;
+          border-radius: $corner;
```
jjeffryes: See comments on contentBox.

js/templates/userPage/store.html
morebrownies: Ok, good to know. 👍

js/templates/userPage/store.html
```diff
@@ -27,8 +27,8 @@ <h1 class="txCtr">Loading...</h1>
 <% } else if (!ob.isFetching) { %>
   <div class="flex">
     <div class="col storeFilters <% if (!ob.listingCount) print('disabled') %>">
-      <div class="clrP clrBr border padMd clrT">
-        <div class="txB tx4 row"><%= ob.polyT('userPage.store.shippingFilter.heading') %></div>
+      <div class="clrP clrBr border padMd clrT contentBox">
```
jjeffryes: You can drop the border class too, that's taken care of by contentBox.

js/views/userPage/CategoryFilter.js
```diff
@@ -23,7 +23,7 @@ export default class extends BaseVw {
   }
 
   className() {
-    return 'clrP clrBr border padMd clrT categoryFilter';
+    return 'clrP clrBr border padMd clrT contentBox categoryFilter';
```
jjeffryes: You can drop the border class here too.

js/views/ListingCard.js
```diff
@@ -66,7 +66,7 @@ export default class extends baseVw {
   }
 
   className() {
-    return 'listingCard col clrBr clrT clrP';
+    return 'listingCard col clrBr clrT clrP contentBox';
```
jjeffryes: We should remove the border style and border width on the listingCard class so we're not assigning the same properties twice.

morebrownies:
Ok this is ready for another look. I had to add clrSh2 to a few places because the shadows went away. Let me know if I added them incorrectly.

jjeffryes:
I didn't realize the shadows were missing, those definitely need to be there, since they're a "color" class that is set by the theme.

morebrownies:
Cool, I added them in the last push.

PR #144, opened 2016-12-02
Fix delete overlay

jjeffryes:
I noticed #139 no longer showed the delete overlay, one instance of this.options.ownListing got missed in the refactor.


rmisio:
Are you sure you want all these other CSS tweaks to go in with this merge?

jjeffryes:
No, that was an accidental merge from master, I just removed them.

PR #149, opened 2016-12-04
Wire up application reload button

hoffmabc:
Ok @rmisio @jjeffryes this really bothered me that it didn't work so I just went ahead and added it in. It uses the Electron remote library to find the current window and reload it. 

Review and merge if you can please.

image

### DIFF
--- a/js/views/PageNav.js
+++ b/js/views/PageNav.js
@@ -1,4 +1,4 @@
-import remote from 'electron';
+import { remote } from 'electron';
 import multihashes from 'multihashes';
 import { View } from 'backbone';
 import loadTemplate from '../utils/loadTemplate';
@@ -67,7 +67,7 @@ export default class extends View {
   }
 
   navReload() {
-    remote.getCurrentWindow().reload();
+    location.reload();
   }
 
   setWinControlsStyle(style) {
--- a/js/views/PageNav.js
+++ b/js/views/PageNav.js
@@ -1,4 +1,4 @@
-import electron from 'electron';
+import { electron, remote } from 'electron';
 import multihashes from 'multihashes';
 import { View } from 'backbone';
 import loadTemplate from '../utils/loadTemplate';
@@ -9,14 +9,13 @@ import { launchEditListingModal } from '../utils/modalManager';
 import Listing from '../models/listing/Listing';
 import { isHiRez } from '../utils/responsive';
 
-const remote = electron.remote;
-
 export default class extends View {
   constructor(options) {
     const opts = {
       events: {
         'click .js-navBack': 'navBackClick',
         'click .js-navFwd': 'navFwdClick',
+        'click .js-navReload': 'navReload',
         'click .js-navClose': 'navCloseClick',
         'click .js-navMin': 'navMinClick',
         'click .js-navMax': 'navMaxClick',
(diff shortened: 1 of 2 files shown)


js/views/PageNav.js
```diff
@@ -67,6 +66,10 @@ export default class extends View {
     window.history.forward();
   }
 
+  navReload() {
+    remote.getCurrentWindow().reload();
```
rmisio: When I run this, I get:

```
PageNav.js [sm]:70 Uncaught TypeError: _electron2.default.getCurrentWindow is not a function
```

An alternate way to run this (which would save you an import) would be to just do `location.reload();`.

rmisio: Ok, I see why it's bombing. You need to change:

```
import remote from 'electron';
```

to

```
import { remote } from 'electron';
```

otherwise in the former, you're essentially importing the full 'electron' module, just assigning it to a variable named 'remote', whereas in the latter you're assigning the 'remote' property of the 'electron' module to a variable named 'remote'.

hoffmabc: ok. interesting that it's not bombing on my side.

morebrownies:
Does this refresh the current page or does it refresh the entire Electron app? It should be the former, like we do in v1.

hoffmabc:
It reloads the entire BrowserWindow and for us that's the whole app. What are you concerned about?

hoffmabc:
I mean we could do a brutal relaunch (i.e. set the relaunch setting, kill the app and reopen it) that is similar to what happens on auto-update, but this is not this feature.

morebrownies:
The user expects a quick refresh of the page here (like a browser) but it instead functions as a Restart Application. 

If i remember correctly, there was a long issues thread on v1 where people were confused by the refresh button restarting the entire application. If you're on a slow computer, it could take a few seconds for this event to finish.

morebrownies:
And the scren goes completely white during those few seconds.

rmisio:
@morebrownies Well, there's a third state there:

a.) refresh entire electron app (which also results in a and b happening).
b.) refresh entire browser app (which also results in c happening).
c.) refresh the current view / url.

I imagine you're after C because that will leave the global components in place (the top address bar, chat, etc... i.e. no flicker for them).

@hoffmabc To do that, you would:

change the backbone import to:

```
import Backbone, { View } from 'backbone';
```

then reload via: 

```
Backbone.history.loadUrl()
```

The reason I didn't suggest that right away is because on some pages, if you take this approach and the page doesn't end up fetching some server data and / or you're scrolled to the top... when you click 'refresh', the new page returns so fast, it looks like nothing happened.

To fix that, I was going to force the load to take at least 200 ms or something and work in some type of spinner / loading state.

Anyhow, that's more work and this didn't seem like some crucial thing for milestone 1, so I thought Brians reloading via option B would be ok for now.

hoffmabc:
This just refreshes the webcontents of the window so yes there'd be a white screen in the app window.

morebrownies:
That's correct @rmisio. C is the correct way of doing it.

morebrownies:
I'm fine with going with the quick solution for now, but we should log a bug to come back and fix it.

rmisio:
Yeah, but for now can we do B, which takes 10 seconds to implement vs. C that will take a few hours to implement correctly (if you want to avoid the scenario I described above where a super fast reload looks like nothing happened and like the button is broken)?

hoffmabc:
This is ready for review/merge. There will be no more updates on this AFAICT.

rmisio:
@jjeffryes This one looks good to me. Feel free to merge if it's ok with you.

jjeffryes:
I've filed the suggested improvements as issue #186

PR #150, opened 2016-12-05
Patch 132 to match address box design

jjeffryes:
- removes secondary color behind address boxes
- add padding
- uses col6 to space the address boxes
- adjusts the gutterH code to account for only child columns

att @morebrownies, take a look at this, if it looks good merge it with 132 and I'll merge 132.


morebrownies:
Nice, thank you. Totally missed this because I didn't have any addresses on file. 

We should add a heading above the **Name** input field too, e.g. **New Address**

screenshot 2016-12-05 10 28 58

morebrownies:
Once we move to the new modal design, we can put some more defined separation between existing address and the new address form.

PR #151, opened 2016-12-05
[WIP] Crash reporting

hoffmabc:
This reorganizes main.js for some of the process detection and enable crash reporting.

Please do not merge. We are experimenting/testing.

### DIFF
--- a/main.js
+++ b/main.js
@@ -274,6 +274,8 @@ function createWindow() {
         },
         {
           role: 'quit',
+          accelerator: 'CmdOrCtrl+Q',
+          click: function() { closeConfirmed=true; app.quit(); },
         },
       ],
     });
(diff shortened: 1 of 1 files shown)


rmisio:
@hoffmabc In my server-connect branch I've refactored parts of main.js and the starting of the local server quite a bit. If we're both working on that area, we're probably going to step on each other's toes.

Not sure how to best to handle this...? Maybe in the next day or so I'll push a PR from my branch (I should be done with a lot of the main.js changes) so we can get that into Master and you'd be integrating the crash reporting code into the already refactored main.js?

hoffmabc:
Probably best for you to just take the lead and incorporate whatever we come up with and decide here is appropriate. I'm fine with nuking this PR once that's happened.

rmisio:
This one is still on my radar, but I'll circle back to it when I'm more done with server connect. It looks like a minor thing to just integrate that crashReporter object.

rmisio:
integrated this into #206.

PR #158, opened 2016-12-05
Fix Inaccurate Follow Count

### LINKED ISSUE #200: Home: Follower count incorrect
screen shot 2016-12-13 at 10 07 56 am

jjeffryes:
- if the follower or following counts update in the profile model, the tab count will be updated without re-rendering
- when the followers or following collection is loaded for the follow tab, it updates the tab count if it is different
- the count retrieved from the server in the profile was not accurate (IIRC it's not updated very frequently because that triggers a publish)
- when the collection is updated the tab count will be updated

Closes #200

### DIFF
--- a/js/views/userPage/Follow.js
+++ b/js/views/userPage/Follow.js
@@ -40,6 +40,10 @@ export default class extends BaseVw {
         this.render();
       });
     }
+
+    this.listenTo(this.followCol, 'update', (col) => {
+      this.trigger('updateCount', col.length);
+    });
   }
 
   updateViewerFollowers() {
--- a/js/views/userPage/UserPage.js
+++ b/js/views/userPage/UserPage.js
@@ -53,6 +53,10 @@ export default class extends baseVw {
 
     this.listenTo(this.model.get('avatarHashes'), 'change', () => this.updateAvatar());
     this.listenTo(this.model.get('headerHashes'), 'change', () => this.updateHeader());
+    this.listenTo(this.model.get('followerCount'), 'change', () =>
+      this.updateFollowerCount(this.model.get('followerCount')));
+    this.listenTo(this.model.get('followingCount'), 'change', () =>
+      this.updateFollowingCount(this.model.get('followingCount')));
   }
 
   className() {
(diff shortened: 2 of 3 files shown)


js/views/userPage/Follow.js
```diff
@@ -40,6 +40,10 @@ export default class extends BaseVw {
         this.render();
       });
     }
+
+    this.listenTo(this.followCol, 'update', (col) => {
```
rmisio: This only works if we're not going to paginate the collection, which I believe we're going to do for at least one of the collections, right? (we really should paginate both, but that's a separate issue)

For any collection we paginate, we will need to rely on the server count. The user won't know it's off and if and when we fully fetch the collection, we could adjust the count then.

js/views/userPage/UserPage.js
```diff
@@ -132,17 +136,29 @@ export default class extends baseVw {
   }
 
   createFollowersTabView(opts = {}) {
-    return this.createChild(this.tabViews.Follow, {
+    const newView = this.createChild(this.tabViews.Follow, {
       ...opts,
       followType: 'Followers',
     });
+
+    newView.on('updateCount', (count) => {
```
rmisio: This could be written as one line, but I suppose that's a stylistic preference.

js/views/userPage/UserPage.js
```diff
@@ -132,17 +136,29 @@ export default class extends baseVw {
...
 
   createFollowingTabView(opts = {}) {
-    return this.createChild(this.tabViews.Follow, {
+    const newView = this.createChild(this.tabViews.Follow, {
       ...opts,
       followType: 'Following',
     });
+
+    newView.on('updateCount', (count) => {
+      this.updateFollowingCount(count);
```
rmisio: If we do decide to leave this event, can you only fire it if the count actually changes?

js/views/userPage/UserPage.js
```diff
@@ -53,6 +53,10 @@ export default class extends baseVw {
 
     this.listenTo(this.model.get('avatarHashes'), 'change', () => this.updateAvatar());
     this.listenTo(this.model.get('headerHashes'), 'change', () => this.updateHeader());
+    this.listenTo(this.model.get('followerCount'), 'change', () =>
```
rmisio: followerCount is just a number, what I think you want here is:

```
this.listenTo(this.model, 'change:followerCount', () => ...
```

jjeffryes:
Closed, we need to rethink how this will work.

PR #163, opened 2016-12-06
Output a placeholder when search criteria = 0 listings (fixes #157)

### LINKED ISSUE #157: Output a placeholder when the selected search filter(s) / search term results in 0 listings found
**How it currently looks**
screenshot 2016-12-05 11 52 47

**What we should aim for** 
screenshot 2016-12-05 12 13 29

morebrownies:
This fixes #157


rmisio:
I was just going to give you direction, but looking into it, it's a bit more complex than I could quickly explain. Is it alright if I submit a PR to this branch?

An open question... what do you want the experience to be if no listings are found, but there's not filters applied? Should it just say 'No Listings Found' and the button be hidden?

morebrownies:
> I was just going to give you direction, but looking into it, it's a bit more complex than I could quickly explain. Is it alright if I submit a PR to this branch?

Thanks @rmisio Yes, feel free to submit a PR. 

> An open question... what do you want the experience to be if no listings are found, but there's not filters applied? Should it just say 'No Listings Found' and the button be hidden?

Yeah, that sounds good to me.

morebrownies:
Ready for review. Thank you for the help @rmisio

rmisio:
@jjeffryes ready for your review, at your leisure.

styles/_variables.scss
```diff
@@ -35,6 +35,7 @@ $padSm: $pad / 2;
 $padMd: $pad * 1.5;
 $padMd2: $pad * 1.2;
 $padLg: $pad * 2;
+$padLg2: $pad * 4;
```
jjeffryes: We should keep numbered classes in sequence, so if we need something bigger than the huge class, we should make another class after Huge. Maybe Gigantic? (padGi) Or just end at huge and do padHg2, padHg3, etc.

Generally I tried to avoid the numbered size classes, and only used them when something had to fit between two other sizes.

styles/components/_layout.scss
```diff
@@ -257,6 +257,11 @@
   padding: $padLg;
 }
 
+.padBg2,
+.padBgKids > * {
```
jjeffryes: If we switch the variable, switch the class here to match it.

morebrownies: All set

morebrownies:
Naming is updated @jjeffryes.

styles/components/_layout.scss
jjeffryes: My fault, I was unclear. I meant update the name of the class here too to padGi.

(padBg should be padLg, I forgot to change it when I did a refactor a while back)

morebrownies: I'm not seeing padBg here. I think you mean padBg2 should be changed to padLg? Is that correct?

jjeffryes:
```
.padBg2,
.padBgKids > * {
  padding: $padGi;
}
```

Should be

```
.padGi,
.padGiKids > * {
  padding: $padGi;
}
```

And change the class in the template to match.

PR #173, opened 2016-12-06
Moves active tab style to mixin

jjeffryes:
- this moves the active tab style to the textColor mixin
- textColor() now has a 3rd parameter, for the tab color
- the other text colors now have a tab color set, so if a tab has clrT3, for example, when active it will have a color one step up.
- currently the tab color is the same as the link color, but this provides the option to set it to something different

### DIFF
--- a/styles/_mixins.scss
+++ b/styles/_mixins.scss
@@ -71,7 +71,7 @@
   }
 }
 
-@mixin textColor($col1, $col2: $col1) {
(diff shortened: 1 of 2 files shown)


jjeffryes:
@morebrownies this is an update to your other PR, it moves the active tab into the text color mixin.

morebrownies:
Ok, seem to be working well @jjeffryes. I merged into 142. Thank you

Issue #75, opened 2016-10-18
Create Listing Card


jjeffryes:
Wiki Page: https://github.com/OpenBazaar/openbazaar-desktop/wiki/Listing-Card

![image](https://cloud.githubusercontent.com/assets/1584275/19487994/fb939cba-9532-11e6-8bf2-9079a5bd16f3.png)

![image](https://cloud.githubusercontent.com/assets/1584275/19488453/f1642dfc-9534-11e6-9917-6958761a5849.png)

The Listing Card displays basic information about a listing. It should have the following data:
- [x] the default image of the listing (Small, Medium on retina devices)
- [x] the title of the listing.
- [x] the price of the listing
- [x] the average rating of the listing and the number of total reviews
- [ ] if it can ship to one of the user's addresses
- [ ] if the Card is a paid placement (ad)
- [ ] if the Card is NSFW, and the user has "don't show NSFW" set, the Card should be blurred or not shown (not yet determined)

When shown in the User Page Store tab, the Listing Card should also show:
- [ ] whether it has free shipping
- [ ] whether it is pinned
- [ ] whether it is hidden

Outside the User Page Store tab, the Listing Card should also show:
- [x] the avatar of the seller
- [ ] the country the listing ships from, if physical
- [ ] the country the listing applies to, if service

Interactive Elements
- [x] The title is clickable, clicking on it will open the Listing Detail modal.
- [ ] hovering over the country it ships from should show the name of the country (probably in a tooltip, not determined yet)

ekerstein:
@jjeffryes I see you have "the average rating of the listing and the number of total reviews" checked above. The number of reviews doesn't seem to be showing in OB2.

I was about to open an issue asking for this feature, but you seem to already have it in mind above (love the layout by the way). Has this been implemented or still being worked on?

For example: ob://@BTCEmporium/store/classic-pc-premium-tee

This item has a review but nothing shows up on it's listing card with regard to how many reviews:

![screen shot 2017-10-01 at 5 58 15 pm](https://user-images.githubusercontent.com/6055508/31060156-945fb7ba-a6dc-11e7-83cd-008c7ffc4b60.png)

jjeffryes:
The feature has been implemented, you don't see it because the number of reviews isn't available in search results for the OB1 search engine currently. Once a search provider returns that data, the reviews will be visible.

If you look at the same item in the store of the item's owner, it will show the number of reviews.

ekerstein:
[...]

PR #175, opened 2016-12-07
Improves listing card (fixes #135, #126, #124)

### LINKED ISSUE #124: The listing card deleted and deleting state need styling
@morebrownies Could you also style the Deleting and Deleted state of the listing card? You could see them by applying .listingDeleting or .listingDeleted to the .listingCard element.

The styles were very much placeholder styles before, but now, with the removal of the overlay panel, they are even worse:

before:

image

now:

image

Anyhow, they're important for a few reasons:

a.) The delete is an async process and can complete at any time - the user may be happily scrolling further down the list when the delete completes, so we don't want to just remove the item having the page jump abruptly.
b.) Removing a listing may affect filters, e.g. a category may need to be removed from the cat filter. Having this update on the fly is a bit of a can of worms to make the experience smooth, so we've decide to take the approach of popping in a message urging the user to refresh:

image

c.) The delete could be initiated elsewhere, namely the listing detail modal. If it takes a bit, the user could close the modal landing back on the store. We would want a "deleting" state so the user understands the listing is in the process of being deleted.

### LINKED ISSUE #126: Please style the Listing Data Changed <refresh> pop in message.
When on your store and listing data changes, a pop in message appears urging you to refresh. The message is not super visible right now and is easily missed.

It's currently set with fixed positioning. The fixed positioning is important because the user may be scrolled very far down the store when we need to show the message, so having it be fixed will allow it to be at the top of their view no matter what scroll position they're at. Hopefully, the fixed positioning can stay. All else though is probably fair game.

image

### LINKED ISSUE #135: Improve the design of the listing card
**Changes needed**
- The edit/delete buttons could be simplified (my fault for designing them to be so large)
- Add a slight transition on hover
- ~~Needs border radius~~
- #124 
- Delete confirmation

**OLD**
screenshot 2016-12-02 11 45 36

**NEW**
screenshot 2016-12-02 11 48 17

morebrownies:
**Changes**
- I changed the styling and placement of the **edit/delete buttons** within the listingCard (fixes #135)
- I touched up the **refresh pop in**. I think it works well as is @rmisio. (fixes #126)
- I touched up the **deleting/deleted** experience a bit. Again, mostly good as is. (fixes #124)
- I also cleaned up the **listView** to ensure the edit/delete buttons are useable 

Note: I'm going to open a separate issue to add in the delete confirmation.


js/templates/listingCard.html
```diff
@@ -2,15 +2,17 @@
 <div class="gridViewContent">
   <div class="listingImage" style="background-image: url(<%= ob.getServerUrl(`ipfs/${ob.isHiRez() ? ob.thumbnail.medium : ob.thumbnail.small}`) %>), url('../imgs/defaultItem.png')">
       <% if (ob.ownListing) { %>
-        <div class="editOverlay overlayPanel">
-          <div class="overlayPanelInner clrS"></div>
-          <a class='btn clrP clrBr js-edit'><span class="ion-edit"></span></a>
-          <a class='btn clrP clrBr js-delete'><span class="ion-trash-b"></span></a>
+        <div class="editOverlay overlayPanel pad">
+          <div class="overlayPanelInner"></div>
+          <div class="overlayPanelButtons">
```
morebrownies: @jjeffryes I needed to add overlayPanelButtons wrapper in order to set a z-index so that the edit/delete buttons display over overlayPanelInner. Without the z-index, the buttons aren't clickable.

styles/components/_buttons.scss
```diff
@@ -58,6 +62,10 @@
   font-weight: normal;
   height: auto;
 
+  &:hover {
```
morebrownies: This was added to prevent the radio button wrapper from getting a border

styles/components/_listingCard.scss
```diff
@@ -3,6 +3,9 @@
   height: 283px;
   font-size: $tx5b;
   cursor: pointer;
+  position: relative;
+  bottom: 0;
+  transition: all .2s ease;
```
morebrownies: I set the transition here to **all** because the bottom param changes on hover. Is it possible to do this without setting **all**?

morebrownies:
Ready for review

hoffmabc:
Borders are still whack if this was supposed to fix it.

image

morebrownies:
This PR doesn't address that issue. Josh was working on it in another branch.

js/templates/listingCard.html
rmisio: hmm.. not sure what happened. The overlayPanelInner and the buttons coexisted for weeks. Something else must have changed if another wrapping element is needed.

morebrownies: That's very true, it's working in master. I look around to see what may have caused it.

styles/components/_listingCard.scss
jjeffryes: Yes, that should be possible.

On general comment, don't put colors on anything in any file except theme. If you do (like the border color on the listing card), that color cannot be changed by the user's custom theme color.

We'll have to make a more generic "change from x to y border color on hover" class in theme.scss for the listing card. I'll do that in my patch.

styles/components/_listingsGrid.scss
```diff
@@ -17,9 +17,31 @@
     .listingCard {
       margin: 0;
       border-bottom-width: 0;
+      border-radius: 0;
+      padding: 6px;
+      -webkit-box-sizing: border-box;
```
jjeffryes: You don't have to use the -webkit prefix for box sizing.

jjeffryes:
@morebrownies take a look at #212 and merge when you're ready, it will resolve the conflicts and makes various updates to the code.

morebrownies:
@jjeffryes I merged in your patch. Are we good to go with this one?

tenthhuman:
Closes #75

Issue #74, opened 2016-10-18
Create Listing Detail


jjeffryes:
Wiki Page: https://github.com/OpenBazaar/openbazaar-desktop/wiki/Listing-Details

![image](https://cloud.githubusercontent.com/assets/1584275/19492904/a3c08034-9546-11e6-9880-d111e419c5a7.png)

The Listing Details view is a modal. It should be created and shown when:
- [x] the user clicks on a Listing Card's listing link in the UserPage Store tab
- [ ] the user clicks on a Listing Card's listing link in a Channel
- [x] the user clicks on a Listing Card's listing link in Search Results
- [x] the user enters the address of the listing in the address bar and presses the enter key
- [x] the user clicks on an external link with the address of the listing
- [x] the route for the listing is navigated to for any reason

The Listing Detail modal should show the following information:
- [x] the listing title
- [x] the listing price in the user's chosen currency
- [x] the listing type (physical, digital, service)
- [x] the listing condition
- [x] the item description
- [x] the default image for the listing (Medium, Large for retina screens)
- [x] the other images, if any, for the listing (shown at reduced size, but loaded at Medium/Large size)
- [ ] the attachments, if the listing is digital
- [x] shipping information
- [x] the terms and condition
- [x] the return policy
- [x] the average review rating
- [x] the total number of reviews
- [x] whether the listing has free shipping (if physical)
- [x] the reviews
- [x] the tags

The following elements are interactive:
- [x] clicking the Buy Now button will start the purchase flow.
- [x] if the listing has variants, they will be shown as select lists, with the default for each variant selected.
- [x] if variants are out of stock, they are shown, but are grayed out and can not be selected
- [x] if clicked on, the small images for the listing will be shown at a larger size (still being determined)
- [x] the large image can be zoomed (still being determined)
- [x] the Load More button will load additional reviews when clicked
- [x] the close button closes the modal, per the standard modal behavior

If the user is the owner of the listing:
- [x] an edit button is shown. Clicking on it opens the edit modal for this listing.

jjeffryes:
This was completed.

jjeffryes:
This is complete other than features that will be detailed in new requirements issues.

PR #180, opened 2016-12-09
Listing detail

jjeffryes:
Initial PR for the Listing Detail view. This formats the listing detail view to match the current design and adds in all currently available data and functionality.

Known Issues:
- the sections with data from the reviews, variants, and attachments have been commented out and are mostly just stubs, as that data isn't available yet.
- the drag library for dragging the large photo in the photo section doesn't handle a background image set to background-position: center correctly. It will need to be updated or replaced.
- the scroll effect on the photo thumbnails is a placeholder.


hoffmabc:
If I edit from the Listing View page the View Listing link no longer works.

image

1) Visit Listing View
2) Click Edit Listing
3) Edit and Save
4) Click View Listing at bottom.

jjeffryes:
@hoffmabc good catch. I didn't update that from Rob's existing code, I'll take a look at it.

hoffmabc:
Listing isn't viewable in List Mode.
image

None of the links work or details button.

rmisio:
> If I edit from the Listing View page the View Listing link no longer works.

This was a known issue to me. The link appears not to work because you're address is actually for the listing detail of the listing already (the detail modal is behind the edit modal). Since you're already at that address, clicking on the same address does nothing.

I have some ideas how to handle that case. @jjeffryes I'll look to put a patch in early next week.

rmisio:
I'm thinking maybe these arrows would work better if they showed themselves on hover? For touch devices we could have them always showing.

I don't know if that highlight color works here. The green looks nice on the Buy Now button, but IMO just looks really weird here. I know this is configurable, but do we want the default color to look so odd here?

@morebrownies What do you think?

morebrownies:
@rmisio which controls are you referencing?

rmisio:
Josh, I had this pop in message working in a way in which it was fixed, so as you scroll it stays in position in the viewport. That behavior seems to have changed. Can you look into why?

I think its good for it to be fixed because Editing the listing is an async process and the user may be further scrolled down the details modal when it completes and they could miss the message. Plus I think it's kind of good it's in their face, so if they notice a discrepancy they know why and could refresh at their leisure.

The fixed positioning works properly on the Store, but something here may have gotten borked (perhaps with the new CSS styling related to the scrolling of the modal).

rmisio:
> @rmisio which controls are you referencing?

@morebrownies Added a screenshot to the comment.

js/views/modals/listingDetail/Listing.js
```diff
@@ -127,6 +202,20 @@ export default class extends BaseModal {
     }
   }
 
+  setShippingDestination(e) {
```
rmisio: Kinda minor, but I would name this `onSetShippingDestination`. It expects a jQuery event object, so it's really only designed to be used as an event handler. Without the 'on' part, it seems like an API method of the view to set a shipping destination.

rmisio:
Josh, you mentioned the image dragging plug-in was quirky... anyways:

1.) You already know about the flicker issue.
2.) It also let's you drag beyond the bounds of your image:

image

Also, if an image is draggable, it would be nice for the cursor to change to some type of panning thing, otherwise it's hard to tell this functionality exists.

rmisio:
Minor thing... but I'd give more bottom padding to this modal, so at least the status bar isn't obscuring it:

image

js/utils/draggable_background.js
```diff
@@ -0,0 +1,161 @@
...
+ *
+ * v1.2.4
+ *
+ * Copyright (c) 2014 Kenneth Chung
+ *
+ * Licensed under the MIT license:
+ *   http://www.opensource.org/licenses/mit-license.php
+ */
+
+/* eslint-disable */
```
rmisio: I'm assuming this isn't available via npm?

Anyhow, for 3rd party code that we need to have in our code base (e.g. not npm-able), I have a js/lib folder for that, which I believe is set so the entire folder is ignored by eslint.

js/views/modals/listingDetail/Listing.js
```diff
@@ -168,13 +282,30 @@ export default class extends BaseModal {
         shipsFreeToMe: this.shipsFreeToMe,
         ownListing: this.model.isOwnListing,
         displayCurrency: app.settings.get('localCurrency'),
+        shipsFromCountry: app.settings.get('country'),
```
rmisio: Perhaps I don't understand the reqs well, but should the ships from field really display the users local country? Instead wouldn't ships from be the country the listing ships from?

js/views/modals/listingDetail/Listing.js
```diff
@@ -168,13 +282,30 @@ export default class extends BaseModal {
...
+      this.$photoSelected.backgroundDraggable();
+
+      const shippingDest = this.$('#shippingDestinations');
+
+      shippingDest.select2({
+        data: this.select2CountryData,
+        placeholder: app.polyglot.t('listingDetail.shipToPlaceholder'),
+      });
+
+      shippingDest.val(app.settings.get('country')).trigger('change');
```
rmisio: Do we need to trigger the event here? Can't we just directly call the function here? You'd probably need to change the function signature so it accepts the data it needs rather than a click handler and then seperate the click handler into a separate function that calls the other one. Seems like a much cleaner way to me than synthetically firing a change event.

js/views/modals/listingDetail/Listing.js
```diff
@@ -168,13 +282,30 @@ export default class extends BaseModal {
...
         displayCurrency: app.settings.get('localCurrency'),
+        shipsFromCountry: app.settings.get('country'),
       }));
 
       super.render();
 
       this._$deleteListing = null;
       this._$shipsFreeBanner = null;
       this._$popInMessages = null;
+      this._$photoSection = null;
```
rmisio: There's a bunch more you need to reset to null here, basically anything that you're getting via this type of getter:

```
  get $shippingOptions() {
    return this._$shippingOptions ||
      (this._$shippingOptions = this.$('.js-shippingOptions'));
  }
```

Otherwise, upon re-render, those will point to elements that are no longer in the DOM.

That is why currently if you Edit the listing's shipping options and click 'refresh' in the pop-up message, the shipping options are ends up being empty:

image

rmisio: A local pickup setting option looks a little bare-bone to me. I think we need some other text / indicator to make it clearer what's going on:

image

paging senior lobo @morebrownies

rmisio:
@jjeffryes We got some conflicts that need some attention.

js/languages/en-US.json
```diff
@@ -155,7 +155,35 @@
...
+    "reviews": "Reviews",
+    "attachments": "Attachments",
+    "refundPolicy": "Refund Policy",
+    "termsAndConditions": "Terms Of Service",
+    "shipping": "Shipping",
+    "shipsFrom": "Ships from %{country}",
+    "shipTo": "Ship to:",
+    "shipToPlaceholder": "Select a Country",
+    "noShippingFound": "This listing does not ship to the selected country.",
+    "formats": {
```
jjeffryes: I was kind of torn on this too. I think in this case, since they are referring to labels that are repeated elsewhere, it would be ok to move them outside the sections and use a single set in multiple places.

js/templates/listingCard.html
```diff
@@ -38,7 +38,7 @@
       </div>
       <% if (ob.shipsFreeToMe) { %>
       <div class="freeShipCol flexNoShrink txCtr">
-        <span class="clrE1 clrTOnEmph txTn txCtr upper padTn">Free Shipping</span>
+        <span class="clrE1 clrTOnEmph txTn txCtr txUp padTn">Free Shipping</span>
```
jjeffryes: No reason, I just missed this one. Good catch.

js/views/modals/listingDetail/Listing.js
jjeffryes: Do we have that data? I don't see a ships from parameter in the listing data.

js/views/modals/listingDetail/Listing.js
hoffmabc: Senior wolf or Señor wolf?

js/views/modals/listingDetail/Listing.js
jjeffryes: This is how the select2 docs say to do it. How would changing the function make it simpler?

js/utils/draggable_background.js
jjeffryes: It's old, and not available from npm. It's de-linted because it's a placeholder, it's going to need some tweaking or to just be replaced. 

I can move it to the lib folder until then.

js/views/modals/listingDetail/Listing.js
jjeffryes: I forgot to go back and add the nulls, thanks for catching that.

js/views/modals/listingDetail/Listing.js
rmisio: I don't think we have that data, but until we do, I would just hide the field rather than putting in the users Local Country, which IMO makes no sense here.

js/views/modals/listingDetail/Listing.js
rmisio: I think select2 is doing a poor practice. Why would a module, instead of calling one of its functions directly, take the circuitous route of programmatically firing an event handler and having the code path go from point A to B to C to D... (remember, it's going to go through jQuery event handling code) instead of point A to B.

I won't hold up the merge if you don't think it's important, I'm just flagging what IMO is not clean code.

rmisio: I'm a little confused.

> To elaborate, what we really want here is a way to set the selected element when select2 is initialized

Well, that has nothing to do with synthetically firing the 'change' event which is what I'm flagging as a poor practice.

I'm just suggesting changing this:

```
shippingDest.val(app.settings.get('country')).trigger('change');
```

to

```
shippingDest.val(app.settings.get('country'));
setShippingDestination(app.settings.get('country'));
```

That also requires the `setShippingDestination` signature to accept the data it needs rather than a jQuery event. An `onSetShippingDestination` handler should accept the jQuery event and then call `setShippingDestination`.

==========

Anyways, `shippingDest.val(app.settings.get('country'))` part doesn't exactly set the default value on initialization, but it does set it immediately afterwards, which should accomplish the same thing.

If you really want to set it on initialization, look at how I build the select2 markup in the EditListing and Settings modals. Essentially, I don't pass any data into the select2 element at initialization. I just build the <option> elements in the template and set the default one by giving it a "selected" attribute. Then, when I do initialize select2, it builds it's data and default selection from the markup.

morebrownies:
Ok I just had a chance to get caught up on this PR. 

A few things to call out...

1. I agree, the arrows within the photo carousel probably don't need to inherit the highlight color. Could we hardcode them to white? 

2. The draggable image zoom seems very buggy. I'm having a difficult time using it. I agree with @rmisio that we should change the cursor to a pointer.

3. I also agree that the bottom of the overlay needs a lot more margin so that it doesn't interfere with other elements.

4. The main photo should be clickable. On click it should scroll you to the carousel.

screenshot 2016-12-13 09 34 28


5. The overlay does not appear to be horizontally centered. Notice in the screenshot that the store grid in the foreground is visible to the right of the overlay. The two layers should stack perfectly since they're the same width.

screenshot 2016-12-13 09 41 13

6. The Refund policy, terms of service, description and review sections should all display even if they're empty. As a potential customer, it's helpful to know when there is no return policy vs not showing it at all. 
screenshot 2016-12-13 09 46 56

7. Small thing... Can we cut the vertical margin between the address bar and modal in half? 

screenshot 2016-12-13 09 49 50
 
Looks great overall @jjeffryes. Nice work. I'd like to clean up some spacing and finer details, but I can work that in after you guys finalize the PR.

morebrownies:
There's also a white space displaying at the bottom of the photo carousel  

screenshot 2016-12-13 10 13 13

morebrownies:
I noticed the close icon displays differently based on if I'm viewing my listing vs viewing another store's listing.

**My store (correct)**
screenshot 2016-12-13 10 23 30 

**Washo's store (incorrect)**
screenshot 2016-12-13 10 23 23

hoffmabc:
agree on that @morebrownies noticed it and forgot to file it.

drwasho:
screen shot 2016-12-14 at 2 02 29 pm

Might be worthwhile keeping the right arrow when you reach the last image so that it can take you back to the first image.

+1 to changing the color of the arrow
+1 to having the arrows appear only when you hover over

drwasho:
Based on the last PR, this is popping up when I open any listing:

screen shot 2016-12-16 at 2 35 16 pm

drwasho:
I've been reluctant to bring this up, but if we're going to _decide_ to fix this, perhaps it should be now.

screen shot 2016-12-16 at 2 36 35 pm

There is an extraordinary amount of white space in the top section of the listing, probably as a function of the size of the initial listing image. Even with a reasonable amount of tags to occupy some of this white space, it still looks like we're wasting a lot of space.

Perhaps I'm making this observation prematurely, was there other data meant to be in this space that we haven't added _yet_ @morebrownies ?

drwasho:
Also the borders have an inconsistent thickness.

rmisio:
The arrow navigation on the photos doesn't work. There are exceptions in the console:

image

Also, there is some extraneous white-space at the bottom of the image (under the thumb bar):

image

Also, still no cursor change on hover to indicate pan-ability.

But, overall the panning works much better than the initial version +1

rmisio:
Since description is not a required field, can we get some type of "empty" state?

image

rmisio:
The thumbs have some extra top and bottom space and it looks like the image is repeating into those spaces:

image

.gitignore
```diff
@@ -1,5 +1,5 @@
 node_modules
-npm-debug.log*
+npm-debug.log
```
rmisio: I think Cris S put in the wildcard because he was getting some npm debug logs with random chars at the end, e.g. `npm-debug.log23123123`.

js/templates/modals/listingDetail/listing.html
```diff
@@ -1,11 +1,187 @@
...
-<% if (ob.ownListing) { %>
-<div>
-  <a class="btn js-editListing">Edit</a>
-  <a class="btn js-deleteListing">Delete</a>
+
+<div class="modalContentInner">
+  <div class="listingControls flex">
+    <% if (ob.ownListing) { %>
+    <div class="btnStrip clrSh3">
+      <a class="btn js-editListing clrP clrBr">Edit</a>
```
rmisio: Can we translate these please?

js/templates/modals/listingDetail/listing.html
```diff
@@ -1,11 +1,187 @@
...
+        </div>
+      </div>
+    </div>
+
+    <div class="contentBox padLg clrP clrBr clrSh3">
+      <h2 class="txUnb"><%= ob.polyT('listingDetail.description') %></h2>
+      <%= ob.item.description %>
+    </div>
+
+  <div class="flexRow gutterH" id="avatarCropper">
```
rmisio: I think this is the block @drwasho pointed out that shouldn't be here. Looks like a missed copy-paste cleanup.

js/views/modals/listingDetail/Listing.js
```diff
@@ -103,6 +120,67 @@ export default class extends BaseModal {
...
+    this.setSelectedPhoto(targetIndex);
+    this.setActivePhotoThumbnail(targetIndex);
+  }
+
+  onClickFreeShippingLabel() {
+    this.gotoShippingOptions();
+  }
+
+  gotoShippingOptions() {
+    this.$el.animate({
```
rmisio: Absolutely no need to change this, but as an FYI, I did bring [Velocity](http://velocityjs.org/) into the codebase and am using it in some places. There is a utility module to pull it in:

```
import '../../../utils/velocity';
```

I also have one for the additional [UI pack](http://velocityjs.org/#uiPack), which has some cool pre-packaged effects available:

```
import '../../../utils/velocityUiPack.js';
```

Anyways, Velocity gives much smoother animations than jQuery and has a dev friendly API.

js/views/modals/listingDetail/Listing.js
```diff
@@ -168,13 +287,41 @@ export default class extends BaseModal {
...
+        this.$photoSection.cropit({
+          smallImage: 'stretch',
+          allowDragNDrop: false,
+        });
+
+        this.$photoSection.cropit('imageSrc', initialPhoto);
+      }, 0);
+
+      this.$('#shippingDestinations').select2({
+        placeholder: app.polyglot.t('listingDetail.shipToPlaceholder'),
```
rmisio: Why is this placeholder necessary? Will there ever be a scenario when it's shown?

rmisio:
I'm thinking we should remove the hover state from the tags, since AFAIK clicking on the tag does nothing:

image

morebrownies:
> There is an extraordinary amount of white space in the top section of the listing, probably as a function of the size of the initial listing image. Even with a reasonable amount of tags to occupy some of this white space, it still looks like we're wasting a lot of space.

@drwasho the right side could vary in height quite a bit. The screenshot you posted is literally the bare minimum a listing can have. We're also not outputting all the data yet. Currently we're not outputting review information under the Buy now button. 

screenshot 2016-12-16 09 34 09

We should also have an empty state when no tags are entered. This will eat up a bit of space too.

screenshot 2016-12-16 09 18 48

On the complete opposite side of the spectrum, a listing that has a few variants choices and a handful of tags will eat up a majority of the right side. 

screenshot 2016-12-16 09 26 05

I'm sure we can continue to improve on this too. Open to suggestion if you have any. Is there any additional data you would like to see outputting on the right side? Let me know and I can try to work it in.

morebrownies:
> I'm thinking we should remove the hover state from the tags, since AFAIK clicking on the tag does nothing:

@rmisio clicking tags will take you to 3rd party search and display listing that also use that tag.

morebrownies:
Question about the Photo Carousel @jjeffryes @rmisio...

The photo(s) in the carousel appear to automatically zoom in by default and I'm thinking this may not match the expected behavior. I initially was expecting to first see the full view of the photo (as pictured below) with the ability to hover/click to zoom in on certain sections of it. 

Thoughts? 

**Expected**
image-zoom-expected 2x

**Current**
image-zoom-current 2x

jjeffryes:
@morebrownies I'm actually using the same cropit library we use for image cropping here, with the file input turned off, so we could easily switch to zoomed out with a zoom bar. 

I thought zoomed in looked nicer.

morebrownies:
I like the ability to zoom too, but I find it confusing when it's zoomed by default. It's especially a bad experience when viewing small photos. 

Here's an example of what a 200px by 200px image looks like be default:
screenshot 2016-12-16 13 39 56

It would be better if it displayed in its default state (like so), with the option to click to zoom. 
screenshot 2016-12-16 13 41 23

js/views/modals/listingDetail/Listing.js
```diff
@@ -103,6 +121,81 @@ export default class extends BaseModal {
...
+    this.setSelectedPhoto(targetIndex);
+    this.setActivePhotoThumbnail(targetIndex);
+  }
+
+  onClickFreeShippingLabel() {
+    this.gotoShippingOptions();
+  }
+
+  gotoShippingOptions() {
+    this.$shippingSection.velocity(
```
rmisio: Don't forget to import velocity:

```
import '../../../utils/velocity';
```

The only reason it's working is because the Store imports it which adds it to the jQuery object, but if the Store were to change and not needed, this module would break. Much safer for this module to also declare it as a dependency.

rmisio:
@morebrownies Do we need an 'Any country' option here? Is it a reasonable use case that I would want to see everywhere a listing ships to?

image

--------------

@jjeffryes I'll ask one more time and then forever hold my peace... still no cursor to indicate panning is possible on the images.

--------------

There's no hover state on the thumbs. I would think we should have the cursor be hand (like other CTAs in the app) and / or a border change or something.

image

--------------

@hoffmabc If there's any impetus to merging this sooner rather than waiting until Josh gets back, let me know. The remaining issues from my end are minor and I could log bugs for them.

Of course, in either case, I would also wait on @morebrownies approval before merging.

morebrownies:
@jjeffryes @rmisio here's a prototype for the suggested photo carousel changes. I used jQuery Zoom (http://www.jacklmoore.com/zoom/) + 15 lines of js to get working. 

Notice the photos are not zoomed by default, but they can be clicked to zoom, and the cursor can be used to focus on different parts of the photo.  

It feels pretty good to me. Thoughts? 

https://jsfiddle.net/65shu8xq/13/

tenthhuman:
Closes #74

PR #215, opened 2016-12-15
Make avatar/header size warning less intimidating (189)

### LINKED ISSUE #189: Make avatar/header size warning less intimidating
Reduce the "in your face-ness" of the warnings when a user uploads an image that is too small.

morebrownies:
**Changes**
- Made the Avatar in Settings round instead of square (it's round everywhere else but here)
- Updated copy
- I cleaned up the photo size tips. They're not as overwhelming anymore. (partially fixes #189)

screenshot 2016-12-15 11 51 08

screenshot 2016-12-15 11 52 26

**To do**
The discussion in #189 questions how we should treat zoom on small photos. Should we disable it? Should we completely hide it? Should zoom always be available? 

I did some research to see what other services are doing in this scenario and Twitter seems to go with the latter option; zoom is always available. In the screenshot below, I uploaded an abnormally small photo to Twitter and they still allowed me to zoom. 

screenshot 2016-12-15 11 56 39

I also tried many large photos from unsplash.com in OB 2.0 and in almost all cases I wasn't able to zoom on my cover photo. I say we ditch the restraint for now and see how it goes. 

screenshot 2016-12-15 12 01 06


js/templates/modals/settings/page.html
```diff
@@ -79,10 +79,11 @@ <h2 class="h3 txUnb"><%= ob.polyT('settings.pageTab') %></h2>
     <div class="flexRow">
       <div class="col3">
         <label><%= ob.polyT('settings.avatar') %></label>
+        <div class="clrT2 txBase"><%= ob.polyT('settings.loadAvatarHelp', {minWidth: ob.avatarMinWidth, minHeight: ob.avatarMinHeight}) %></div>
       </div>
       <div class="col9">
         <div class="flexRow gutterH" id="avatarCropper">
-          <div class="contentBox avatarPreview clrP clrBr flexNoShrink js-avatarPreview"></div>
+          <div class="contentBox avatarPreview clrP flexNoShrink js-avatarPreview"></div>
```
jjeffryes: If we take the border color off here, and you upload a photo with a white background, you can't see the circle any more (if you have the primary color still set to the default). 

image

styles/modules/modals/_settings.scss
```diff
@@ -14,14 +14,21 @@
     }
   }
 
+  .avatarPreview {
+    border: 0;
+
+    .cropit-preview-image-container {
+      border-radius: 60px;
```
jjeffryes: It would be slightly safer to set the radius to 50%, that is always a circle no matter what the actual size is.

js/templates/modals/settings/page.html
morebrownies: Good point. However, when I leave the border in there, it results in this:

screenshot 2016-12-16 15 01 55

I can't figure out where to put the border color so that it wraps the circle instead of the container...

morebrownies: Something with cropit-preview is causing the issues

styles/modules/modals/_settings.scss
morebrownies: Cool, thanks. Never thought to use percentages on border-radius for some reason. Always thought it had to be pixels. No idea why. 🤔

js/templates/modals/settings/page.html
jjeffryes: Add this CSS:

```
. settings .avatarPreview {
border-width: 1px; (or 2px, 1px was pretty thin when I tried it)
border-style: solid;
border-radius: 50%;
}

That will make the container that wraps the preview have the same shape, and the clrBr class will work.

morebrownies:
ok looking much better now

screenshot 2016-12-16 16 12 36



@jjeffryes can we disable this warning?  we can do without it 

screenshot 2016-12-16 16 06 41

jjeffryes:
This looks good, though the border around the preview is hard to see on my monitor, we may want to switch it to a different color or shadow later. 

image

PR #220, opened 2016-12-16
Upgrade editor closes #95

### LINKED ISSUE #95: Replace the Medium Editor with a Better One
The current medium text editor used in the settings and edit listings views isn't ideal. It should be replaced by a WYSIWYG editor with the following criteria:
- is WYSIWYG, either as you type or with a preview button.
- adds clean and minimal HTML to the text.
- can be applied to a textarea element, and keeps the content of that textarea element in sync with the HTML it generates, so saving the content requires no or very few extra steps.
- is open source, with an MIT or compatible license.
- can restrict the allowed tags (see list below) to match the tags allowed by the server.

Allowed Tags:
- h2
- h3
- h4
- h5
- h6
- p
- a - can have the href, title, and alt attributes, http, https, ftp, mailto, and ob2 schemes
- u
- ul
- ol
- nl
- li
- b
- i
- strong
- em
- strike
- hr
- br
- img - can have the src and style attributes, is assumed to be an external image
- blockquote
- span

The following text editors are suggested:
- [Trumbowyg](http://alex-d.github.io/Trumbowyg/)
- [Quill](https://github.com/quilljs/quill)

tenthhuman:
**Complete** 👍 👍 🌮 

- **Upgrade:** A new editor ( [trumbowyg ](https://alex-d.github.io/Trumbowyg/)) is installed in 1 location: Listing Item Description.
- **Hygeine:** The install attempts to prevent XSS vectors and maintain data hygeine by conforming to our set of allowed tags, attributes and schemes.
- **Affordance**: The tags are filtered ( currently 2 seconds ) after a human has typed, when the data is saved, and when it is rendered. A byline explains we modify the HTML. 

**Excepting** 👻 

- **Unhappy ghost:** Nothing for the ghost to find this time. Requirements for #95 are met. Ghost is on the lookout.

- **XSS**: XSS vulnerability existed when using textarea because `TEXTAREA` [converts entities to characters](http://stackoverflow.com/a/8519356/7003753) which allowed attack vectors to hide from sanitization by using suitable escapes. For this reason changes were made to replace textarea with `DIV[contenteditable]` where editor operates. See the commit prior to this pull request ( fbea701e27bed32cc8eb6c4b38ee8375b2b515e4 ) for more details. 
- **Allowed tags explained**: Currently the link to explain the tags we allow is pastebin. It may work to link to a markdown file on GitHub.


js/utils/dom.js
```diff
@@ -19,3 +19,41 @@ export function isScrolledIntoView(element) {
   // Check its within the document viewport
   return top <= document.documentElement.clientHeight;
 }
+
+export function throttle(handler, fps) {
```
rmisio: These are already both provided by [underscore](http://underscorejs.org/) which is a dependency of backbone and a lib we very heavily utilize. Can we just use those?

js/views/modals/editListing/EditListing.js
```diff
@@ -18,6 +20,29 @@ import app from '../../../app';
 import BaseModal from '../BaseModal';
 import ShippingOption from './ShippingOption';
 
+$.trumbowyg.svgPath = '../node_modules/trumbowyg/dist/ui/icons.svg';
```
rmisio: Since this is a global setting, IMHO it shouldn't be made here, which is just one module using the editor.

I would probably make a wrapper module that does this - since modules are cached, it will only execute once:

```
// js/utils/trumbowyg.js
import 'trumbowyg';
$.trumbowyg.svgPath = '../node_modules/trumbowyg/dist/ui/icons.svg';
```

then any module / view that wants to use the plugin could import our wrapper:

```
import 'js/utils/trumbowyg';
```

package.json
```diff
@@ -36,7 +36,7 @@
     "babel-preset-es2015": "^6.9.0",
     "babel-preset-stage-1": "^6.5.0",
     "babel-register": "^6.9.0",
-    "browser-sync": "^2.13.0",
+    "browser-sync": "^2.18.5",
```
rmisio: Why are we raising the browser-sync and node-sass version numbers? We have to be a little careful when updating packages since it may break existing stuff (and I'm not sure what these packages have to do with this request).

js/views/modals/editListing/EditListing.js
```diff
@@ -18,6 +20,29 @@ import app from '../../../app';
...
+
+function installHTMLSanitizationWatcher() {
```
rmisio: Is it really necessary to sanitize the HTML as the user is typing (or 2 seconds after)?

One problem is that when you update the field like that, the cursor position is lost, which is annoying. Let's say I was typing and took a two second or more break to read what I typed. Now, suddenly the cursor position jumps somewhere else when the text is sanitized.

The second issue is that, IMHO, this adds a whole lot of complexity to the code (this is not an easy function to follow) with a very low ROI.

I would opt for just initializing the plugin on render (`this$('#editListingDescription').trumbowyg();`) and just sanitizing on save.

rmisio: Along the same lines, I'm not sure this messaging (`We modify your HTML...`) is really necessary, particularly if we were just to sanitize on save:

image

IMO, sanitization just seems like something that we should take care of behind the scenes and the user doesn't really need to be concerned with it. Our set of [allowed tags](http://pastebin.com/Fy31muVC) is pretty inclusive. I would think the user would really have to be doing some advanced / questionable stuff for sanitization to even come into play.

jjeffryes:
I'm going to close this for now, since we realized the requirements weren't nailed down yet.

Cris, if you want to change it to just the editor, with no sanitation, you can reopen it when you commit that. Or you can make it a separate PR, whichever is easier. 

We'll discuss the sanitation requirements in Slack, once we have those settled I'll write them up as a new issue.

js/views/modals/editListing/EditListing.js
tenthhuman: Okay

package.json
tenthhuman: Okay

tenthhuman:
**Recommendation:** Closed before merge because requirements not finalized. 
**Status:** File closed.
Okay

PR #222, opened 2016-12-16
Add border to search wrapper.

### LINKED ISSUE #123: Will the address bar not have any border?
I find it really hard to figure out where to grab the window when I want to drag it because there's no border on the address bar.

image

jjeffryes:
- Adds a border to the search wrapper around the search in the page nav, so it's more obvious which areas can be dragged to move the window.
- Closes #123 

@morebrownies I'm not sure I like the border there, it kind of clutters up the design, but it does solve the issue of giving an indicator of where you can drag the window. If you want to modify this, please do so.


hoffmabc:
Oh god this is so much better!

morebrownies:
>  I'm not sure I like the border there, it kind of clutters up the design,

+1, looks really bad

> but it does solve the issue of giving an indicator of where you can drag the window. If you want to modify this, please do so.

There may be a better way achieve this, but I'm not sure if it's possible / easy to do...

If I **click and hold** anywhere within pageNavContainer (or specifically within the address bar input), it's probably safe to assume I'm attempting to drag and change the placement of the window. This could help us expand the target area of where you can click to move the window. 

We would need to figure out if we could accurately detect click and hold events though. It seems https://api.jquery.com/mousedown/ might help us achieve this, but we'd have to also time how long the mousedown event is running for. Maybe it could work?

morebrownies:
This may get complicated though since the user may be attempting to highlight text in the input. hmmm

morebrownies:
Another option may be to only show the address bar border on the hover and active state.

jjeffryes:
I think it would be possible to show the border when hovering and active. I'll take a look at it.

hoffmabc:
Usability overrides looks here. It's unusable without a border.

hoffmabc:
Anyone who uses it for more than one minute will feel the pain of having to click around to find a grabbable chrome. Spotify is light years better here in its current state.

jjeffryes:
@morebrownies I changed it so the border only appears on hover.

morebrownies:
@jjeffryes can we get the border to display when the addressBar is focused too?

jjeffryes:
@morebrownies updated to show the border on focus too. Slightly tricky to do.

jjeffryes:
@morebrownies this is just waiting for your check off to merge.

morebrownies:
@jjeffryes I noticed the magnifying glass (search icon) no longer displays in the address bar. Was that intentional? Or a bug? 

screenshot 2017-01-09 12 14 05

jjeffryes:
@morebrownies  Not intentional, good catch. I'll take a look and see why that happened.

jjeffryes:
@morebrownies magnifying glass restored.

morebrownies:
Looks good to me 👍

PR #224, opened 2016-12-16
Select all text in address bar on click

### LINKED ISSUE #223: When clicking address bar auto-select all text inside of it (i.e. Chrome)
Self-explanatory.

hoffmabc:
Fixes #223

### DIFF
--- a/js/views/PageNav.js
+++ b/js/views/PageNav.js
@@ -20,6 +20,7 @@ export default class extends View {
         'click .js-navMin': 'navMinClick',
         'click .js-navMax': 'navMaxClick',
         'keyup .js-addressBar': 'onKeyupAddressBar',
+        'focusin .js-addressBar': 'onFocusInAddressBar',
         'click .js-navListBtn': 'navListBtnClick',
         'click .js-navSettings': 'navSettingsClick',
         'click .js-navCreateListing': 'navCreateListingClick',
@@ -140,6 +141,10 @@ export default class extends View {
     if (!$(e.target).closest('.js-navListBtn, .js-navNotifBtn').length) {
       this.togglePopMenu();
     }
+  }
+
+  onFocusInAddressBar() {
+    this.$addressBar.select();
   }
 
   onKeyupAddressBar(e) {


morebrownies:
this.$addressBar.select() should only fire on the initial click. Any subsequent click should not highlight the entire address each time.

hoffmabc:
Yes I know, I'm working on it.

hoffmabc:
How's this @morebrownies ?

morebrownies:
It seems a bit buggy for me @hoffmabc. The initial click works well (it highlights everything), but every 2nd or 3rd click (it's not very consistent) seems to also highlight the entire address. 

We have this working in 1.0. Might be worth checking what we do there.

hoffmabc:
Are you sure you're trying the latest code? It should only select the whole address if nothing has been selected yet.

morebrownies:
Yeah I'm on the latest

http://tinypic.com/r/1zfn8yv/9

js/views/PageNav.js
```diff
@@ -20,6 +20,7 @@ export default class extends View {
         'click .js-navMin': 'navMinClick',
         'click .js-navMax': 'navMaxClick',
         'keyup .js-addressBar': 'onKeyupAddressBar',
+        'mouseup .js-addressBar': 'onMouseUpAddressBar',
```
rmisio: I think it's much simpler to just use the `focusin` event instead of the mouseup, then you don't have to do all this checking to see if it's selected.

rmisio: Also, can you please lint the code? `npm run lint`

hoffmabc:
Yeah I can lint it but the question is whether it's actually doing the proper action. Mike seems to see that it's not. I'll take a look.

rmisio:
@Yeah I can lint it but the question is whether it's actually doing the proper action. Mike seems to see that it's not. I'll take a look.

Please see my first comment. If you switch to the `focusin` event, it should work as Mike expects and 75% of the code could be eliminated.

hoffmabc:
https://media.tenor.co/images/457a8a9210ce4018408a15feea0189b0/raw

rmisio:
at ease, soldier.

hoffmabc:
Ok @rmisio I cleaned it up.

PR #229, opened 2016-12-17
Upgrade editor closes #95 progress one

### LINKED ISSUE #95: Replace the Medium Editor with a Better One

tenthhuman:
**Complete** 👍 👍 🌮 

- **Upgrade:** A new editor ( [trumbowyg ](https://alex-d.github.io/Trumbowyg/)) is installed in 1 location: Listing Item Description.
- **Hygeine:** The install attempts to prevent XSS vectors and maintain data hygeine by conforming to our set of allowed tags, attributes and schemes.
- **Affordance**: The tags are filtered when the data is saved and when it is rendered. 

**Excepting** 👻 

- **Eager ghost:** Ghost has discovered a possible opportunity for increased clarity in requirements relating to HTML sanitization. 

- **XSS**: XSS vulnerability existed when using textarea because `TEXTAREA` [converts entities to characters](http://stackoverflow.com/a/8519356/7003753) which allowed attack vectors to hide from sanitization by using suitable escapes. For this reason changes were made to replace textarea with `DIV[contenteditable]` where editor operates. See the commit prior to this pull request ( fbea701e27bed32cc8eb6c4b38ee8375b2b515e4 ) for more details.


lint.list
```diff
@@ -0,0 +1,62 @@
+
```
tenthhuman: Please note this temporary file is removed in https://github.com/OpenBazaar/openbazaar-desktop/pull/229/commits/14cabeaba589f34b57f1455637545eb95008b83b

js/data/security/htmlFilter.js
```diff
@@ -0,0 +1,19 @@
+export const htmlFilter = {
```
rmisio: Since we haven't decided what (if any) type of sanitization we want to do here and since we're leaning towards none being necessary, can we remove any sanitization related code? I'd rather not put code into the code-base that -might- be necessary.

js/utils/trumbowyg.js
```diff
@@ -0,0 +1,29 @@
...
+    ['formatting'],
+    ['bold', 'italic'],
+    ['link'],
+    ['insertImage'],
+    'btnGrp-justify',
+    'btnGrp-lists',
+  ],
+};
+
+export function installRichEditor(attachPoint) {
```
rmisio: Can we please pass in the options as an argument and default it to the constant declared above? It would be nice to give the user the option of providing different options.

Something like:

```
export function installRichEditor(attachPoint, editorOptions = editorOptions) {
```

rmisio: I'm not sure if this needs to necessarily be changed, but flagging it as a concern. When Backbone View's call this method, they should really only pass in the attachPoint as an element (jQuery or otherwise) that was scoped to their view.

The risk with passing it in as a selector is a.) The whole DOM is needlessly being searched rather than a subset of it (i.e. the el of your view) b.) related to A, there is a risk of finding an element a different view, for example, if you passed in a somewhat simple selector of `.description`, there may very well be other elements in the dom matching that.

So... we can either leave it as it is and leave it up to the dev using it (and the reviewers reviewing it) to be disciplined or we could also integrate this specific function into the baseVw and have the selector be scoped to the view `const attach = this.$(attachPoint);`. I tend to lean as the latter option being safer.

js/views/modals/editListing/EditListing.js
```diff
@@ -445,6 +447,9 @@ export default class extends BaseModal {
 
     this.$saveButton.addClass('disabled');
 
+    // one reason this is here since slots are not modified within validate
+    formData.listing.item.description = sanitizeHtml(formData.listing.item.description, htmlFilter);
```
rmisio: Let's remove this until we really need it. As of now, we're not allowing code-mode, so the user really won't be able to add any "bad" html.

tenthhuman: Okay

js/utils/trumbowyg.js
rmisio: FWIW, after letting this sink in for a day, if you haven't made this change already... if you left this is as is and just changed the following line in EditListing.js, I think that would fine:

```
installRichEditor('#editListingDescription');
```

to:

```
installRichEditor(this.$('#editListingDescription'));
```

tenthhuman:
Okay

Kris

js/views/baseVw.js
```diff
@@ -37,7 +37,11 @@ export default class baseVw extends View {
    */
   getFormData(selector) {
     const $formFields = selector instanceof $ ?
-      selector : this.$(selector || 'select[name], input[name], textarea[name]');
+      selector : this.$(selector ||
+        `select[name], input[name], 
+        textarea[name]:not([data-do-not-collect]),
```
rmisio: Instead of having to add in the `data-do-not-collect` attribute, couldn't the selector just be `textarea[name]:not(.trumbowyg-textarea)`?

tenthhuman:
Okay

Kris

js/views/baseVw.js
tenthhuman: Okay

js/utils/trumbowyg.js
tenthhuman: Okay

rmisio:
Couple minor style issues:

1.) Can you please remove references to `.medium-editor-element` from the scss files?
2.) Please ensure the trumbo field is integrated into the styles in _form.scss. It looks like it's as simple as:

- add `.trumbowyg div[contenteditable]` at line 22. Among other things, that will ensure our Trumbo field gets the same font as the other fields.
- add `.trumbowyg div[contenteditable]:focus` at line 49. Among other things, that will ensure our Trumbo field gets a border when it's focused.

Other than that, this LGTM. Thanks!

tenthhuman:
Okay

package.json
```diff
@@ -52,6 +52,8 @@
     "node-sass": "^3.13.0",
     "nodemon": "^1.9.2",
     "npm-run-all": "^2.2.0",
+    "sanitize-html": "^1.13.0",
```
jjeffryes: Is sanitize still being used in this PR?

tenthhuman:
Good find. Okay

styles/components/_form.scss
```diff
@@ -55,8 +63,8 @@ textarea,
   line-height: 40px;
 }
 
-.medium-editor-element {
-  padding-top: $padSm;
+.trumbowyg div[contenteditable].rows6 {
```
jjeffryes: If the rows6 class is only used for the trumbowyg div, it would be better to put this on .trumbowyg-editor instead, and remove the rows6 class entirely, rather than adding a new class.

If the height should only be applied to the editor in the edit listing view, not all instances of the editor, then this rule should be in the editListing.scss file. 

It looks like trumbowyg sets its own height in the style attribute, so it would be best to just remove this rule, since it is overridden by trumbowyg.

styles/components/_form.scss
```diff
@@ -55,8 +63,8 @@ textarea,
...
+  height: 18ex;
```
jjeffryes: I don't recommend using the x-height for this, as some fonts may have a total height larger than twice the x-height, which will throw things off. Using em or rem as the unit would be better.

tenthhuman: Okay. Good info.

styles/components/_form.scss
tenthhuman: Okay.

jjeffryes:
@rmisio this looks good to me, when you close your review you can merge this.

tenthhuman:
🆗 ❗️ ❗️ 
🎉 🎈 👍 👍 🌮 👻 🌭 🍓 
Good merge. 
Thank you @jjeffryes and @rmisio for all the useful and patient reviews. 
Onward! :D ;p :) xx

PR #230, opened 2016-12-19
Fixes issue 128 & 199

### LINKED ISSUE #199: Home: Font color
screen shot 2016-12-13 at 9 50 54 am

Need a similar tweak that we have for OBv1 so that the text is easily visible irrespective of the image color.

### LINKED ISSUE #218: Wrapping on User Menu looks odd
Are we really doing it this way for wrapping?

image

morebrownies:
**Changes**
- Wrapping on User Menu looks odd (fixes #218)
- I also removed the margin to the right of the User Menu so that it aligns with the user's avatar
screenshot 2016-12-19 12 42 23

- Home: Font color (fixes #199) 

I used an outline here in addition to a shadow. It makes the text stand out much better. Also I moved this out of theme.scss since these colors can be hardcoded. 

screenshot 2016-12-19 12 44 17

- Fixed padding issues in About > Information

**Before** 
screenshot 2016-12-19 12 46 57

**After**
screenshot 2016-12-19 12 26 52


styles/modules/_pageNav.scss
```diff
@@ -148,8 +148,7 @@
 
   .navListWrapper {
     z-index: 1;
-    padding: 0 $pad $pad $pad; // give room for border and shadow
-    right: $pad; // compensate for padding
+    padding: 0 0 $pad $pad; // give room for border and shadow
```
jjeffryes: This removes the shadow on the right side, which is probably not what you intended. 

Instead, leave the padding as-is, and change the right to negative, which is what it should have been.

styles/modules/_userPage.scss
```diff
@@ -20,11 +20,16 @@
       padding: 268px 0 40px 0;
     }
 
+    .listBox {
+      .listItem {
+        padding: $pad 0;
+      }
+    }
+
```
jjeffryes: This is needed because of other changes made to the listbox, rather than layering on more changes, it would be better to fix the original problem caused by adding too much padding to the container.

I'll put in a patch.

PR #236, opened 2016-12-22
New style modals

rmisio:
Refactoring the Settings and Edit Listing modals to use the new modal style.


js/templates/modals/listingDetail/listing.html
```diff
@@ -1,161 +1,159 @@
 <div class="popInMessageHolder js-popInMessages"></div>
 
-<div class="modalContentInner">
```
rmisio: Oops, I actually made minimal changes to the file. It looks like a lot because I removed the wrapping .modalContentInner element which didn't seem necessary and had a global style in _modals.scss.

js/views/modals/BaseModal.js
```diff
@@ -46,9 +46,39 @@ export default class BaseModal extends baseVw {
...
+   * That should be done by overriding this method. You will very likely want to include
+   * a call to super in the array you are returning, e.g:
+   *
+   *     return [myCustomEl, myCustomEl2, ...super.closeClickTargets];
+   *
+   * Also, the overridden method will be called on EVERY click in this modal, so it is
+   * critical you are caching the elements returned (i.e. DO NOT go to the DOM every time
+   * the method is called).
+   */
+  get closeClickTargets() {
```
rmisio: @tenthhuman Once this code is integrated, you're going to want to override this method in the About modal so clicks on the overlay part of the modal result in the modal being closed. Otherwise, a click on the part under the nav menu wouldn't result in the modal being closed:

image

I did this in the Settings modal, so you could use that as an example. It's virtually the same thing that you'll want to apply to the About modal.

tenthhuman: Okay.
The new modals look good.

jjeffryes:
@rmisio looks like there are some minor merge conflicts. Other than that, it looks good.

PR #237, opened 2016-12-23
[WIP] Wolf style fixes

morebrownies:
**Changes**
- Added shadow to tags
- Reduced padding in tags
- Buy now button is caps now
- Empty state for Terms, Return Policy, Tags display in italics now (this makes it more clear that it's not user entered copy)
- Cleaned up the publishing status bar


js/templates/modals/listingDetail/listing.html
```diff
@@ -2,8 +2,19 @@
 
 <div class="modalContentInner">
   <div class="listingControls flex">
+    <div class="jsModalClose contentBox clrP clrBr clrSh3" style="height: 35px; line-height: 35px; padding: 0 8px; cursor: pointer">
```
jjeffryes: This block should have an if around it, it's not guaranteed that this data will be available (for instance, if the listing address is put directly in the address bar).

js/templates/modals/listingDetail/listing.html
```diff
@@ -2,8 +2,19 @@
...
+      <div class="userIcon floL discTn clrBr2 clrSh1"
+        <% var avatarHash = ob.avatarHashes ? ob.isHiRez() ? ob.avatarHashes.small : ob.avatarHashes.tiny : ''; %>
+        <% if (avatarHash) { %>
+          style="background-image: url(<%= ob.getServerUrl(`ipfs/${avatarHash}`) %>), url('../imgs/defaultAvatar.png')"
+        <% } else { %>
+         style="background-image: url('../imgs/defaultAvatar.png');"
+        <% } %>>
+      </div>
+      <div style="overflow: auto; font-size: 13px">Store Name <u style="font-size: 12px">View store</u></div>
```
jjeffryes: Store name should actually be the store name, shouldn't it? I assume this is a placeholder.

js/templates/modals/listingDetail/listing.html
```diff
@@ -2,8 +2,19 @@
 
 <div class="modalContentInner">
   <div class="listingControls flex">
+    <div class="jsModalClose contentBox clrP clrBr clrSh3" style="height: 35px; line-height: 35px; padding: 0 8px; cursor: pointer">
+      <div class="userIcon floL discTn clrBr2 clrSh1"
```
jjeffryes: The avatar is aligned strangely, the vertical flex class is probably what you want here. 

image

js/templates/modals/listingDetail/listing.html
```diff
@@ -2,8 +2,19 @@
...
+          style="background-image: url(<%= ob.getServerUrl(`ipfs/${avatarHash}`) %>), url('../imgs/defaultAvatar.png')"
+        <% } else { %>
+         style="background-image: url('../imgs/defaultAvatar.png');"
+        <% } %>>
+      </div>
+      <div style="overflow: auto; font-size: 13px">Store Name <u style="font-size: 12px">View store</u></div>
+    </div>
     <% if (ob.ownListing) { %>
-    <div class="btnStrip clrSh3">
+    <div class="btnStrip clrSh3" style="margin-left: auto; margin-right: 40px">
```
jjeffryes: Instead of putting an inline margin here, the listingControls class should have a rule to add space to the right if the close button is present. 

Then these buttons can be aligned to the right either with a flex alignment class, or the previous sibling of the btnStrip class having a flexExpand class.

js/templates/modals/listingDetail/listing.html
```diff
@@ -48,7 +59,7 @@ <h2 class="txUnb"><%= ob.convertAndFormatCurrency(ob.item.price, ob.metadata.pri
                 </div>
               </div>
               -->
-              <button class="btnHg clrBAttGrad clrBr3 clrTOnEmph">Buy Now</button>
+              <button class="btnHg clrBAttGrad clrBr3 txUp clrTOnEmph">Buy Now</button>
```
jjeffryes: I think it would be better to capitalize this in the language file, since we don't know if capitalization would be appropriate for all languages.

morebrownies:
This pr had WIP label on it. It wasn't ready for review.

jjeffryes:
Ah, my fault, didn't see the WIP.

rmisio:
image

js/templates/modals/listingDetail/listing.html
morebrownies: @rmisio @jjeffryes @cpacia do we have the actual store name available here yet?

jjeffryes: No, we only have data from the contract API, which doesn't return the name of the node that owns the listing. 

We'd have to either add it when the listing modal is created (if the modal was opened from the store page), or make a profile call to get it using the guid (the listing does have the guid of the seller). 

Calling the profile would be needed when the listing was opened from a channel or search results, since in that case the client wouldn't already have the profile information for the seller.

(btw, a quick way to find out what data is available in a template is to add <% console.log(ob) %>somewhere, that will list all the functions and data you can access in the console).

rmisio: It's on my radar to wire this in as much as possible.

On the store we have the data. On search (at least the 3rd party search), we'll also have the data. Not sure about channel, but I think we'll have the data there too since the little listing card needs some basic profile data.

The only case I'm seeing where we might not have this data is if you directly enter a channel based listing url (e.g. @themes/channel/big-brown-shoe) in the address bar, so we'd probably just need to lazy load it in there. Depending on how we design the channel, it's possible we may be lazy loading many things related to the underlying channel page, so that if and when the user closes the modal, they have some semblance of a channel waiting for them.

morebrownies:
I'm going to close this PR for now. I cherry picked some of these changes into the Modal style fixes PR and I'm blocked on not having access to the Store name in the API. Once that gets integrated, I can circle back and make this work:

screenshot 2017-01-09 12 18 07

PR #242, opened 2016-12-29
Switches to jQuery Zoom for listing images.

### LINKED ISSUE #233: Photo carousel changes
@jjeffryes @rmisio here's a prototype for the suggested photo carousel changes. I used jQuery Zoom (http://www.jacklmoore.com/zoom/) + 15 lines of js to get working. 

Notice the photos are not zoomed by default, but they can be clicked to zoom, and the cursor can be used to focus on different parts of the photo.  

It feels pretty good to me. Thoughts? 

https://jsfiddle.net/65shu8xq/13/

jjeffryes:
- changes the non-zoomed image to an image object and adds the required styles so the image will be sized correctly and align with the zoomed image when its smaller than the containing parent.
- some clean up in the render
- changed the photoSection ID to a class, so it didn't override the hide class.
- moved trumboqyg in package.json from dev dependencies to regular dependencies
- Closes #233


rmisio:
A few thoughts...

1.) Can we have an hourglass icon to indicate that zooming is possible? Right now we have a hand and I'm not sure that really reflects what's possible.

2.) If you have a particularly small image, weird things happen when you try and zoom / pan. Maybe we should disable the functionality in that case? (or even zooming into a pixelated region would be better than what's happening now)

image

3.) It feels like a smoother experience to me to have it like the fiddle where you click to activate zoom and then any mouse movements pan and then you click again to deactivate zoom and zoom out. Now you have to click to zoom and click-hold-mousemove to pan. It seems like a bit of a pain to have to hold the mouse down as you pan. What was the reasoning for deviating from the fiddle behavior?

rmisio:
Would you also please take a look at the merge conflict?

js/views/modals/listingDetail/Listing.js
```diff
@@ -158,8 +158,37 @@ export default class extends BaseModal {
...
+                this.$photoSelected.addClass('open');
+              },
+              onZoomOut: () => {
+                this.$photoSelected.removeClass('open');
+              },
+            });
+      } else {
+        this.$photoSelected.addClass('unzoomable');
+      }
+    }, 500);
```
jjeffryes: The timeout here is a hack, but I haven't been able to find another way around it. Even though this waits for the image load event to complete, the image hasn't rendered yet, and has a height and width of zero. 

(a timeout of 0 doesn't work, and requestAnimationFrame didn't either)

If there's another way to detect when the image has rendered, and test whether it should be zoomable at that point, I'm all ears. Possibly it would need a loop to check if the image has size yet, which seems like overkill.

rmisio: The height should be available on load:

http://stackoverflow.com/questions/14134393/get-image-height-after-image-has-loaded

If it's not, the first things I would check are a.) is it in the DOM b.) is it visible (not display: none).

jjeffryes: It's definitely not available on load. That surprised me too.

rmisio: It is available on load:

https://jsfiddle.net/p3z68ccv/2/

If it's not for you, it's because something else is happening. Have you checked the things I mentioned above to check?

js/views/modals/listingDetail/Listing.js
```diff
@@ -278,14 +300,24 @@ export default class extends BaseModal {
...
-      (this._$photoSection = this.$('#photoSection'));
+      (this._$photoSection = this.$('.js-photoSection'));
   }
 
   get $photoSelected() {
     return this._$photoSelected ||
       (this._$photoSelected = this.$('.js-photoSelected'));
   }
 
+  get $photoSelectedHeight() {
```
rmisio: Your caching both $photoSelectedHeight() & $photoSelectedWidth(), but you're never clearing `_$photoSelectedHeight` and `._$photoSelectedWidth` on render. So, on a re-render that will be cached to an inaccurate value.

I don't think it necessary to cache these guys since you're already caching the $photoSelected dom element, and beyond getting the element, I don't think determining height and width makes any special trips to the DOM or it's a particularly intensive process.

morebrownies:
Awesome. Taking a look now.

morebrownies:
Looks good to me @jjeffryes @rmisio. I pushed a one line change to make the photo gallery always black.

PR #246, opened 2016-12-29
Fixes external images that are too big.

### LINKED ISSUE #227: Images are not outputting in the listing description
I added an image tag within the listing description field like so:

screenshot 2016-12-16 10 46 45

I click **Save**, and the image is now rending inline within the edit listing modal (cool!!) 

screenshot 2016-12-16 10 45 00

I then check the listing detail screen and I see the raw tag (even after refreshing):

screenshot 2016-12-16 10 49 14

### LINKED ISSUE #245: Saving a Large Image Blows Up the Edit Listing Layout
image

If you use the trumbowyg editor to add an external image that is larger than the tabContent container, it will size itself correctly. But when you click the save button, the image will re-size to it's actual size, and push the width of the tabColum to fit it.

jjeffryes:
- sets the tabColumn in tabbed modals to a fixed max width to prevent content from making it too big.
- sets images in the description area to no larger than 100% width.
- Closes #227
- Closes #245

Also fixes weird bug where Trumbowyg sets its height to thousands of pixels tall when images are used (I think it's setting it's height based on the original height of the images, then shrinking the images to fit in the editor, leaving a huge gap at the bottom).

image

### DIFF
--- a/styles/_base.scss
+++ b/styles/_base.scss
@@ -103,6 +103,12 @@ ul, ol {
   background-repeat: no-repeat;
 }
 
+.trumbowyg .trumbowyg-editor {
+  height: auto !important;
+  /* gross hack to stop trumbowyg from setting inline height to huge numbers when external
+  images are added */
+}
+
 #appFrame {
   position: absolute;
   height: 100%;
--- a/styles/_base.scss
+++ b/styles/_base.scss
@@ -103,12 +103,6 @@ ul, ol {
   background-repeat: no-repeat;
 }
 
-.trumbowyg .trumbowyg-editor {
-  height: auto !important;
-  /* gross hack to stop trumbowyg from setting inline height to huge numbers when external
-  images are added */
-}
-
 #appFrame {
   position: absolute;
   height: 100%;
--- a/styles/components/_tabbedModal.scss
+++ b/styles/components/_tabbedModal.scss
@@ -30,6 +30,10 @@
     }
   }
 
+  .tabContent {
+    max-width: 757px;
+  }
+
   &.fixedNav {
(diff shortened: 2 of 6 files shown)


styles/_base.scss
```diff
@@ -103,6 +103,12 @@ ul, ol {
   background-repeat: no-repeat;
 }
 
+.trumbowyg .trumbowyg-editor {
```
rmisio: Would this be more appropriate inside of components/misc? It seems like a really random thing to be inside of base.

styles/components/_tabbedModal.scss
```diff
@@ -30,6 +30,10 @@
     }
   }
 
+  .tabContent {
```
rmisio: Should we really be putting this max on the global component? I guess the question is if the designs will be that all tabbed modals will have a similar width. If that's the case, then this is fine.

(thinking out loud... it's probably ok for now and if some designs present themselves where this doesn't work, I suppose we could adjust).

styles/_base.scss
jjeffryes: Good idea.

styles/components/_tabbedModal.scss
jjeffryes: There may be a better way to do this, I'm still not 100% sure why the image blows things up on save.

The parent modal has a fixed width, I think that makes it ok for now to have fixed width columns in it. We could also do percentages.

rmisio: Let's leave it as is. I definitely wouldn't do percentages, since as of now, these modals aren't responsive and there could be a good amount of scope to make them so.

PR #247, opened 2016-12-30
Brand About Modal Take Two closes #87

### LINKED ISSUE #87: Create About Modal
image

The About Modal is a modal that should be shown when 
- [x] the About link in the page navigation list is clicked. 
- [x] it is programmatically opened  (we don't have a use case where this would be triggered other than the About link currently, but it should still work).

Currently, there is no plan to have functionality to open the About modal to a specific tab.

When the About Modal is open, it does not change the address shown in the address bar. Like most modals, the About Modal is not recorded in history. 

**The About Modal should show the following elements:**
- [x] the current value of one Bitcoin in the user's fiat currency..
    - [x] if the user has selected BTC as their currency, the current value element is not shown.
- [x] a tab menu with the following tabs:
    - The Story
    - Contributors
    - Donations
    - Licensing
- [x] clicking each tab will show that tab section, and put an active indicator on the tab element.
- [x] The headline of the currently active tab.
- [ ] The version number of the client and server, in the format ("client X.X, server X.X"). The client version number can be retrieved from package.json. The server version should be in the settings data (an issue has been posted for this).
- [ ] A "Check for Update" button. This will activate the update code, and update the app if an update is available (stub this in for now, the functionality is not yet available). 
- [x] The logo of OpenBazaar, along with the text "OpenBazaar" and some icons. This is visual only, and has no functionality.
- [x] A close button. This will close the modal. The Escape key should also close the modal (these are both built-in functions of the base modal view).

**The tabs have the following elements/functionality:** 

**The Story**
- [x] The headline, "The Story of OpenBazaar". 
- [x] The main text in the modal (content TBD). This will be static formatted HTML text. (check with Jenn for this text)

**Contributors**
- [x] The headline, "Contributors". 
- [x] The main text area in the modal will have some text, and a list of the contributors to OpenBazaar.
- [ ] TODO later: when we implement an endpoint on openbazaar.org with dynamic contributor data, this section can be updated to pull from that endpoint instead.

**Donations**
- [x] The headline, "Donations". 
- [x] A QR code for the address to send donations to the OpenBazaar project. This can be supplied by Sam.
- [x] The same address in plain text.
- [x] Clicking on the address will add it to the user's clipboard. If the user clicks on the address and it is added to their clipboard, a message should be shown, similar to when the guid on the about tab of the user page is clicked (the same styles can be used).
- [x] A "Copy Address" button. Clicking on it will add the BitCoin address to the user's clipboard.
- [ ] A "Open in Wallet" button. Clicking on it will open the transaction in the app's internal wallet. (stub this in if the internal wallet hasn't been built. As of 12/12/16, work has not yet begun on the internal wallet).
- [x] A static text area is at the bottom.

**Licensing**
- [x] The headline, "Licensing". 
- [ ] The main text in the modal includes the MIT license, and a list of the libraries being used. 
- [ ] A link to our github repo.
- [ ] A list of the major libraries we use (look at package.json for a basis).

image

image

image

tenthhuman:
**Complete** 🎉 👍 👍 🌮 

- **Brand Story:** A new modal is installed that provides a place to tell the OpenBazaar story. Including Contributors, the software License, and a dashboard for Donations. 
- **Image assets:** Attractive image assets exposed on the team Zeplin boards have all been linked.

**Excepting** 👻 

- **Ghost ceremony:** This PR is explicitly a work in progress. The ghost is happy to be actively involved in discovering opportunities for improvement. 
- **The QR code:** A library will me imported and a QR code created on demand for display in the Donations tab. 
- **The BTC address:** Shall be copied upon click.
- **Various:** Various improvements in style and sundry to be identified by the team.

- **BTC Widget:** A Bitcoin price widget shall be incorporated above the content panel of all tabs of this model.

**Open Questions** 💯 ⛏ 

- **Dynamic donation count:** shall it be dynamic? An API? Parse the [blockchain.info](https://blockchain.info/address/3MXYUBLWNETa5HTewZp1xMTt7AW9kbFNqs) page?
- **Dynamic contributors list:** shall it be dynamic? An API? Parse the GitHub [server](https://github.com/OpenBazaar/openbazaar-go/graphs/contributors) and [client](https://github.com/OpenBazaar/openbazaar-desktop/graphs/contributors) ( and other ? ) pages?
- **Dependencies list:** shall the app nod to its open source giant's shoulders? Shall such a list be dynamic? Parse the `package.json`?

**Relevant issues**

- This PR closes #87 when merged


js/languages/en-US.json
```diff
@@ -87,6 +87,57 @@
     "userLoading": "Loading...",
     "userNotFound": "Information for %{guid} was not found on the network."
   },
+  "about": {
+    "_linkText": "About OpenBazaar",
```
rmisio: I'm still wondering :) ... Is it critical to maitain this '_ ' preface to some translation keys? Even after your explanation, I still don't understand the reasoning. My concern is that other developers / translators (like me) see this convention and have no idea what is going on.

js/languages/pl.json
```diff
@@ -1,410 +0,0 @@
-{
```
rmisio: I think this could stay. A small number of the countries are actually translated and it's useful for testing purposes to have them there.

js/views/PageNav.js
```diff
@@ -199,6 +201,13 @@ export default class extends View {
     this.togglePopMenu();
   }
 
+  navAboutClick() {
+    if (!this.settingsModal || !this.settingsModal.isOpen()) {
```
rmisio: The references to the `settingsModal` need to be changed to `aboutModal`.

js/templates/modals/about/story.html
```diff
@@ -0,0 +1,36 @@
+<div class="contentBox padMd clrP clrBr clrSh3">
+  <div class="flexSpaceBetween">
+    <div class="flexRow flex0">
+      <div class="brandSmall" style="background-image: url('../imgs/baseRounded.svg');"></div>
+      <div>
+        <div><%= ob.polyT('about._brand') %></div>
+        <div class="flexRow">
+          <img src=1>
```
rmisio: Looks like these image sources are off.

js/templates/modals/about/donations.html
```diff
@@ -0,0 +1,51 @@
...
+      <a class="btn clrP clrBr js-copyAddress flexNoShrink">
+        <%= ob.polyT('about.btnUpdateCheck') %>
+      </a>
+    </div>
+  </div>
+  <hr class="clrBr" />
+  <div class="padLg">
+    <div id=donate-cta class="padMd flexBar">
+      <div id=donation-qr-code class="padMd">
+        <img src=qr>
```
rmisio: This image source looks off and results in no image loaded:

image

js/views/modals/About/About.js
```diff
@@ -0,0 +1,100 @@
...
+import Story from './Story';
+import Contributors from './Contributors';
+import Donations from './Donations';
+import License from './License';
+
+export default class extends BaseModal {
+  constructor(options = {}) {
+    const opts = {
+      removeOnClose: true,
+      removeOnRoute: false,
```
rmisio: The `removeOnRoute` option should be removed from here, so the default of 'true' will be enacted by via the baseModal. The options when set to true, ensures the modal will be removed if we route to another page.

I know you copied this code from settings, but the settings modal had some special reason to override the default.

js/templates/modals/about/story.html
```diff
@@ -0,0 +1,36 @@
...
+    <p>
+      <%= ob.polyT('about.storyTab.paragraphs.beginning') %>
+    </p>
+    <p>
+      <%= ob.polyT('about.storyTab.paragraphs.middle') %>
+    </p>
+    <p>
+      <%= ob.polyT('about.storyTab.paragraphs.end') %>
+    </p>
+    <hr class="clrBr" />
```
rmisio: I think this hr could be removed. It's not in the designs.

js/templates/modals/about/contributors.html
```diff
@@ -0,0 +1,33 @@
...
+  <hr class="clrBr" />
+  <div class="padMd">
+    <p>
+      <%= ob.polyT('about._blurbs.tradeFree') %>
+      <%= ob.polyT('about.contributorsTab.listPreface') %>
+    </p>
+    <ul>
+      <% ob.names.forEach(name => print(`<li>${name}</li>`)) %>
+    </ul>
+    <hr class="clrBr" />
```
rmisio: This hr could be removed since it's not in the design.

js/templates/modals/about/license.html
```diff
@@ -0,0 +1,37 @@
...
+    <p>
+      <%= ob.polyT('about.licenseTab.paragraphs.permission') %>
+    </p>
+    <p>
+      <%= ob.polyT('about.licenseTab.paragraphs.inclusion') %>
+    </p>
+    <p>
+      <%= ob.polyT('about.licenseTab.paragraphs.indemnity') %>
+    </p>
+    <hr class="clrBr" />
```
rmisio: this hr can also be removed.

js/templates/modals/about/license.html
```diff
@@ -0,0 +1,37 @@
...
+          <img src=1>
+          <img src=2>
+          <img src=3>
+        </div>
+      </div>
+    </div>
+    <h2 class="h3 clrT"><%= ob.polyT('about.licenseTab._sectionHeader') %></h2>
+    <div class="txtRight">
+      <div><%= ob.polyT('about.version') %> <%= ob.polyT('about.versionCode') %></div>
+      <a class="btn clrP clrBr js-copyAddress flexNoShrink">
```
rmisio: I think the `js-copyAddress` is a copy and paste error since semantically it doesn't make sense here.

js/templates/modals/about/story.html
```diff
@@ -0,0 +1,36 @@
...
+          <img src=1>
+          <img src=2>
+          <img src=3>
+        </div>
+      </div>
+    </div>
+    <h2 class="h3 clrT"><%= ob.polyT('about.storyTab._sectionHeader') %></h2>
+    <div class="txtRight">
+      <div><%= ob.polyT('about.version') %> <%= ob.polyT('about.versionCode') %></div>
+      <a class="btn clrP clrBr js-copyAddress flexNoShrink">
```
rmisio: Right now, this button isn't doing anything. Can we wire in that functionality. You could see a similar example of it on the User Page > Home tab when you click the GUID in the Information box:

image

styles/modules/modals/_about.scss
```diff
@@ -0,0 +1,30 @@
+/*
```
rmisio: Since this is all commented out, can we remove this file until there comes a point we need it (which may be now if you bring over any module specific styles from _layout.scss)?

styles/components/_layout.scss
```diff
@@ -425,3 +425,46 @@
...
+
+  .flex0 {
+    flex: 0;
+  }
+
+  .txtRight {
+    text-align: right;
+  }
+
+  .txtCent {
```
rmisio: We already have a txCtr class in _type.scss. Other generic type specific classes, like txtRight should also go there.

styles/components/_layout.scss
```diff
@@ -425,3 +425,46 @@
...
+    flex-direction: row;
+  }
+
+  .flexSpaceBetween {
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+  }
+  
+  #donation-qr-code {
```
rmisio: This is a generic module and stylings for a specific modal, like #donate-cta and #donate-qr-code should go into a seperate style sheet for that modal (like you were doing with _about.scss).

It's basically about determining what are generic styles, potentially applicable to many different parts of the app (which would go into some of our generic style files (mostly in the components folder) -versus- what styles are specific to a particular module (which would go into a file in the module folder).

js/views/modals/About/About.js
```diff
@@ -0,0 +1,100 @@
...
+
+    this.tabViewCache = {};
+    this.tabViews = {
+      Story,
+      Contributors,
+      Donations,
+      License,
+    };
+
+    this.listenTo(app.router, 'will-route', () => {
```
rmisio: This block could also be removed. Settings has some specific reason to manage this.. but for the About modal, we could just let the BaseModal take care of this (and as I mentioned, that's as simple as removing the `removeOnRoute: true` from our options hash above).

js/views/modals/About/About.js
```diff
@@ -0,0 +1,100 @@
...
+
+    if (!this.currentTabView || this.currentTabView !== tabView) {
+      this.$('.js-tab').removeClass('clrT active');
+      targ.addClass('clrT active');
+      if (this.currentTabView) this.currentTabView.$el.detach();
+
+      if (!tabView) {
+        tabView = this.createChild(this.tabViews[tabViewName]);
+        this.tabViewCache[tabViewName] = tabView;
+        this.listenTo(tabView, 'saving', (...args) => { this.onTabSaving(tabView, ...args); });
```
rmisio: This line can be removed. It's specific to the settings modal where tabs were communicating their 'saving' state to the parent. It doesn't apply here.

package.json
```diff
@@ -49,6 +49,7 @@
     "eslint-plugin-jsx-a11y": "^1.5.3",
     "eslint-plugin-react": "^5.2.2",
     "jsdom": "9.4.1",
+    "medium-editor": "^5.22.1",
```
rmisio: This isn't needed anymore. Some dude coded up a PR where we are now using the trumbo WYSIWYG editor ;)

styles/components/_layout.scss
```diff
@@ -425,3 +425,46 @@
 .posF {
   position: fixed;
 }
+.about { 
+  .brandSmall {
+    width: 50px;
+    hegiht: 50px;
```
rmisio: typo alert!

imgs/baseRounded.svg
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
rmisio: Could we not call this file something like 'logoSmall'? The current name gives me no indication as to what this image actually is.

rmisio:
I know this is WIP, but before we're done, can you be sure to fix the lint issues?

js/languages/en-US.json

js/languages/pl.json
tenthhuman: Okay

js/views/PageNav.js
tenthhuman: Okay

js/views/modals/About/About.js
tenthhuman: Okay.

I guess the special reason it exists in Settings is so that modal stays open when something like #connected-peers is routed to.

js/views/modals/About/About.js
tenthhuman: Okay

styles/modules/modals/_about.scss
tenthhuman: There are some specific styles in there it could work to use in an `_about.scss`

I do not know why the styles did not load. Any ideas?

styles/components/_layout.scss
tenthhuman: Okay

js/views/modals/About/About.js
tenthhuman: Okay

js/templates/modals/about/story.html
tenthhuman: Okay

js/templates/modals/about/contributors.html
tenthhuman: Okay

js/templates/modals/about/license.html
tenthhuman: Okay

package.json
tenthhuman: Okay

styles/components/_layout.scss
tenthhuman: Okay

js/templates/modals/about/license.html
```diff
@@ -0,0 +1,36 @@
+<div class="contentBox padMd clrP clrBr clrSh3">
+  <div class="flexSpaceBetween">
```
jjeffryes: See earlier comment on restructuring this, and doing it once instead of repeating it.

styles/components/_layout.scss
```diff
@@ -425,3 +425,50 @@
...
+    background: black;
+  } 
+
+  #donate-cta {
+    max-width: 80%;
+    margin: auto;
+  }
+
+  .flex0 {
+    flex: 0;
```
jjeffryes: See comments above for a better way to get the layout this class was used for.

As a general rule, we probably don't ever want an element to have flex-basis of 0%, which is one of the things this shorthand dose.

styles/modules/modals/_about.scss
tenthhuman: Okay. The import in `modules/_all.scss` was missing.

styles/components/_layout.scss
tenthhuman: Okay

js/languages/en-US.json
```diff
@@ -87,6 +87,57 @@
...
+    },
+    "version": "version",
+    "versionCode": "2.0",
+    "client": "Client",
+    "server": "Server",
+    "btnUpdateCheck": "Check for Update",
+    "menu": {
+      "sectionHeader": "MENU",
+      "btnNext": "Next",
+      "btnPrevious": "Previous"
```
tenthhuman: Okay.

This was related to our discussion about the menu not scrolling as the content pane scrolled. IIRC it was @rmisio who first considered what would happen if the menu was so long it would also need to scroll. Thinking about that I considered a menu that could be paged forward and back using Next and Previous buttons and not scrolled.

styles/components/_layout.scss
```diff
@@ -425,3 +425,50 @@
 .posF {
   position: fixed;
 }
+.about {
```
tenthhuman: Okay.

styles/components/_layout.scss
```diff
@@ -425,3 +425,50 @@
...
+  }
+
+  .gemSmall {
+    width: 16px;
+    height: 16px;
+    margin: 1px;
+    background-size: contain;
+  }
+
+  .flexBar {
```
tenthhuman: Okay.

js/templates/modals/about/contributors.html
```diff
@@ -0,0 +1,32 @@
+<div class="contentBox padMd clrP clrBr clrSh3">
```
tenthhuman: Okay. 
Sub-template seems most faithful to the structure of new style modals.

js/templates/modals/about/contributors.html
```diff
@@ -0,0 +1,32 @@
...
+      </a>
+    </div>
+  </div>
+  <hr class="clrBr" />
+  <div class="padMd">
+    <p>
+      <%= ob.polyT('about.blurbs.tradeFree') %>
+      <%= ob.polyT('about.contributorsTab.listPreface') %>
+    </p>
+    <ul>
```
tenthhuman: Okay.

js/templates/modals/about/contributors.html
tenthhuman: Okay.

js/templates/modals/about/story.html
```diff
@@ -0,0 +1,35 @@
+<div class="contentBox padMd clrP clrBr clrSh3">
+  <div class="flexSpaceBetween">
```
tenthhuman: Okay.

js/templates/modals/about/contributors.html
```diff
@@ -0,0 +1,32 @@
+<div class="contentBox padMd clrP clrBr clrSh3">
+  <div class="flexSpaceBetween">
+    <div class="flexRow flex0">
+      <div class="brandSmall" style="background-image: url('../imgs/obVectorIcon.svg');"></div>
+      <div>
+        <h4><%= ob.polyT('about.brand') %></h4>
+        <div class="flexRow">
```
tenthhuman: Okay.

js/templates/modals/about/donations.html
```diff
@@ -0,0 +1,51 @@
+<div class="contentBox padMd clrP clrBr clrSh3">
+  <div class="flexSpaceBetween">
```
tenthhuman: Okay.

js/templates/modals/about/license.html
tenthhuman: Okay.

js/templates/modals/about/contributors.html
```diff
@@ -0,0 +1,32 @@
+<div class="contentBox padMd clrP clrBr clrSh3">
+  <div class="flexSpaceBetween">
```
tenthhuman: Okay

js/templates/modals/about/donations.html
```diff
@@ -0,0 +1,51 @@
...
+    <h2 class="h3 clrT"><%= ob.polyT('about.donationsTab.sectionHeader') %></h2>
+    <div class="txtRight">
+      <div><%= ob.polyT('about.version') %> <%= ob.polyT('about.versionCode') %></div>
+      <a class="btn clrP clrBr js-copyAddress flexNoShrink">
+        <%= ob.polyT('about.btnUpdateCheck') %>
+      </a>
+    </div>
+  </div>
+  <hr class="clrBr" />
+  <div class="padLg">
```
tenthhuman: Okay

js/views/modals/About/About.js
```diff
@@ -0,0 +1,98 @@
...
+        tabView = this.createChild(this.tabViews[tabViewName]);
+        this.tabViewCache[tabViewName] = tabView;
+        tabView.render();
+      }
+
+      this.$tabContent.append(tabView.$el);
+      this.currentTabView = tabView;
+      this.currentTabName = tabViewName;
+      this.$('#tabTitle').text(app.polyglot.t(
+      `about.${this.currentTabName.toLowerCase()}Tab.sectionHeader`));
```
rmisio: Can we please indent the second line here so it's clear it's an extension of the first one?

js/templates/modals/about/about.html
```diff
@@ -0,0 +1,41 @@
...
+              <h4><%= ob.polyT('about.brand') %></h4>
+              <div class="flex">
+                <div class="gemSmall" style="background-image: url('../imgs/btcIcon128.png');"></div>
+                <div class="gemSmall" style="background-image: url('../imgs/globeIcon64.png');"></div>
+                <div class="gemSmall" style="background-image: url('../imgs/peaceIcon64.png');"></div>
+              </div>
+            </div>
+          </div>
+          <h2 id=tabTitle class="h3 clrT txCtr flexExpand"></h2>
+          <div class="txtRight">
```
jjeffryes: What's the txtRight class? I don't see it in the style files.

styles/modules/modals/_about.scss
```diff
@@ -0,0 +1,34 @@
...
+    background-size: contain;
+  }
+
+  #donation-qr-code {
+    width: 200px; 
+    height: 200px;
+    background: black;
+  } 
+
+  .columns {
```
jjeffryes: Creating column classes is a good idea. I think this would get used in other places, can you move it to _type.scss?

I think they should go in _type rather than _layout because the column properties applies to content.

Since there is no column width set in the column class, it will be one column, which isn't something we'd ever want. Can you set the column-count to 2 there, as the default? Then the twoColumn class isn't needed, and if we want to have other counts (like 3 or 4) we can create classes for those.

The li styles, however, are specific to the about modal, so they could stay here. I would give the container an about specific class for that, like contributorList.

js/views/modals/About/About.js
tenthhuman: Okay.

styles/modules/modals/_about.scss
tenthhuman: Okay.

js/views/modals/About/Donations.js
```diff
@@ -9,9 +14,37 @@ export default class extends baseVw {
...
+    };
+  }
+
+  copyDonationAddress() {
+    const address = document.querySelector('#donationAddress');
+    const range = document.createRange();
+    const selection = window.getSelection();
+    range.selectNodeContents(address);
+    selection.removeAllRanges();
+    selection.addRange(range);
```
jjeffryes: Forcing any selection to select the whole address is probably a little bit of overkill, but it's ok.

js/templates/modals/about/donations.html
```diff
@@ -3,13 +3,15 @@
     <img src=qr>
   </div>
   <div id=donation-dashboard class="padMd flexCol">
-    <h2><%= ob.polyT('about.donationsTab.bannerCountDonations') %></h2>
+    <h2><%= ob.polyT('about.donationsTab.bannerCountDonations', { donationCountFloor: ob.donationCountFloor }) %></h2>
     <div class="padMd">
```
jjeffryes: The right side of the h2 should line up with the right side of the address and buttons. I'd take the padMd off of the address and buttons, and use a gutter class on the flexCol parent to handle the vertical space.

styles/modules/modals/_about.scss
```diff
@@ -17,18 +17,26 @@
...
 
-  .columns {
-    column-gap: $padMd;
-    max-height: fit-content;
+  .contributorList li {
+    padding-left: $padLg;
+    padding-bottom: $padSm;
+  }
+
+  #donationAddress {
```
jjeffryes: This should be done by putting a clrT2 class on the span. We set all colors using classes on the elements, so they can be changed by changing the color classes (that's going to be important later when we add theming).

styles/modules/modals/_about.scss
```diff
@@ -0,0 +1,42 @@
...
+    color: $text2;
+  }
+
+  #copyNotification {
+    @extend .clrBr, .border;
```
rmisio: Same thing with the clrBr class. It should be put on the span.

Basically, any of the classes in _theme.scss need to be applied directly in the markup. That way when we override those classes when a new theme is applied, the elements will take on the overridden color.

Doing it in the style sheet like this means that, at compile time, SASS will just compile out the default clrBr color and embed it in the #copyNotification selector and it would ignore the fact that we may override that class at some point at run-time.

js/views/modals/About/Donations.js
tenthhuman: Okay.

js/templates/modals/about/donations.html
tenthhuman: Okay

js/templates/modals/about/donations.html
```diff
@@ -0,0 +1,28 @@
...
+  <div id=donation-dashboard class="padMd flexCol">
+    <h2><%= ob.polyT('about.donationsTab.bannerCountDonations', { donationCountFloor: ob.donationCountFloor }) %></h2>
+    <div class="padMd">
+      <span id=donationAddress>
+        <%= ob.obDonationAddress %> 
+      </span>
+      <span id=copyNotification>Copied</span>
+    </div>
+    <div class="padMd">
+      <a id=copyAddress class="btn clrP clrBr js-copyAddress flexNoShrink">
```
tenthhuman: Okay

styles/modules/modals/_about.scss
tenthhuman: Okay.

styles/modules/modals/_about.scss
tenthhuman: Okay

js/views/modals/About/Donations.js
```diff
@@ -16,34 +18,27 @@ export default class extends baseVw {
...
-    const range = document.createRange();
-    const selection = window.getSelection();
-    range.selectNodeContents(address);
-    selection.removeAllRanges();
-    selection.addRange(range);
-
     clipboard.writeText(obDonationAddress);
 
-    this.$('#copyNotification').addClass('active');
+    this.$('.copyNotification').addClass('active');
```
rmisio: There's a certain convention we follow when targeting elements via a class from JS: we preface those classes with `js-`. The reasoning is that it's clear from looking at the markup that a particular element is being managed via JS. If we just used a non 'js' prefaced class, then if some css changed where that class wasn't needed, removing the class would break the JS link.

Similarly, if we're using a class for styling purposes, we don't preface it with 'js'.

So, if an element has both CSS and JS applied to it, it would have both a js and non-js class (e.g. `js-copyNotification` & `copyNotification`).

Sorry for the confusion on this.

js/views/modals/About/Donations.js
```diff
@@ -16,34 +18,27 @@ export default class extends baseVw {
...
     if (!!hiderTimer) {
       clearTimeout(hiderTimer);
     }
     hiderTimer = setTimeout(
-      () => this.$('#copyNotification').removeClass('active'), 3000);
+      () => this.$('.copyNotification').removeClass('active'), 3000);
```
rmisio: Can we please cache the `.copyNotification` element, rather than querying the DOM multiple times for it?

Since it's only being used in this method, it could be as simple as:

```
const $copyNotification = this.$('.copyNotification').addClass('active');
```

then in the timeout:

```
() => $copyNotification.removeClass('active'), 3000);
```

If you think that element will be used in other parts of the view and you want to get real fancy, you could take this approach:

in render - 

```
this._$copyNotification = null; // prefaced with an underscore
```

separate getter function -

```
  get $copyNotification() {
    return this._$copyNotification ||
      (this._$copyNotification = this.$('.copyNotification'));
  }
```

This approach has the benefit in that it will only query the DOM when and if the element is needed and it would make it available view wide. It's an approach we take in most of our other views. Of course, it only really makes sense if an element will conditionally be necessary depending on the code path a user takes. If the element will always be necessary and potentially be necessary more than once, then we could just cache in render (e.g. `$myEl = $('.js-myEl')`).

js/views/modals/About/Donations.js
tenthhuman: Okay.
@rmisio thanks for the very useful explanation. You are always so generous with your time. I really appreciate that.

js/views/modals/About/Donations.js
tenthhuman: Okay.
That is also a great write up and explainer. Thanks a lot @rmisio!

js/views/modals/About/Contributors.js
```diff
@@ -1,35 +1,302 @@
...
+
+  const obGitHubName = 'OpenBazaar';
+  const base = 'https://api.github.com';
+
+  // this is from the GitHub entity created
+  // for purposes of testing this code: theob1ioaccount000
+  // it has no access to anything except public information
+  // because of the limit of 5K calls
+  // there can only be around 600 ( 25 * 24 ) such calls to the
+  // contributor_names process per day ( each call makes around 200 requests )
```
jjeffryes: This is neat, but we probably want to just maintain a static list. The list won't change very often, and it seems fragile to count on a special github account to access it. And we should hope we'll be successful enough to exceed the call limit.

(I see you have made comments along those lines later on, and this was just for testing)

js/views/modals/About/Contributors.js
```diff
@@ -1,35 +1,302 @@
...
-  'LeBron James',
-  'Zach Galifianakis',
-  'LeBron James',
+let names;
+let fetchNames;
+
+// Linter notes:
+  // linter will not be happy with some things which also work:
+  // no-use-before-define -- because of const variable hoisting and invocation time
+  // this works and is a way to share data in closures. The linter is also unhappy.
```
jjeffryes: If we're doing things to get around the linter, there's probably a problem with what we're doing.

js/views/modals/About/Contributors.js
```diff
@@ -1,35 +1,302 @@
...
+      {
+        api: 'name',
+        filter: (job, item) => item.name || job.data.user,
+        isList: false,
+      },
+    ],
+  };
+
+  // Sooth linter by "using" 'regeneratorRuntime'
+  console.info(regeneratorRuntime);
```
jjeffryes: I think there's a better way around the linter issue than this.

js/views/modals/About/Contributors.js
```diff
@@ -1,35 +1,302 @@
...
+
+  function updateNameList(newNames) {
+    const everPresentNames = [
+      'Jenn Cloud',
+    ];
+    names = [...new Set([...everPresentNames, ...newNames])];
+    console.info(`Name list updated ${names}`);
+  }
+
+  // In Depth -- Task Processor
```
jjeffryes: Isn't this serious overkill for what we need to do here?

js/views/modals/About/Donations.js
```diff
@@ -0,0 +1,50 @@
+import loadTemplate from '../../../utils/loadTemplate';
+import baseVw from '../../baseVw';
+import { clipboard } from 'electron';
+import QR from 'qr-encode';
+
+const qr = QR;
```
rmisio: Why make this assignment instead of just using the `QR` form the import? Essentially, what you are doing is not much different from:

```
const QR = 'hello';
const qr = QR;
```

js/views/modals/About/Contributors.js
tenthhuman: Okay

js/views/modals/About/Contributors.js
tenthhuman: Okay

js/views/modals/About/Contributors.js
tenthhuman: Okay

js/views/modals/About/Contributors.js
tenthhuman: Okay

js/views/modals/About/Donations.js
tenthhuman: I am fine with doing that. The linter is not fine.

tenthhuman: Okay

js/views/modals/About/About.js
```diff
@@ -0,0 +1,102 @@
...
+
+      this.$tabContent.append(tabView.$el);
+      this.currentTabView = tabView;
+      this.currentTabName = tabViewName;
+      this.$('#tabTitle').text(app.polyglot.t(
+        `about.${this.currentTabName.toLowerCase()}Tab.sectionHeader`));
+    }
+  }
+
+  get $closeClickTargets() {
```
rmisio: The closeClickTarget code could come about. A UX decision was made to not allow the modals to be closed on an overlay click because accidental clicks were causing the modal to close.

Sorry about the switch-aroo on that one.

js/views/modals/About/Contributors.js
```diff
@@ -0,0 +1,151 @@
...
+  'tenthhuman',
+  'David Albrecht',
+  'Yuri Zhykin',
+  'Leo Arias',
+  'Jack Kleeman',
+  'Alexander Harkness',
+  'Duo Search',
+  'Michael Lynch',
+  'Oyebanji Jacob Mayowa',
+  'Check your git settings!', // it may be cool to list "names" as <name> (<login>)
```
rmisio: Should this guy come out?

js/views/modals/About/Contributors.js
```diff
@@ -0,0 +1,151 @@
+import loadTemplate from '../../../utils/loadTemplate';
+import baseVw from '../../baseVw';
+
+// note: this list was generated automatically
```
rmisio: Is this comment no longer accurate since we're manually maintaining a list?

package.json
```diff
@@ -34,6 +34,7 @@
   "devDependencies": {
     "babel-cli": "^6.10.1",
     "babel-eslint": "^6.0.5",
+    "babel-plugin-transform-regenerator": "^6.21.0",
```
rmisio: Since we've removed the code that used the generator and we're not using any at this time, can we remove this for now? We could always bring it in should we want to implement a generator at some point.

js/views/modals/About/About.js
tenthhuman: Okay

js/views/modals/About/Contributors.js
tenthhuman: Okay

js/views/modals/About/Contributors.js
tenthhuman: Okay

package.json
tenthhuman: Okay

tenthhuman: Removing causes build to fail. Appears to be dependency of `babel-preset-es2015`. Keeping it in.

rmisio: I don't see how that's possible seeing how it's not in master.
https://github.com/OpenBazaar/openbazaar-desktop/blob/master/package.json

js/templates/modals/about/btcticker.html
```diff
@@ -0,0 +1,5 @@
+<div class="pureFlex cent currentBTCPrice clrP clrBr clrSh3 tx4 contentBox padSm">
```
jjeffryes: Why does this have new classes instead of using flexVCent? As far as I can tell, the existing flexVCent would result in the same layout.

js/templates/modals/baseModal.html
```diff
@@ -1,4 +1,22 @@
+<div class="popInMessageHolder js-popInMessages"></div>
 <div class="<%= ob.modelContentClass %>">
-  <i class="jsModalClose <%= ob.closeButtonClass %> <% !ob.showCloseButton && print('hide') %>"></i>
+  <div class="topControls pureFlex spread js-closeClickTarget js-btcTicker">
```
jjeffryes: I'm not sure how I feel about putting this section in the baseModal. As we add new modals, this could become very long. What's the reason to put all of the view-specific markup in the base template?

js/views/modals/About/BTCTicker.js
```diff
@@ -0,0 +1,51 @@
...
+export default class extends baseVw {
+  constructor(options = {}) {
+    super({
+      className: 'aboutBTCTicker pureFlex alwaysFirst',
+      ...options,
+    });
+  }
+
+  getCurrentPrice() {
+    return getExchangeRate('USD');
```
jjeffryes: I assume USD is a placeholder here?

styles/components/_layout.scss
```diff
@@ -90,6 +90,30 @@
   flex-direction: column;
 }
 
+.pureFlex {
+  display: flex;
```
jjeffryes: What are the advantages of the pureFlex class over the existing classes?

There are two things we want to achieve with the CSS to consider here:

1. we want to avoid needing a long list of classes on each element. Classes should have more than one property in most cases, and if more than 3 non-color classes are needed on an element, a single class should be created that incorporates those properties. 

2. we want to enforce visual consistency, which we can do by using consistent classes. We should be very hesitant to add more layout classes if classes already exist that do the same thing. Whenever possible the minimum number of classes should be used, as that will make the UX consistent, vs. lots of slightly different classes which will make the UX inconsistent.

js/templates/modals/baseModal.html
rmisio: Yeah, this IMHO absolutely should not go into the baseModal. The baseModal is intended to be very generic and stuff should only go in there that would be applicable for every modal. Certain modals will want to add their own custom content and that should be done in their respective templates.

Doing it this way would would mean the base modal would just keep growing to handle every different permutation we'd need. The general object-oriented approach is to centralize any common functionality into a base class and then have child classes specify the differences.

Could you please revert this to how it was before as well as moving the respective parts that were brought in here back to where they were? Once those are back in place, they should serve as a good example of how to implement some custom topControls content into the about modal. If there's any specific stumbling block you are running into, please let us know and we'll be happy to help.

js/views/modals/About/About.js
```diff
@@ -0,0 +1,95 @@
...
+      this.$tabContent.append(tabView.$el);
+      this.currentTabView = tabView;
+      this.currentTabName = tabViewName;
+      this.$('#tabTitle').text(app.polyglot.t(
+        `about.${this.currentTabName.toLowerCase()}Tab.sectionHeader`));
+    }
+  }
+
+  render() {
+    this.btcTicker = this.createChild(BTCTicker);
```
rmisio: To be safe, I would remove the btcTicker if it already exists:

```
if (this.btcTicker) this.btcTicker.remove();
```

Otherwise, if you re-render this view, you'll have a ghost view left behind. I know that, at this time, this view is never re-rendered, but IMO it's safer if it's in place in case we refactor and do re-render.

js/views/modals/About/About.js
```diff
@@ -0,0 +1,95 @@
...
+    loadTemplate('modals/about/about.html', (t) => {
+      this.$el.html(t({
+        ...this.options,
+        version,
+      }));
+      super.render();
+
+      this.$tabContent = this.$('.js-tabContent .contentBox');
+
+      this.selectTab(this.currentTabName);
```
rmisio: You also could have just:

```
this.$('.js-btcTicker').append(this.btcTicker.el);
```

In this case it's not critical to cache the jQuery selector, since it's not being used anywhere else. If you want to leave it as is, so the selector is cached in case we do need it in the future, that's fine. Totally up to you.

js/views/modals/About/BTCTicker.js
rmisio: You would want to pass in `app.settings.get('localCurrency')` so it's displayed in the user's local currency.

js/templates/modals/about/btcticker.html
rmisio: It doesn't look like the `currentBTCPrice` is being used anywhere. If that's correct, can we remove it then please?

js/templates/modals/about/btcticker.html
```diff
@@ -0,0 +1,5 @@
...
+  <i class="gemSmall" style="background-image: url('../imgs/btcIcon128.png');"></i>
+  $<%= ob.currentBTCPrice %>
```
rmisio: You'll want to remove that $ and use formatCurrency() (it's in the currency module) to get the amount formatted into the user's local currency. For example, in your view's render you can:

```
    loadTemplate('modals/about/btcticker.html', (t) => {
      this.$el.html(t({
        currentBTCPrice: formatCurrency(this.currentBTCPrice, app.settings.get('localCurrency')),
      }));
    });
```

You could also do the formatting in the template, since formatCurrency is available there as a template helper, but completely up to you which way you want to go.

FWIW, you can change your local currency in Settings > General if you want to play around and see the BTC price formatted in different ways. You could also play around with changing the language in that tab, since that also affects the formatting.

js/templates/modals/baseModal.html
tenthhuman: it's not meant as finished. I plan to find a clean way to remove the top control duplication. the current sketch is not ideal and it was quicker than injecting custom markup into top control. I can revert it back. 

and I also think factoring it out is the right move. because we can already see the top control is overloaded for many tasks. it's useful. the number of standard and custom things we want to put into there may continue to grow. it works to have a standard and robust way to style that and organise that to permit both common and custom items in the top control in a clean way with minimal duplication.

styles/components/_layout.scss
tenthhuman: okay. I can do it like the way you say in 1. and I can also see a couple of reasons to do it the way in these changes. 1 may lead to a larger number of style rules to hold the combinations.  smaller rules have clear pupose and combine together in a flexible way . it's a trade-off.  also existing rules define properties unneeded to achieve the layouts in this case. I don't think it works to include those unnecessary properties. for example, absolute positioning and more rules when it can be done using flex is more brittle. finally, I totally agree with 2 and it's about the trade-off. for example, if some existing rules did not style optimally could they not be improved? and if having smaller rules that fit together in flexible combinations made it easier to add new layout and styles that maintained consistency is that not something that works?

when there's time I'll try to present these ideas more closely in some finished code and we can see if it's a contribution that works.

js/views/modals/About/About.js
tenthhuman: okay cool thanks for that consideration.

I think btc ticket is re rendered if the rate updates. we are dynamically updating the rate from some server right?

package.json
tenthhuman: I don't know what's going on with this.

js/templates/modals/about/btcticker.html
tenthhuman: great info in all these currency comments. thank you Rob.

js/views/modals/About/About.js
tenthhuman: that's true. it's not being cached so may as well just use it in one line.

FWIW the overall structure of this BTC ticker integration is not ideal IMO. it's still a WIP. I plan to see if it can be improved some. 

thanks for all your review @jjeffryes and @rmisio

js/templates/modals/baseModal.html
rmisio: I personally think there is too much variance in-between what the modals are doing to abstract out the top-control piece. I think the most flexibility is having individual modals manage it themselves. If we find that we are repeating ourselves (which I don't see right now), we could look into abstracting out the common functionality. But, I certainly wouldn't put that in the baseModal. Modals are implemented in a variety of ways, for example, the BaseModal view and template are used for error / message dialogs (which are small dialogs in the center of the screen). IMO, the base modal should stay very generic.

I would recommend you do the top-control part of the About modal inside the About modal template.

js/views/modals/About/About.js
rmisio: > I think btc ticket is re rendered if the rate updates

It's if the About view is re-rendered that you would have a ghost btcTicker view.

> we are dynamically updating the rate from some server right?

It checks every 15 minutes, I believe.

js/views/modals/About/About.js
rmisio: > FWIW the overall structure of this BTC ticker integration is not ideal IMO.

Other than the comments we left, it seems fine to me. I wouldn't spend too much time on it.

package.json
rmisio: I don't know... maybe you need to clean your npm cache. I think it's `npm cache clean`. </shot-in-the-dark>

js/views/modals/About/About.js
tenthhuman: Okay

jjeffryes:
This is an orphaned PR.

PR #262, opened 2017-01-05
Server connect ui

rmisio:
Implements the Connection Management modal (accessible via Settings > Advanced > Connection Management) and the ability to define server configurations and connect to different servers.

*** This does not include wiring in connection switching to the page nav menu. I'll circle back to that at some point later.


docs/connectionIssues.md
```diff
@@ -0,0 +1,36 @@
+Troubleshooting connection issues
```
rmisio: If you have an error attempting to connect to a server, an error message will have a 'need help' link which links to this doc. It's very much a work in progress and can definitely be cleaned up and beefed up.

js/templates/spinner.svg
```diff
@@ -0,0 +1,16 @@
+<svg version="1.1" class="spinner<% if (ob.className) print(` ${ob.className}`) %>" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
```
rmisio: While the spinner svg can be declared as an image url in the markup or CSS, that way won't allow you to colorize the spinner. So, it's preferable to use this template which is available as a template helper, for example:

```
<div><%= ob.spinner({ className: 'optionalClassname' }) %>
```

The css is such that the spinner will be inherited text color. So, if you put a clrT somewhere on the parent, the spinner will be that color.

tenthhuman:
That is totally awesome @rmisio.

js/start.js
```diff
@@ -419,29 +488,7 @@ app.serverConfigs.fetch().done(() => {
...
-          connectToServer();
-        });
-      } else {
-        const validationErr = serverConfig.validationError;
-
-        // This is developer error.
-        throw new Error('There were one or more errors saving an initial server configuration' +
-          `${Object.keys(validationErr).map(key => `\n- ${validationErr[key]}`)}`);
-      }
+      app.connectionManagmentModal.open();
```
tenthhuman: BOOM! 👍 🎉

jjeffryes:
I found an edge case that could cause some issues.

If you start or restart the app, and it is on a user page, and the connection fails before everything is loaded, the connection modal will appear, but it will be covered by one of the data fetch failure modals. 

The only option in the data fetch failure modal is "retry", clicking it will uncover the connection modal briefly, but if you can't fix your connection in that brief window the data fetch failure modal covers it up again and you're stuck.

We may need to set the connection modal to always be on top of any other modals. 

(I created this situation with a throttled connection and killing the server before it was done connecting)

image

image

rmisio:
Holy cow! That is quite the edge case because the server has to successfully connect in order to proceed and start loading the app. Then, it needs to fail in that, likely fraction of a second, before it's fully loaded.

Anyhow, I think I'll account for it by adding a 'Manage Connections' button to those failure dialogs.

styles/modules/modals/_connectionManagement.scss
```diff
@@ -0,0 +1,94 @@
...
+        padding-right: 0;
+      }
+
+      .connectedText {
+        position: relative;
+
+        span {
+          font-size: $tx1;
+          // Icon is getting an abnormally large height, which messes up the vertical centering
+          // of the text. Therefore, we'll absolutely position.
```
jjeffryes: You could also use flexVCent on the parent, which would work on any text size.

js/start.js
```diff
@@ -351,33 +360,84 @@ function start() {
       fetchStartupData().done(() => {
         app.pageNav.navigable = true;
         app.pageNav.setAppProfile();
+        getChatContainer().removeClass('hide');
```
rmisio: For the reason I mentioned in stand-up, I'm very leery of giving any modal a z-index. It opens up a can of worms that I think is much better avoided.

I don't think it's a problem to hide / show chat because it's done in only one circumstance. If there's no connection, it's hidden, otherwise it's visible and never obscured. I believe the UX req is that the chat should never be hidden, unless there's no connection in which case it's not usable.

js/templates/modals/connectionManagement/configurationForm.html
```diff
@@ -0,0 +1,94 @@
+<div class="contentBox padMd clrP clrBr clrSh3">
+  <div class="flexVCent" style="height: 35px">
```
rmisio: Since either way isn't great, I would just tend to leave in what's in place. If you really feel strongly about it, please feel free to submit a PR :)

js/views/modals/connectionManagement/StatusBar.js
```diff
@@ -0,0 +1,73 @@
...
+
+    e.currentTarget.classList.forEach((className) => {
+      if (className.startsWith('js-')) events.push(className.slice(3));
+    });
+
+    if (events.length) {
+      events.forEach(event => {
+        this.trigger(`click${capitalize(event)}`, { view: this, e });
+      });
+    }
```
rmisio: Hmm... maybe, but I often am scoping it to some selector, which I suppose could be passed in as an option. I'm not sure this is global enough to put in the BaseView, so maybe I'd put this in a new view that extends from the BaseView.

It's a good thought. I'll keep it in mind and if I find I keep re-using it, I'll think of the best way to centralize it.

styles/modules/modals/_connectionManagement.scss
rmisio: Well, I actually had to add `flexVCent flexHRight gutterHTn` to the parent, a `tx1` to the icon span and then wrap the "Connected" text in a span so the gutterHTn works.

PR #264, opened 2017-01-05
Fixes the rich text editor in the Page section.

jjeffryes:


js/views/modals/Settings/Page.js
```diff
@@ -228,7 +229,8 @@ export default class extends baseVw {
         headerMinWidth: this.headerMinWidth,
       }));
 
-      this.$formFields = this.$('select[name], input[name], textarea[name]');
+      const formFields = 'select[name], input[name], textarea[name], div[contenteditable][name]';
```
jjeffryes: One thing that's not obvious is you have to update the list of elements that are collected when saving to include the trumbowyg field.

js/views/modals/Settings/Settings.js
```diff
@@ -198,10 +197,6 @@ export default class extends BaseModal {
       this.selectTab(this.$('.js-tab[data-tab="General"]'));
     });
 
-    installRichEditor(this.$('#settingsAbout'), {
-      topLevelClass: 'clrBr',
-    });
-
```
jjeffryes: The rich text editor was moved to the page tab view, it has to be in the view that matches the template, otherwise the element can't be selected because the view can't access it.

js/views/modals/Settings/Page.js
```diff
@@ -309,6 +311,10 @@ export default class extends baseVw {
           this.headerCropper.cropit('imageSrc',
             app.getServerUrl(`ipfs/${this.profile.get('headerHashes').get('original')}`));
         }
+
+        installRichEditor(this.$('#settingsAbout'), {
```
rmisio: @jjeffryes Can you take the init of the trumbo editor outside of the timeout? It doesn't need to be in one and having it in one breaks the auto-scroll to the first error (when saving with a client side error). I just fixed a similar issue on the Edit Listing modal.

rmisio:
@morebrownies You may very well be aware of this, but in case you're not, this one is set to merge into your modalStyleFixes branch. So, if it looks good to you, go ahead and merge it in.

morebrownies:
Perfect, ty. Merged and ready to go.

Issue #61, opened 2016-09-28
Show number of incoming/outgoing connections


JustinDrake:
A common issue for vendors in OB1.0 is that their store seems online to them (they can see their storefront) but actually they are not visible to the outside world.

Similar to other P2P software (e.g. BitTorrent clients, or Bitcoin Core), it would be useful to show the number of connected peers, detailing the number of incoming and outgoing connections.

This would help vendors know when their store is not visible to the outside world, or when their store is being flaky. It may also be "exciting" for store owners to see people connect to their node.

jjeffryes:
We don't have a design up for this, but our plan is to include an area with details on your local network, a list of nodes you're connected to, etc. 

It's definitely useful, both for debugging and just to see what's going on.

hoffmabc:
Related: https://github.com/OpenBazaar/openbazaar-desktop/pull/103

PR #270, opened 2017-01-06
Quick and dirty patch to show user cards in the connected peers.

jjeffryes:
- shows user cards in the connected peers view so it's clear which ones are working profiles
- this is for testing in milestone 1, not production use.

### DIFF
--- a/js/templates/connectedPeersPage.html
+++ b/js/templates/connectedPeersPage.html
@@ -1,12 +1,10 @@
 <div class="pad tx4">
-  <% console.log(ob) %>
   <h1><%= ob.polyT('connectedPeersPage.heading') %></h1>
   <p><em><%= ob.polyT('connectedPeersPage.infoMessage') %></em></p>
 
   <div class="pageContent">
-    <div class="userPageFollow flexRow js-peerWrapper">
+    <div class="userPageFollow flexRow js-peerWrapper"></div>
 
-    </div>
     <div class="js-morePeers hide">
       <hr class="clrBr">
       <a class="btn floR clrBr clrP js-morePeersBtn">Load More</a>
@@ -16,8 +14,4 @@
   <% if (!ob.peers.length) { %>
     <p><%= ob.polyT('connectedPeersPage.noResults') %></p>
   <% } %>
-
-
-
-
 </div>
--- a/js/templates/connectedPeersPage.html
+++ b/js/templates/connectedPeersPage.html
@@ -2,15 +2,22 @@
   <% console.log(ob) %>
   <h1><%= ob.polyT('connectedPeersPage.heading') %></h1>
   <p><em><%= ob.polyT('connectedPeersPage.infoMessage') %></em></p>
+
+  <div class="pageContent">
+    <div class="userPageFollow flexRow js-peerWrapper">
+
+    </div>
+    <div class="js-morePeers hide">
+      <hr>
+      <a class="btn js-morePeersBtn">Load More</a>
+    </div>
+  </div>
+
   <% if (!ob.peers.length) { %>
     <p><%= ob.polyT('connectedPeersPage.noResults') %></p>
   <% } %>
-  <div class="userPage">
-    <div class="pageContent">
-      <div class="userPageFollow flexRow js-peerWrapper">
 
-      </div>
-    </div>
-  </div>
+
+
 
 </div>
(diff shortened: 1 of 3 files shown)


js/templates/connectedPeersPage.html
```diff
@@ -1,11 +1,23 @@
 <div class="pad tx4">
+  <% console.log(ob) %>
```
rmisio: Stray console statement here.

js/templates/connectedPeersPage.html
```diff
@@ -1,11 +1,23 @@
...
+    <div class="js-morePeers hide">
+      <hr class="clrBr">
+      <a class="btn floR clrBr clrP js-morePeersBtn">Load More</a>
+    </div>
+  </div>
+
+  <% if (!ob.peers.length) { %>
     <p><%= ob.polyT('connectedPeersPage.noResults') %></p>
   <% } %>
+
```
rmisio: Can we clean up this white space here please?

js/templates/connectedPeersPage.html
```diff
@@ -1,11 +1,23 @@
...
   <p><em><%= ob.polyT('connectedPeersPage.infoMessage') %></em></p>
-  <% if (ob.peers.length) { %>
-    <ul>
-    <% ob.peers.forEach(peer => print(`<li><a href="#${peer}">${peer}</a></li>`)) %>
-    </ul>
-  <% } else { %>
+
+  <div class="pageContent">
+    <div class="userPageFollow flexRow js-peerWrapper">
+
```
rmisio: Can we clean up this empty line?

js/views/ConnectedPeersPage.js
```diff
@@ -11,6 +12,48 @@ export default class extends baseVw {
...
+
+  get $morePeers() {
+    return this._$morePeers ||
+        (this._$morePeers = this.$('.js-morePeers'))
+  }
+
+  loadPeers() {
+    if (this.peers.length > this.peersLoaded) {
+      this.peers.slice(this.peersLoaded, this.peersLoaded + this.peersIterator).forEach((peer) => {
+        console.log(peer)
```
rmisio: Stray console statement here.

js/views/ConnectedPeersPage.js
```diff
@@ -11,6 +12,48 @@ export default class extends baseVw {
...
+        const user = this.createChild(userShort, {
+          guid: peer,
+        });
+        this.$peerWrapper.append(user.render().$el);
+      });
+      this.peersLoaded += this.peersIterator;
```
rmisio: There's a good chance the last page will not be a full page, so for peersLoaded to be accurate, you would want to add the amount you sliced above.

But... it is inconsequential here. For what we are using peersLoaded for, it doesn't matter if it's wrong, as long as it's over not under.

js/views/ConnectedPeersPage.js
```diff
@@ -11,6 +12,48 @@ export default class extends baseVw {
...
+  }
+
+  loadPeers() {
+    if (this.peers.length > this.peersLoaded) {
+      this.peers.slice(this.peersLoaded, this.peersLoaded + this.peersIterator).forEach((peer) => {
+        console.log(peer)
+        const user = this.createChild(userShort, {
+          guid: peer,
+        });
+        this.$peerWrapper.append(user.render().$el);
```
rmisio: Could we add these to a document fragment and then add the whole fragment instead of one by one to the DOM?

rmisio:
Could you also please fix the linting errors?

js/views/ConnectedPeersPage.js
jjeffryes: I'll change the variable name, it's not actually intended to track the amount loaded, just the amount it's attempting to load. How it works here is intentional.

js/views/ConnectedPeersPage.js
jjeffryes: Yup.

jjeffryes:
I'll clean it up. It was a spur of the moment thing on my older computer (I got inspired while checking an old Bitcoin wallet), that computer didn't have linting set up the way we do now.

tenthhuman:
That is super nice. 
People can explore the network like this.
It would be cool if this was more accessible. Instead of hidden in settings.

tenthhuman:
This closes #61.

PR #273, opened 2017-01-09
Adds "Peers" tab to main tab nav. Also fixes i18n for words there. Pairs well with #270

tenthhuman:
**Commit detail:** 👾 

 - Added a "Peers" tab handle for the Connected Peers page to the main tab navigation.
 - Converts English words to i18n values for links in main nav: "Home", "Store", "Followers", "Following"
 

**What this does:** 🎉 🥇 💯 

When merged with #270 the new "Connected Peers" view from @jjeffryes is linked in the main tab bar. 

The idea is that this useful way ( the new Connected Peers tab ) to explore your local network is more accessible. 

**Lineage:** 👑 

This builds on the peers pages lineage: #270 by @jjeffryes and #103 by @rmisio 
And makes it more accessible by putting it in the main tab nav.

**Additional i18n bonus:** 🎏 

The words in main tab navigation were plain English. These are replaced with i18n key references in this change.

**Bling:** 💎 

The new connected peers tab by @jjeffryes

### DIFF
--- a/js/languages/en-US.json
+++ b/js/languages/en-US.json
@@ -27,6 +27,13 @@
   "serverConnect": {
     "defaultServerName": "Default Server"
   },
+  "mainNav": {
+    "followers": "Followers",
+    "following": "Following",
+    "home": "Home",
+    "store": "Store",
+    "peers": "Peers"
+  },
   "userPage": {
     "followsYou": "Follows You",
     "message": "Message",
--- a/js/templates/userPage/userPage.html
+++ b/js/templates/userPage/userPage.html
@@ -12,7 +12,7 @@
             <% } %>></div>
         </div>
         <div class="flexExpand">
-          <h1 class="h2 txUnb txUnl"><%= ob.name %></h1>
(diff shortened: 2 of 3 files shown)


rmisio:
@tenthhuman There already was a link to the Connected Peers in the main page nav, accessible when you click the user avatar on the far top right of the app. That's were we put general links like this.

I'm not sure we want to add it to the user page like this, since this really isn't a tab. It's a link to another page.

@morebrownies What do you think?

js/languages/en-US.json
```diff
@@ -27,6 +27,13 @@
   "serverConnect": {
     "defaultServerName": "Default Server"
   },
+  "mainNav": {
+    "followers": "Followers",
```
rmisio: These are stings in the User Page and therefore should be embedded in the userPage key.

jjeffryes:
This is an interesting idea, but I think we should keep it as a separate page.

morebrownies:
+1 to separate page link in user navigation

jjeffryes:
I think we have consensus that we shouldn't put this in the user page.

js/languages/en-US.json

PR #274, opened 2017-01-09
Add count to peers

### LINKED ISSUE #61: Show number of incoming/outgoing connections

jjeffryes:
- adds a count to the connected peers page
- tweaks the styles and text
- userShort views that are not found on the network are disabled, so their buttons can't be clicked to fire actions that won't work.
- fully closes #61

### DIFF
--- a/js/languages/en-US.json
+++ b/js/languages/en-US.json
@@ -22,7 +22,7 @@
   "connectedPeersPage": {
     "heading": "Connected Peers",
     "totalPeers": "%{totalPeers} Found",
-    "infoMessage": "If you do not see any peers below, you are not connected to the OpenBazaar network. If a peer does not have information available on the network, clicking on their guid below will result in a User Not Found page.",
+    "infoMessage": "You should see up to 12 peers below. If you are connected to more than 12 peers, click the more button to load the next 12.",
     "noResults": "No connected peers"
   },
   "serverConnect": {
--- a/js/languages/en-US.json
+++ b/js/languages/en-US.json
@@ -22,7 +22,6 @@
   "connectedPeersPage": {
     "heading": "Connected Peers",
     "totalPeers": "%{totalPeers} Found",
-    "infoMessage": "You should see up to 12 peers below. If you are connected to more than 12 peers, click the more button to load the next 12.",
     "noResults": "No connected peers"
   },
   "serverConnect": {
--- a/js/languages/en-US.json
+++ b/js/languages/en-US.json
@@ -22,7 +22,7 @@
   "connectedPeersPage": {
     "heading": "Connected Peers",
     "totalPeers": "%{totalPeers} Found",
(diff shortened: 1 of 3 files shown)


js/languages/en-US.json
```diff
@@ -21,7 +21,8 @@
   },
   "connectedPeersPage": {
     "heading": "Connected Peers",
-    "infoMessage": "Please note: If any peer does not have their profile set, clicking on their guid below will result in a User Not Found page.",
+    "totalPeers": "%{totalPeers} Found",
+    "infoMessage": "If you do not see any peers below, you are not connected to the OpenBazaar network. If a peer does not have information available on the network, clicking on their guid below will result in a User Not Found page.",
```
rmisio: > If you do not see any peers below, you are not connected to the OpenBazaar network.

This will never be applicable to the user because if they are not connected to a server, the Connection Management modal will be up and obscuring everything else.

With that said, should we remove that part and save the user the cognitive load of a case that will never be true for them?

rmisio: Ok, actually this whole message could go away since nodes for which we can't obtain info, will be unclickable (and the disabled state makes that obvious).

jjeffryes: Aren't there cases where users can connect to their server, but their server can't see any other nodes?

I know this happens in v1.

js/templates/connectedPeersPage.html
```diff
@@ -1,17 +1,20 @@
...
-    <div class="userPageFollow flexRow js-peerWrapper"></div>
-
-    <div class="js-morePeers hide">
-      <hr class="clrBr">
-      <a class="btn floR clrBr clrP js-morePeersBtn">Load More</a>
+<div class="pageContent">
+  <div class="rowLg">
+    <div class="flex">
+      <h1 class="flexExpand"><%= ob.polyT('connectedPeersPage.heading') %></h1>
+      <div class="tx3 btn"><%= ob.polyT('connectedPeersPage.totalPeers', { totalPeers: ob.peers.length }) %></div>
```
rmisio: I don't think this should get a btn class, since that is giving it a hover state.

js/languages/en-US.json
jjeffryes: Good point on the part about nodes you can't click on.

js/templates/connectedPeersPage.html
jjeffryes: I'll put some other class on it. It just needs to be visually separated.

rmisio:
The man, the myth, the legend has said the sub-text message can go away:

image

Aside from that, this one can be merged as far as I could see.

PR #275, opened 2017-01-09
Resizes select2 tag inputs to match other input heights.

jjeffryes:
@morebrownies this patch will change the select2 inputs with tags to be 40px tall, like the other inputs.

However, it does so by making the tags inside the input shorter (they were 40px tall, which caused the overall input to be 52px tall after padding was added). 

Check and see if this is the effect you were looking for.

image


morebrownies:
Sweet looks much better thank you

PR #284, opened 2017-01-11
More robust style for the border fill in.

jjeffryes:


jjeffryes:
This doesn't match up on Mac, we'll need to come up with a better solution that works on all screens and OSes.

PR #285, opened 2017-01-11
Moderator model and settings tab scaffolding

rmisio:
This change set provides the initial scaffolding of the Moderator model as well as the the Moderation tab in settings.

Some things to implement still:
- Validations in the Moderator model. The server is not validating the Profile (or the moderator info which is nested inside it), but there's some validations we can surmise (required fields, length of fields) and we could fill the rest in when the server implements validations on their side.

- Build out the moderator model. Looking at the proto, you will find there's some nested data in there (and some of the nested data has nested data), so nested models need to be created and wired in. Keep in mind that validations of nested attributes should go inside the respective nested model. They'll automatically be integrated into the parent when the parent is validated.

- Convert price fields:

Prices need to be converted from Decimals to Integers and vice versa. The conversion should be done in the Profile model. Convert from Decimal to Integer in Parse and convert from Integer to Decimal in Sync. Please look at the `// convert price fields` sections in the Listing model for an example of how to do this.

For the UI, please look at the Create Listing modal to see an example of a text field with an adjacent Currency select dropdown. There should be some CSS you can reuse:

image

Also, please bind JS handlers when the price field changes and format the currency properly (i.e. add in the right amount of decimals using `formatPrice` from the currency module). Please see `change .js-price': 'onChangePrice` in `EditListing.js` as an example.


js/models/BaseModel.js
```diff
@@ -121,17 +123,30 @@ export default class extends Model {
...
+            }
           }
         });
       }
     }
 
-    return super.set(attrs, opts);
+    const superSet = super.set(attrs, opts);
+
+    // Since the standard change event doesn't properly take into
```
rmisio: An example best illustrates why this event is necessary... Let's say we have a Model with a nested Model, e.g.

```
const luckyLarry = new Human({
  eyeColor: 'blue',
  car: {
    make: 'Toyota',
    model: 'Corolla',
    year: 2006,
  }
});

luckyLarry.on('change', () => { // do something });
luckyLarry.get('car').on('change', () => { // do something });

luckyLarry.set('car', {
  year: 2008,
});
```

By default, the second change event will fire, but not the first, since the instance isn't changing. So, I've temporarily introduced a `someChange` event that will fire if any part of the model (nesting and all) changes, whereas you could still directly bind to nested instances if you need to know whether a specific nested attribute changed.

I've actually brainstormed and I think I have a good strategy to integrate this behavior into the standard change event (without modifying Backbone source!), but there's a little bit of work there to also make changedAttributes() and hasChanged() work as well, so this `someChange` event is a bridge to that happening at some point.

js/models/Settings.js
```diff
@@ -16,6 +16,8 @@ export default class extends BaseModel {
       refundPolicy: '',
       blockedNodes: [],
       storeModerators: [],
+      shippingAddresses: new ShippingAddresses(),
```
rmisio: This is the new set-up to differentiate between an optional nested model and a required one... in either case, you are still declaring them in the nested getter:

```
get nested() {
  return {
    nestedModelRequired: Model,
    nestedModelOptional: Model,
  }
}
```

The difference is that the required model should be instantiated in the default method:

```
defaults() {
  return {
    nestedModelRequired: new Model(),
  }
}
```

Basically, what this structure does is that when the parent model is created, any nest model instantiated in the defaults will also be created, whereas "optional" models will be created the first time they are set.

js/views/modals/Settings/Settings.js
```diff
@@ -75,104 +71,6 @@ export default class extends BaseModal {
     }
   }
 
-  saveClick() {
```
rmisio: This is code I forgot to remove when I was refactoring settings to integrate with the new modal design. Now tabs handle saving on their own.

js/models/BaseModel.js
tenthhuman: Seems like good progress. Keep going!

js/models/Settings.js
tenthhuman: This is such an improvement on how it was before. Great work!

PR #295, opened 2017-01-12
Change the bundled server location

hoffmabc:
Currently it references the 1.x location folder.

### DIFF
--- a/main.js
+++ b/main.js
@@ -75,7 +75,7 @@ if (handleStartupEvent()) {
   console.log('OpenBazaar started on Windows...');
 }
 
-const serverPath = `${__dirname}${path.sep}..${path.sep}OpenBazaar-Server${path.sep}`;
+const serverPath = `${__dirname}${path.sep}..${path.sep}openbazaar-go${path.sep}`;
 const serverFilename = process.platform === 'darwin' || process.platform === 'linux' ?
       'openbazaard' : 'openbazaard.exe';
 const isBundledApp = _.once(() => fs.existsSync(serverPath + path.sep + serverFilename));


main.js
```diff
@@ -75,7 +75,7 @@ if (handleStartupEvent()) {
   console.log('OpenBazaar started on Windows...');
 }
 
-const serverPath = `${__dirname}${path.sep}..${path.sep}OpenBazaar-Server${path.sep}`;
+const serverPath = `${__dirname}${path.sep}..${path.sep}openbazaar-go${path.sep}`;
```
rmisio: @jjeffryes @hoffmabc 

Do you guys see any chance that someone using the stand-alone app (from source or otherwise) would have an openbazaar-go folder one directory up from where they are running the client?

I think it's unlikely, but if it did happen, then the client would think it's running the bundled app and some weird things could happen.

Would it be safer and reasonable do-able to have the bundled app install the server in a more uniquely named folder, like `OpenBazaar-Bundled-Server`?

rmisio: Hmm... now that I think about it, an alternative would be to have the bundled app start the client with a command line flag (e.g. `app-type="packaged"`).

That would give the added benefit in that we could also pass in a flag when it's a stand-alone client *exectuable*  (e.g. `app-type="packaged-client-only"`) since I remember there being a use-case for the client knowing it's being run that way (as opposed to from source).

What do you think of that? (for next release)

hoffmabc: a) That folder would also have to have the daemon binary but yes it's possible.

```
const isBundledApp = _.once(() => fs.existsSync(serverPath + path.sep + serverFilename));
```

hoffmabc: b) I think that's a good option and much less brittle.

rmisio: > a.) That folder would also have to have the daemon binary but yes it's possible.

Oh right. Yeah, that makes it even less likelier

> b) I think that's a good option and much less brittle.

Ok, I create a bug #296. It's low priority for now. I don't think we'd need it until we have a use case for the client needing to know if it was started via the client-only package.

hoffmabc: LGTM

rmisio:
@jjeffryes feel free to merge if this one LGTY.

PR #299, opened 2017-01-12
294 connection bugs

### LINKED ISSUE #294: New connection bugs
If the client has no stored connections (by running localStorage.clear(), or starting a new client), the connection modal will come up normally. When a new connection is created, it will connect and then never leave the loading state. 

image

Refreshing the client will also result in being stuck in the loading state, with this console error:

```
start.js:547 Uncaught TypeError: Cannot read property 'get' of undefined
```
The error is fired because "activeServer" is undefined, because if  app.serverConfigs.activeServer is undefined in line 536, then setting it in 541 doesn't set activeServer.

Restarting again at this point will load everything normally.

rmisio:
closes #294


hoffmabc:
@jjeffryes if you can review and LGTM/merge this I will make another tagged build for testing.

PR #300, opened 2017-01-12
Linux build fixes

hoffmabc:
I moved it to a token called openbazaar2 which does not match the openbazaar-go binaries which are just openbazaar. Also the grunt configuration files are updated now.


hoffmabc:
Ok tested this @jjeffryes on a Linux machine.

PR #307, opened 2017-01-13
Page nav conn management

rmisio:
Adds the ability to manage server connections from within a submenu in the main navigation menu.

### DIFF
--- a/js/views/PageNav.js
+++ b/js/views/PageNav.js
@@ -27,6 +27,10 @@ export default class extends View {
         'click .js-navSettings': 'navSettingsClick',
         'click .js-navCreateListing': 'navCreateListingClick',
         'click .js-navListItem': 'onNavListItemClick',
+        'mouseenter .js-connectedServerListItem': 'onMouseEnterConnectedServerListItem',
+        'mouseleave .js-connectedServerListItem': 'onMouseLeaveConnectedServerListItem',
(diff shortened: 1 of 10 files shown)


js/views/PageNav.js
```diff
@@ -121,24 +142,61 @@ export default class extends View {
...
-      this.$navOverlay.addClass('open');
-    } else {
-      this.$popMenus.removeClass('open');
-      this.$navOverlay.removeClass('open');
+  togglePopMenu() {
+    this.$navList.toggleClass('open');
+    this.$navOverlay.toggleClass('open');
+
+    if (!this.$navList.hasClass('open')) {
+      this.$connManagementContainer.addClass('open');
```
jjeffryes: Isn't this adding the open class to the connection management container if the menu is closed?

It can result in the connection management container being open by itself when the menu is gone if you open it, and click away on the page. 

image

Can it just be removed?

js/views/PageNavServersMenu.js
```diff
@@ -0,0 +1,106 @@
...
+      }));
+
+    this.listenTo(serverConnectEvents, 'connected',
+      e => this.setState({
+        connectedServer: e.server.id,
+      }));
+  }
+
+  className() {
+    return 'listBox clrBr listBox clrP clrSh1';
```
jjeffryes: You have listBox twice here.

jjeffryes:
This looks good. If you can fix the conflicts with master I'll merge it.

PR #311, opened 2017-01-13
Adds a milestone indicator to the index.

jjeffryes:


rmisio:
This thing is ugly as heck, especially with that Feedback button there too, but I guess Form should take a backseat to Function.

@morebrownies Are you alright with this?

jjeffryes:
The intention is to have some thing in place that @morebrownies can shine up when he's able to.

And I didn't think it was *that* ugly. ;)

rmisio:
Ok, sorry :) I don't think the indicator by itself is so ugly, but the way the feedback button overlaps it makes for quite the busy bottom right corner.

If it's important to have something in right away... please feel free to merge.

PR #322, opened 2017-01-17
Handle tbt ccurrency

jjeffryes:
- saves whether the server is using the testnet to app.testnet
- if the server is using testnet, shows a blue bitcoin icon in the page nav
- if the user hovers over the icon, a tooltip explaining they are using the testnet appears
- also fixes a display bug that was hiding the vertical bars n the page navigation.
- also adds placeholders for the bookmarks and search buttons.

image


jjeffryes:
@morebrownies this is something you can tweak or restyle when you have time.

Note the two vertical bars on the navbar which were missing due to a change a while ago, I hadn't noticed they were gone until now.

js/templates/pageNav.html
```diff
@@ -30,6 +30,11 @@
           </div>
         </div>
       </div>
+      <% if(ob.testnet) { %>
```
rmisio: Could we get a space after the if, please?

styles/modules/_pageNav.scss
```diff
@@ -197,4 +197,19 @@
...
+    padding: $pad;
+
+    .testnetIcon {
+      display: block;
+      height: 29px;
+      width: 29px;
+      background: url('../../imgs/testnetIcon.png') no-repeat;
+      background-size: contain;
+    }
+
```
rmisio: Are the two empty lines here intentional?

morebrownies:
Hey @jjeffryes I'm guessing you probably missed the testnet badge designs in Zeplin: https://zpl.io/2uzXvw

jjeffryes:
@morebrownies I totally missed that! I'll update this.

rmisio:
Do we still need that testnetIcon png?

jjeffryes:
Nope. I forgot to delete the icon file.

PR #332, opened 2017-01-23
318 simple ship regions

### LINKED ISSUE #113: Scrolling While Select2 is In Focus Does Bad Things
If you click into any Select2 field inside the Settings modal, and then scroll, all the select2 dropdowns, the shipping tiles, and the photos stay fixed in the modal, while the other content scrolls under them. 

Note: we are planning to refactor this modal to match to the new full-page scrolling design, doing that may fix this issue.

image

### LINKED ISSUE #318: Add Simplified Regions
Note: this is a simplification of #301, which is being deferred until we have more data on whether it is necessary.

## Functional Requirements
-[ ] Near the shipping countries element, there should be 3 buttons
    - European Union
    -[ ] adds all of the countries in the European Union to the currently selected countries
    - European Economic Area
    -[ ] adds all of the countries in the European Economic Area to the currently selected countries
    - Clear All
    -[ ] Clears all the countries from the select list.

## Data 

European Union: 
'AUSTRIA', 'BELGIUM', 'BULGARIA', 'CROATIA', 'CYPRUS', 'CZECH_REPUBLIC', 'DENMARK', 'ESTONIA', 'FINLAND', 'FRANCE', 'GERMANY', 'GREECE', 'HUNGARY', 'IRELAND', 'ITALY', 'LATVIA', 'LITHUANIA', 'LUXEMBOURG', 'MALTA', 'NETHERLANDS', 'POLAND', 'PORTUGAL', 'ROMANIA', 'SLOVAKIA', 'SLOVENIA', 'SPAIN', 'SWEDEN', 'UNITED_KINGDOM'

European Economic Area:
All of the above plus 'ICELAND', 'LIECHTENSTEIN', 'NORWAY'

rmisio:
closes #318 
closes #113

### DIFF
--- a/js/templates/modals/editListing/shippingOption.html
+++ b/js/templates/modals/editListing/shippingOption.html
@@ -10,10 +10,15 @@
         <hr class="clrBr rowMd" />
         <label for="<%= `shipDestinationsSelect_${ob.cid}` %>" class="required"><%= ob.polyT('editListing.shippingOptions.shippingDestinations') %></label>
         <% if (ob.errors['regions']) print(ob.formErrorTmpl({ errors: ob.errors['regions'] })) %>
-        <div class="select2TagsWrap showDropdown">
+        <div class="select2TagsWrap showDropdown js-shipDestinationsSelect">
           <select id="<%= `shipDestinationsSelect_${ob.cid}` %>" multiple name="regions" class="clrBr clrP clrSh2 select2Tags">
-            <% ob.regions.forEach(region =>
-                print(`<option value="${region.value}" selected="selected">${region.text}</option>`)) %>          
+            <%
+              ob.countrySelectOptions.forEach(option => {
+                const selected = ob.regions.indexOf(option.id) > -1;
+
+                print(`<option value="${option.id}" ${selected ? 'selected="selected"' : ''}>${option.text}</option>`);
+              });
+            %>
           </select>
           <div id="<%= `shipDestinationsPlaceholder_${ob.cid}` %>" class="placeholder tagsPlaceholder"><%= ob.polyT('editListing.shippingOptions.regionsPlaceholder') %></div>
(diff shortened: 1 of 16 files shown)


package.json
```diff
@@ -73,7 +73,7 @@
     "multihashes": "^0.2.2",
     "node-polyglot": "^2.0.0",
     "open": "0.0.5",
-    "select2": "^4.0.3",
+    "select2": "4.0.3",
```
rmisio: Even though we are using a modified source of the select2 JS, I left it in NPM because we're still referencing the CSS file from there. Now that I think about it, I think we should just self-host the CSS as well. Otherwise, there's a risk that a developer will import and use the NPM select2 JS instead of our MacGyver'd version.

rmisio: I'll fix that up on Wed.

rmisio:
Ok, this one is ready to go. I've uninstalled select2 from NPM, so we are completely self-hosting it. Another alternative would be to fork the repo and then we could reference the forked repo in NPM... but I think this is probably ok for now.

js/templates/modals/editListing/shippingOption.html
```diff
@@ -5,20 +5,29 @@ <h2 class="h4 clrT flexExpand <% if (ob.listPosition === 1) print('required') %>
...
         </div>
       </div>
+      <% if (ob.errors['regions']) print(ob.formErrorTmpl({ errors: ob.errors['regions'] })) %>
+      <div class="select2TagsWrap showDropdown js-shipDestinationsSelect">
+        <select id="<%= `shipDestinationsSelect_${ob.cid}` %>" multiple name="regions" class="clrBr clrP clrSh2 select2Tags">
+          <%
+            ob.countrySelectOptions.forEach(option => {
+              const selected = ob.regions.indexOf(option.id) > -1;
+
+              print(`<option value="${option.id}" ${selected ? 'selected="selected"' : ''}>${option.text}</option>`);
```
jjeffryes: You don't need selected=selected, just the word selected.

Issue #338, opened 2017-01-26
Add Link to Dispute Resolution Guidelines to Moderation Tab


jjeffryes:
The current link to the Dispute Resolution Guidelines in the moderation tab is a dummy link. 

When the page exists on openbazaar.org, add a link to it.

PR #339, opened 2017-01-26
Moderation tab

### LINKED ISSUE #254: Add Moderation Tab to Settings Modal
image

The Moderation tab is the section in the Settings Modal that allows a user to set their status as being available as a moderator or not, along with their moderator settings. 

## Functional Requirements

- [x] A control to set moderation status to on or off. This control also shows the current status.
- [x] A drop down to select fee options.
    - Percentage of order
    - Fixed fee,
    - Fixed fee + percentage
- [x] An input to set the percentage
- [x] An input to set the fixed fee. 
- [x] A select to set currency of the fixed fee, defaulted to the user's currency if not set yet.
- [x] A text area to set the description. The limit is 300 characters. 
- [x] A text area to set the terms. The limit is 10,000 characters. This box should be stretchable if this can be vertical stretching only.
- [x] A checkbox to confirm the user has read the OpenBazaar Dispute Resolution Guidelines.
- [ ] A link to the OpenBazaar Dispute Resolution Guidelines that opens in an external browser.
- [x] A checkbox to confirm the user understands the requirements to offer dispute resolution.
- [x] A "I agree to offer dispute resolution to anyone that requests my moderation services".
- [x] A save button.

## Data

The data for this modal comes from the Moderator API 
- moderator: boolean, true means this user is a moderator
- moderation_fee: a number, up to two decimal places, it is a percentage of the transaction being moderated

The following parameters do not currently exist in the API:
- type of fee
- either another field for a fixed fee, or moderation_fee acts as a fixed price when the fee type is not percentage
- moderator description, a string
- moderator terms, a string
- languages, an array of language codes

TODO: put moderator data in Profile, and Profile API instead of moderator API
- set description limit to 300 characters
- set terms limit to 10,000 characters
- move languages to the page tab (UI and Backend)
- validate language codes on the server

jjeffryes:
Closes #254, except for the link to the Dispute Resolution Guidelines. See #338.


js/models/profile/FixedFee.js
```diff
@@ -0,0 +1,44 @@
...
+    };
+
+    const currencies = getCurrenciesSortedByCode();
+
+    if (typeof attrs.currencyCode !== 'string') {
+      addError('feeTypeNoCurrency', app.polyglot.t('settings.moderationTab.errors.noCurrency'));
+    }
+
+    // make sure currency is a valid value
+    if (currencies.indexOf(attrs.currencyCode) === -1) {
```
rmisio: I would just use `getCurrencyByCode(code)` here. You're currently using `getCurrenciesSortedByCode`, but there's no reason to need a sorted list here, so it's probably better to save a cycle or two of processing that's needed to sort the list (or at least not couple the sorting to this model if it's not needed by the model).

rmisio: That would also fix the fact that indexOf doesn't work as you expect here because 'getCurrenciesSortedByCode' is not return a list of currency strings, but a list of currency objects.

js/templates/modals/settings/moderation.html
```diff
@@ -5,12 +5,15 @@ <h2 class="h3 clrT">Moderation</h2>
       <a class="btn clrP clrBr clrSh2 modalContentCornerBtn js-save"><%= ob.polyT('settings.btnSave') %></a>
```
rmisio: There's a layout issue:

1. set the fee type to percentage.
2. click save.
3. set the fee type to fixed.

The currency field is cut-off:

image

js/templates/modals/settings/moderation.html
```diff
@@ -5,12 +5,15 @@ <h2 class="h3 clrT">Moderation</h2>
...
     </div>
     <hr class="clrBr" />
+    <h3><%= ob.polyT('settings.moderationTab.disputeResolution') %></h3>
```
rmisio: Maybe an edge case, but since we have the space, should we make the width of the percentage field be able to support 2 digits and 2 decimal places without being cut-off?

image

js/templates/modals/settings/moderation.html
```diff
@@ -29,23 +32,137 @@ <h2 class="h3 clrT">Moderation</h2>
...
+                    <input type="number"
+                           name="modInfo.fee.percentage"
+                           value="<%= perc %>"
+                           data-var-type="number">
+                  </div>
+                </div>
+                <div class="col2 js-feeFixedInput <% if (!ob.fee || (ob.fee && ob.fee.feeType === 'PERCENTAGE')) { print('visuallyHidden') } %>">
+                  <% var amt = ob.fee && ob.fee.fixedFee ? ob.fee.fixedFee.amount : 0;
+                     var ccode = ob.fee && ob.fee.fixedFee ? ob.fee.fixedFee.currencyCode : ob.defaultCurrency;
+                     amt =  amt ? ob.formatPrice(amt, ccode === 'BTC') : 0;
```
rmisio: If someone enters the price as a string, formatPrice will bomb and prevent the error message from the fixedFee model from showing:

image

I think the fix is to only pipe the amount through formatPrice if it's a number, otherwise just let it render as a string and the model validation will catch it:

image

jjeffryes:
Submitting some fixes to the tests shortly.

js/data/currencies.js
```diff
@@ -651,6 +651,10 @@ function getIndexedCurrencies() {
   return _indexedCurrencies;
 }
 
+export function getListofCurrencies() {
+  return _.pluck(currencies, 'code');
+}
```
jjeffryes: This was added because checking a currency with getCurrencyByCode didn't work in the test when the currency was undefined. There may be a better way to do it, though it seems like a simple list of currencies is useful to have.

js/models/profile/FixedFee.js
```diff
@@ -0,0 +1,38 @@
...
+      errObj[fieldName] = errObj[fieldName] || [];
+      errObj[fieldName].push(error);
+    };
+
+    if (is.not.existy(attrs.currencyCode) || typeof attrs.currencyCode !== 'string') {
+      addError('feeType', app.polyglot.t('settings.moderationTab.moderatorModelErrors.noCurrency'));
+    }
+
+    // make sure currency is a valid value
+    if (getListofCurrencies().indexOf(attrs.currencyCode) === -1) {
```
rmisio: Why can't we use getCurrencyByCode() here?

If the only reason that we created getListofCurrencies() was to satisfy a test, I don't that's a good reason to bring it in, as there's cleaner ways to satisfy the test. I'll write more about that in a bit.

js/data/currencies.js
rmisio: I think I'd need more information on the error you're getting in the test, since you shouldn't have an undefined currency list since it's being declared in the module.

There may be some other issue and I'd rather fix the source rather than mask it with a work-around.

I'm not opposed to bringing in a getListofCurrencies function, but I don't really see a use case for it given that getTranslatedCurrencies() and getCurrencyByCode() should be able to satisfy most things.

If it were to stay, I'd want to memoize it so the list isn't being traversed on each call... but I'm doubting its need.

js/languages/en-US.json
```diff
@@ -234,7 +234,54 @@
...
+      "status":{
+        "saving": "Saving moderation settings...",
+        "done": "Moderation settings saved.",
+        "fail": "Saving the moderation settings failed."
+      },
+      "errors": {
+        "confirm": "To become a moderator, you must agree to both statements below.",
+        "save": "The moderation settings were not saved."
+      },
+      "moderatorModelErrors": {
```
rmisio: Can moderatorModelErrors key please be a sibling of the other model error object (e.g.'profileModelErrors')? A model can be used by multiple parts of the UI, so I don't think its errors should be nested inside a UI specific key ('moderationTab' in this case).

js/models/profile/Moderator.js
```diff
@@ -1,45 +1,62 @@
...
+import Fee from './Fee';
 
 export default class extends BaseModel {
   defaults() {
     return {
+      fee: new Fee(),
       description: '',
       termsAndConditions: '',
-      languages: [],
+      languages: app.settings ? app.settings.get('language') : '',
```
rmisio: I think languages is a list so you probably want:

```
app.settings ? [app.settings.get('language')] : []
```

js/views/modals/Settings/Moderation.js
```diff
@@ -21,52 +25,51 @@ export default class extends baseVw {
...
+      'click .js-save': 'saveCheck',
+      'change #moderationFeeType': 'changeFeeType',
     };
   }
 
   getFormData() {
     return super.getFormData(this.$formFields);
   }
 
+  saveCheck() {
```
rmisio: Can we please integrate the save check into the save() method, for example?

```
    this.profile.set(formData);

    if (this.profile.get('moderator') &&!this.$understandRequirements.prop('checked') && !this.$acceptGuidelines.prop('checked')) {
        this.$moderationConfirmError.toggleClass('hide', !blockSave);
        return;
    }
```

This would allow the save() to be self contained and, IMHO, make for a cleaner API. For example, let's say you wanted to initiate the saving of this form from a different view - all you'd need to do is call save(). With the current approach, you'd need to call saveCheck().

js/views/modals/Settings/Moderation.js
```diff
@@ -21,52 +25,51 @@ export default class extends baseVw {
...
-    //   type: 'PATCH',
-    // });
+    const save = this.profile.save(formData, {
+      attrs: formData,
+      type: 'PATCH',
+    });
 
     if (save) {
       const msg = {
         // msg: app.polyglot.t('settings.generalTab.statusSaving'),
```
rmisio: Can we please remove the commented out lines for the status translations?

js/data/currencies.js
jjeffryes: I'll take this out, looking at the code again I was already validating for an undefined currency, and shouldn't be checking getCurrencyByCode if the currencyCode isn't set.

jjeffryes:
Travis is throwing some weird errors on this. I can't reproduce them when running tests locally.

js/views/modals/Settings/Moderation.js
rmisio: There's still a bunch of commented out status message translations. Can they be removed?

```
      const msg = {
        // msg: app.polyglot.t('settings.generalTab.statusSaving'),  // <=== Can we remove these? There's a few in this file.
        msg: app.polyglot.t('settings.moderationTab.status.saving'),
        type: 'message',
      };
```

js/views/modals/Settings/Moderation.js
```diff
@@ -49,24 +47,32 @@ export default class extends baseVw {
...
+      this.$acceptGuidelines.prop('checked');
+
+    if (!this.profile.get('moderator') &&
+    $('input[name=moderator]:checked').val() === 'true' && !confirmChecked) {
+      this.$moderationConfirmError.removeClass('hide');
+      return;
+    }
 
-    const save = this.profile.save();
+    this.$moderationConfirmError.addClass('hide');
```
rmisio: It's not necessary to put the 'hide' class on, since this view is re-rendered at the end of this method and the template will apply it. You'll save a DOM write by eliminating this statement.

js/templates/modals/settings/moderation.html
```diff
@@ -29,23 +32,133 @@ <h2 class="h3 clrT">Moderation</h2>
...
+                         maxlength="5"
+                         name="modInfo.fee.percentage"
+                         value="<%= perc %>"
+                         data-var-type="number">
+                </div>
+              </div>
+              <div class="col2 js-feeFixedInput <% if (!ob.fee || (ob.fee && ob.fee.feeType === 'PERCENTAGE')) { print('visuallyHidden') } %>">
+                <% var amt = ob.fee && ob.fee.fixedFee ? ob.fee.fixedFee.amount : 0;
+                   var ccode = ob.fee && ob.fee.fixedFee ? ob.fee.fixedFee.currencyCode : ob.defaultCurrency;
+                   amt = typeof amt === 'number' ? Number(ob.formatPrice(amt, ccode === 'BTC')) : amt;
```
rmisio: It's pointless to cast the results of formatPrice into a number. Once you put it into a form field, it becomes a string anyways. Which is why `data-var-type="number"` is added to the field. That way getFormData in the baseVw knows to convert the string to a number.

```
let input = document.createElement("INPUT");
input.setAttribute('type', 'text');
input.value = 123; // Set it to a number
console.log(typeof input.value); // still comes out a string

let input = document.createElement("INPUT");
input.setAttribute('type', 'number'); // same behavior even if you use a field type of 'number'
input.value = 123; // Set it to a number
console.log(typeof input.value); // still comes out a string

```

js/models/profile/Moderator.js
```diff
@@ -1,45 +1,62 @@
...
+import Fee from './Fee';
 
 export default class extends BaseModel {
   defaults() {
     return {
+      fee: new Fee(),
       description: '',
       termsAndConditions: '',
-      languages: [],
+      languages: app.settings ? app.settings.get('language') : [],
```
rmisio: You're missing some brackets. The result needs to be an array in either case:

```
languages: app.settings ? [app.settings.get('language')] : [],
```

js/views/modals/Settings/Moderation.js
```diff
@@ -1,9 +1,13 @@
-import _ from 'underscore';
+import $ from 'jquery';
+import 'select2';
```
rmisio: Please use our local mac-gyvered select2 from the 'js/lib' folder.

I did remove the 'select2' package from package.json, but it looks like you need to `npm prune` to get it out of there once you've previously installed it. Then, it will flag that import as an issue, which is what we want.

js/templates/modals/settings/moderation.html
jjeffryes: This was here to truncate trailing zeros when formatPrice returned values like 0.00 or 0.00000000. After merging with master and retesting, the new formatPrice code prevents that issue, so I'll take this out.

PR #340, opened 2017-01-27
335 profile location

### LINKED ISSUE #335: Add Location to Settings/Page
Below the Short Description input on the Page tab of the Settings modal, add an input for location.

The input should be freeform text, with a character limit of 100 (in the current design, only around 50 characters can be shown without being truncated, so we may want a limit that matches that).

If it can be done in around a day or less, it would be ideal to add an auto-suggestion dropdown for city and country as the user types. 

We don't want to call an API for that, so it would require a static open source data set.

rmisio:
partially closes #335

### DIFF
--- a/js/templates/modals/settings/page.html
+++ b/js/templates/modals/settings/page.html
@@ -46,7 +46,7 @@
           </div>
           <div class="col6">
             <% if (ob.errors.location) print(ob.formErrorTmpl({ errors: ob.errors.location })) %>
-            <input type="text" class="clrBr clrSh2" name="location" id="settingsLocation" value="<%= ob.location %>" placeholder="<%= ob.polyT('settings.pageTab.placeholderLocation') %>">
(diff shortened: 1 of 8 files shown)


js/templates/modals/settings/page.html
```diff
@@ -39,6 +39,16 @@ <h2 class="h3 clrT"><%= ob.polyT('settings.pageTab.sectionPage') %></h2>
                       class="clrBr clrSh2" placeholder="<%= ob.polyT('settings.pageTab.placeholderShortDescription') %>"><%= ob.shortDescription %></textarea>
           </div>
         </div>
+        <div class="flexRow">
+          <div class="col3">
+            <label for="settingsHandle"><%= ob.polyT('settings.pageTab.locationLabel') %></label>
```
jjeffryes: This should be for settingsLocation.

jjeffryes: Since it was so minor, and you're travelling, I patched it.

Issue #341, opened 2017-01-31
Currency tests fail due to undefined ajax in test


jjeffryes:
When the automated tests are run, the following errors are  thrown:

```
  2) the currency utility module has functions that involve converting between currencies "before all" hook:
     TypeError: Attempted to wrap undefined property ajax as function
      at Object.wrapMethod (C:\Users\J\ob\openbazaar-desktop\node_modules\sinon\lib\sinon\util\core.js:113:29)
      at Object.stub (C:\Users\J\ob\openbazaar-desktop\node_modules\sinon\lib\sinon\stub.js:67:26)
      at Context.<anonymous> (currency.js:158:20)
      at callFn (C:\Users\J\ob\openbazaar-desktop\node_modules\mocha\lib\runnable.js:326:21)
      at Hook.Runnable.run (C:\Users\J\ob\openbazaar-desktop\node_modules\mocha\lib\runnable.js:319:7)
      at next (C:\Users\J\ob\openbazaar-desktop\node_modules\mocha\lib\runner.js:298:10)
      at Immediate.<anonymous> (C:\Users\J\ob\openbazaar-desktop\node_modules\mocha\lib\runner.js:320:5)

  3) the currency utility module has functions that involve converting between currencies "after all" hook:
     TypeError: Cannot read property 'restore' of undefined
      at Context.<anonymous> (currency.js:235:7)
      at callFn (C:\Users\J\ob\openbazaar-desktop\node_modules\mocha\lib\runnable.js:326:21)
      at Hook.Runnable.run (C:\Users\J\ob\openbazaar-desktop\node_modules\mocha\lib\runnable.js:319:7)
      at next (C:\Users\J\ob\openbazaar-desktop\node_modules\mocha\lib\runner.js:298:10)
      at Immediate.<anonymous> (C:\Users\J\ob\openbazaar-desktop\node_modules\mocha\lib\runner.js:320:5)
```

It doesn't look like these errors happen in Travis, they may be a Windows problem, similar to the issue where jsDoc fails in Windows.

hoffmabc:
I'm not sure what the action is here.

jjeffryes:
This is the bug for those tests that are broken (at least for me) in Windows. See the comment on #342

rmisio:
There's a worrisome inconsistency between the tests on Windows vs. other OSs. It's on my to-do list to set up a Win VM and get to the bottom of the mis-behavior on Windows.

[...]

PR #342, opened 2017-01-31
Fixes broken Listing Test

jjeffryes:
Fixes an issue in the listing test, where the url was malformed and Backbone did not have access to jQuery.

### DIFF
--- a/test/_setup.js
+++ b/test/_setup.js
@@ -16,6 +16,8 @@ global.document = require('jsdom').jsdom(indexPage);
 global.window = document.defaultView;
 global.navigator = window.navigator = {};
 global.$ = require('jquery')(window);
+global.Backbone = require('backbone');
+global.Backbone.$ = global.$;
 
 let getServerUrl;
 
@@ -25,7 +27,7 @@ before(function () {
   // test environement. So... to work around that, we're stubbing
   // getServerUrl.
   getServerUrl = sinon.stub(app, 'getServerUrl',
-    (urlFrag) => `http://localhost:8080/ob/${urlFrag}`);
+    (urlFrag) => `http://localhost:4002/${urlFrag}`);
 });
 
 after(function () {


rmisio:
@jjeffryes The tests work fine for me on the latest master. What error exactly are you getting?

jjeffryes:
@rmisio 2 things, the url always had an ob/ in front of it, even when it should have been /ipfs with no ob/, and it had the old port. 

The second issue was that the sync always failed, because in the test environment Backbone couldn't access jquery.

There must be something different between my environment and yours, because I have multiple tests failing (see also #341).

```
  1) the Listing model does not throw an error if you attempt to fetch with a guid and slug both set:

      AssertionError: expected true to equal false
      + expected - actual

      -true
      +false

      at Context.<anonymous> (listing.js:59:28)
      at callFn (C:\Users\J\ob\openbazaar-desktop\node_modules\mocha\lib\runnable.js:326:21)
      at Test.Runnable.run (C:\Users\J\ob\openbazaar-desktop\node_modules\mocha\lib\runnable.js:319:7)
      at Runner.runTest (C:\Users\J\ob\openbazaar-desktop\node_modules\mocha\lib\runner.js:422:10)
      at C:\Users\J\ob\openbazaar-desktop\node_modules\mocha\lib\runner.js:528:12
      at next (C:\Users\J\ob\openbazaar-desktop\node_modules\mocha\lib\runner.js:342:14)
      at C:\Users\J\ob\openbazaar-desktop\node_modules\mocha\lib\runner.js:352:7
      at next (C:\Users\J\ob\openbazaar-desktop\node_modules\mocha\lib\runner.js:284:14)
      at Immediate.<anonymous> (C:\Users\J\ob\openbazaar-desktop\node_modules\mocha\lib\runner.js:320:5)
```

rmisio:
Hmm... that's quite worrying that there are environment differences. I think we're going to have to get to the bottom of those (maybe I need to set up a win VM) rather than trying to put in fixes that are only needed on a certain environment.

On a different note, I can't see why you would need to bring in Backbone manually and then manually attach jQuery to it. Any test should be importing those if it directly uses Backbone and if it's importing files which depend on them, then they have them declared as dependencies.

Similarly, Backbone has been module friendly for a while now and it has jQuery declared as a dependency, so there shouldn't ever be a need to manually attach jQuery to Backbone.

Somethings a muck here!

rmisio:
Just to confirm... you're getting test errors on master?

jjeffryes:
I agree, something is definitely amuck!

And yes, the tests are broken on Master, and all commits back to the end of December that I re-checked.

hoffmabc:
What's the status on this?

jjeffryes:
I've closed it for now, we need to determine what the issue is with the tests. @rmisio set them up originally, the best route would be for him to see if he can duplicate the problem on a Windows VM.

PR #343, opened 2017-01-31
Cleaned up a number of style issues (address bar, return to store, connection management, etc)

morebrownies:
**Changes**
- Cleaned up spacing in address bar
- Changed search icon 
- Change bookmark icon (we can use the + to simplify) 
- Made search input border always display (Sorry @joshj only showing it on hover was getting annoying. Always seeing the border is growing on me. Looking better now that the + button next to it balances it better)
- Cleaned up spacing/sizing of Return to store button on listing detail overlay
- Added missing shadows to a few places
- Adjusted typography 
- Increased the size of the photo on the listing detail screen in order to shrink down the right column slightly (the balance feels better and will reduce some white space)
- Added shadows to inputs in new connection screen
- Added placeholders to inputs in new connection screen

### DIFF
--- a/js/templates/modals/connectionManagement/configurations.html
+++ b/js/templates/modals/connectionManagement/configurations.html
@@ -4,7 +4,7 @@
       <!-- OB branding comming soon... -->
     </div>  
     <div class="col4">
-      <h2 class="h3 clrT txUnl txCtr"><%= ob.polyT('connectionManagement.configurations.heading') %></h2>
+      <h2 class="h4 clrT txUnl txCtr"><%= ob.polyT('connectionManagement.configurations.heading') %></h2>
     </div>
     <div class="col4">
       <div class="flexHRight">
--- a/js/templates/modals/connectionManagement/configurations.html
+++ b/js/templates/modals/connectionManagement/configurations.html
@@ -1,4 +1,4 @@
-<div class="contentBox padMd clrP clrBr clrSh3">
+<div class="contentBox padMd clrP clrBr clrSh3" style="min-height: 500px">
   <div class="flexVCent">
     <div class="col4">
       <!-- OB branding comming soon... -->
@@ -8,7 +8,7 @@
     </div>
     <div class="col4">
       <div class="flexHRight">
-        <a class="btn clrP clrBr js-btnNew"><%= ob.polyT('connectionManagement.configurations.btnNew') %></a>
+        <a class="btn clrP clrSh2 clrBr js-btnNew"><%= ob.polyT('connectionManagement.configurations.btnNew') %></a>
       </div>
     </div>
   </div>
--- a/js/templates/modals/listingDetail/listing.html
+++ b/js/templates/modals/listingDetail/listing.html
@@ -2,11 +2,11 @@
 
 <div class="topControls gutterHSm flex">
   <% if (ob.vendor) { %>
-  <div class="contentBox clrP clrBr clrT">
-    <div class="padSm gutterHSm overflowAuto flexVCent">
+  <div class="contentBox clrP clrSh3 clrBr clrT">
+    <div class="padSm gutterHSm overflowAuto margRSm flexVCent">
       <a class="clrBr2 clrSh1 disc storeOwnerAvatar flexNoShrink js-storeOwnerAvatar" style="<%= ob.getAvatarBgImage(ob.vendor.avatar) %>"></a>
-      <p class="txUnl clamp"><%= ob.vendor.name %></p>
-      <a class="link flexNoShrink js-goToStore"><% print(ob.openedFromStore ? ob.polyT('listingDetail.returnToStore'): ob.polyT('listingDetail.goToStore')) %></a>
+      <p class="txUnl tx3 clamp"><%= ob.vendor.name %></p>
+      <a class="link flexNoShrink tx6 margRSm js-goToStore"><% print(ob.openedFromStore ? ob.polyT('listingDetail.returnToStore'): ob.polyT('listingDetail.goToStore')) %></a>
     </div>
   </div>
   <% } %>
--- a/js/templates/pageNav.html
+++ b/js/templates/pageNav.html
@@ -16,16 +16,16 @@
       </div>
       <div>
         <div class="flexVCent">
-          <a class="iconBtn ion-chevron-left js-navBack"></a>
+          <a class="iconBtn margLSm ion-chevron-left js-navBack"></a>
           <a class="iconBtn ion-chevron-right js-navFwd"></a>
-          <a class="iconBtn ion-refresh js-navReload"></a>
+          <a class="iconBtn ion-refresh js-navReload margRSm"></a>
         </div>
       </div>
-      <div class="rowDivV clrBrBk"></div>
+      <div class="rowDivV margR clrBrBk"></div>
       <div class="pageNavCenter">
         <div class="flexVCent gutterHSm">
           <div class="searchWrapper">
-            <input type="text" class="js-addressBar flexExpand addressBar clrBr4"
+            <input type="text" class="js-addressBar flexExpand addressBar clrSh2 clrBr4"
                    placeholder="<%= ob.polyT('addressBarPlaceholder') %>" value="<%= ob.addressBarText %>" />
           </div>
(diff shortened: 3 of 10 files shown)


js/templates/modals/connectionManagement/configurations.html
```diff
@@ -1,14 +1,14 @@
-<div class="contentBox padMd clrP clrBr clrSh3">
+<div class="contentBox padMd clrP clrBr clrSh3" style="min-height: 500px">
```
jjeffryes: It would be better to give this a one-off class with the minimum height, and put it in _connectionManagement.scss.

js/templates/modals/listingDetail/listing.html
```diff
@@ -2,11 +2,11 @@
 
 <div class="topControls gutterHSm flex">
   <% if (ob.vendor) { %>
-  <div class="contentBox clrP clrBr clrT">
-    <div class="padSm gutterHSm overflowAuto flexVCent">
+  <div class="contentBox clrP clrSh3 clrBr clrT">
+    <div class="padSm gutterHSm overflowAuto margRSm flexVCent">
```
jjeffryes: It doesn't look like the margRSm does anything here? Or is there a situation where the div has a sibling inside the contentBox?

js/templates/modals/listingDetail/listing.html
```diff
@@ -2,11 +2,11 @@
...
       <a class="clrBr2 clrSh1 disc storeOwnerAvatar flexNoShrink js-storeOwnerAvatar" style="<%= ob.getAvatarBgImage(ob.vendor.avatar) %>"></a>
-      <p class="txUnl clamp"><%= ob.vendor.name %></p>
-      <a class="link flexNoShrink js-goToStore"><% print(ob.openedFromStore ? ob.polyT('listingDetail.returnToStore'): ob.polyT('listingDetail.goToStore')) %></a>
+      <p class="txUnl tx3 clamp"><%= ob.vendor.name %></p>
+      <a class="link flexNoShrink tx6 margRSm js-goToStore"><% print(ob.openedFromStore ? ob.polyT('listingDetail.returnToStore'): ob.polyT('listingDetail.goToStore')) %></a>
```
jjeffryes: The margRSm here is overridden by the gutter class on the parent, so it can be removed. 

If it wasn't overridden, there would be double space to the right of the link, is that what was intended?

js/templates/pageNav.html
```diff
@@ -16,30 +16,30 @@
       </div>
       <div>
         <div class="flexVCent">
-          <a class="iconBtn ion-chevron-left js-navBack"></a>
+          <a class="iconBtn margLSm ion-chevron-left js-navBack"></a>
           <a class="iconBtn ion-chevron-right js-navFwd"></a>
-          <a class="iconBtn ion-refresh js-navReload"></a>
+          <a class="iconBtn ion-refresh js-navReload margRSm"></a>
```
jjeffryes: The extra spacing on the left and right of the buttons seems pretty specific, I think it would make sense to add a class in the page nav style sheet, and put it on the parent (flexVCent) as left and right padding.

js/templates/pageNav.html
```diff
@@ -16,30 +16,30 @@
...
         </div>
       </div>
-      <div class="rowDivV clrBrBk"></div>
+      <div class="rowDivV margR clrBrBk"></div>
```
jjeffryes: Could the margin just be added to the left side of the pageNavCenter class?

js/templates/pageNav.html
```diff
@@ -16,30 +16,30 @@
...
           <% } %>
-          <a class="btn barBtn clrP clrBr ion-navicon-round"></a>
+          <a class="bookmarkPage btn barBtn clrP clrBr clrSh2 tx4 ion-android-add"></a>
         </div>
       </div>
       <div class="rowDivV clrBrBk"></div>
       <div>
         <div class="flexVCent box">
-          <a class="iconBtn ion-eye"></a>
+          <a class="iconBtn margLSm ion-eye"></a>
```
jjeffryes: If you make a class for left and right padding around sets of icons, it could work here too. I think that would make sense as a pattern we'd want to repeat.

To make it more generic, it could also be something like an "iconRow" class that wraps any horizontal set of icons, and puts padding on the left and right side (though this one is missing the padding on the right side?).

js/templates/modals/listingDetail/listing.html
morebrownies: It adds a bit of margin to the right of the container
screenshot 2017-01-31 13 13 07

Without it...
screenshot 2017-01-31 13 13 16

js/templates/modals/listingDetail/listing.html
morebrownies: It adds a bit of margin to the right of the container
screenshot 2017-01-31 13 13 07

Without it...
screenshot 2017-01-31 13 13 16

js/templates/modals/listingDetail/listing.html
jjeffryes: Something funny is going on, I just re-tested it, on my computer it looks like your first screenshot regardless of whether that class is there.

js/templates/pageNav.html
morebrownies: I'm not sure i'm following... I don't want extra spacing on the left and right side. I wanted a bit of extra space on the right side so that the refresh icon has some breathing room and isn't too close to the horizontal line. 

screenshot 2017-01-31 13 16 15

screenshot 2017-01-31 13 17 57

js/templates/pageNav.html
morebrownies: Yeah that works too. I'll switch it over.

js/templates/pageNav.html
jjeffryes: There's a margLSm class on the first link, and a margRSm class on the last link.

Adding the same amount of padding to the right and left side of the container they're in would do the same thing, and be a little cleaner (if we ever add another link, or change the order, it would still work).

morebrownies: Ok I see what you're saying now. I'll switch it over.

js/templates/modals/listingDetail/listing.html
jjeffryes: If you examine the element, do you see it being overridden in the web dev panel? It looks like the screenshot below for me, with the margin for margRSm crossed out.

image

morebrownies: @jjeffryes I was totally looking at the wrong div for some reason. You're right, the margin here doesn't do anything and can be removed. 

I was looking at the margRSm a few lines up. Using the class there actually made the margin work correctly on my computer. 

`<div class="padSm gutterHSm overflowAuto margRSm flexVCent">`

morebrownies:
I think this should be about ready @jjeffryes

js/templates/modals/listingDetail/listing.html
morebrownies: This margin still appears to work for me @jjeffryes. The other one wasn't needed.

js/templates/pageNav.html
```diff
@@ -15,30 +15,30 @@
...
-        <div class="flexVCent">
+        <div class="flexVCent iconPad">
           <a class="iconBtn ion-chevron-left js-navBack"></a>
           <a class="iconBtn ion-chevron-right js-navFwd"></a>
           <a class="iconBtn ion-refresh js-navReload"></a>
         </div>
       </div>
       <div class="rowDivV clrBrBk"></div>
-      <div class="pageNavCenter">
+      <div class="pageNavCenter margL">
```
jjeffryes: Since the pageNavCenter class isn't used anywhere else, can't the margin just be part of that class?

To be consistent and avoid surprises, the right and left sides should both use padding or margin (right now the right side has padding as part of the pageNavCenter class).

styles/components/_buttons.scss
```diff
@@ -198,12 +198,12 @@
   &:before {
     //insert a search icon before the text
     font-family: Ionicons;
```
jjeffryes: Let's take off the font family if it's not using the ionicons font.

morebrownies:
Ok changes pushed @jjeffryes

PR #347, opened 2017-02-01
About modal

### LINKED ISSUE #87: Create About Modal

jjeffryes:
Completes the about modal, and fixes various outstanding issues.

Closes #87


js/templates/modals/about/about.html
```diff
@@ -0,0 +1,45 @@
+<div class="topControls flex js-closeClickTarget js-btcTicker"></div>
```
rmisio: 'js-closeClickTarget's are not needed anymore since we decided not to close the modals on overlay click.

js/views/modals/about/About.js
```diff
@@ -0,0 +1,108 @@
...
+    console.log('checking for an update');
+    // TODO: wire this in
+  }
+
+  render() {
+    // remove any existing tickers from previous renders
+    if (this.btcTicker) this.btcTicker.remove();
+
+    this.btcTicker = this.createChild(BTCTicker);
+    this.btcTicker.render();
```
jjeffryes: The code to only render if the currency was BTC was removed, since the ticker needs to reappear if the user switches their currency to a non-BTC currency, the ticker is always present and the ticker view shows or hides itself instead.

jjeffryes:
@rmisio this has been updated and is ready for a new review.

PR #351, opened 2017-02-07
Coupons

### LINKED ISSUE #303: Add Coupons
## Functional Requirements

- [ ] There is a button to add coupons
    - [ ] When clicked, it will add a coupon control set under the last current 
- [ ] Each coupon control set has the following controls:
    - [ ] A title field (max length of 70)
    - [ ] A Coupon Code field (a string, what is the max character length?)
        - [ ] The coupon code must be unique to this listing
    - [ ] A control to select percentage or fixed discount. When one is selected, the inputs for the percentage or fee appear/disappear as appropriate. 
    - [ ] A Discount field (0.01 to 100 Number as a percentage, cannot be negative)
    - [ ] Fixed fee (must be positive, cannot be higher than the item price)
    - [ ] A delete button that deletes the control set.
- [ ] When the checkout process is completed, the coupon can be added for the discount to be applied to the transaction.

## Data
-[ ]  The server will need the ability to track which coupon codes are associated with 

## TODOS
- Revise the design to include switching between percentage and fixed fee
- Add server capability to retrieve coupon codes based on hashes from the contract.

### No coupons
image

### Coupons added
image

rmisio:
Implements Coupon management into the Edit Listing modal.

closes #303


js/templates/modals/editListing/coupon.html
```diff
@@ -0,0 +1,39 @@
...
+      <%
+        let discountAmount;
+
+        if (typeof ob.priceDiscount !== 'undefined') {
+          discountAmount = typeof ob.priceDiscount === 'number' ?
+            ob.formatPrice(ob.priceDiscount, ob.getCurrency() === 'BTC') :
+            ob.priceDiscount
+        } else {
+          discountAmount = typeof ob.percentDiscount === 'number' ?
+            parseFloat((ob.percentDiscount).toFixed(2)).toString() : ob.percentDiscount;
```
tyler-smith: I know this isn't really my domain here but from past experience it definitely think this  type of logic shout be encapsulated and reusable.

js/templates/modals/editListing/coupon.html
```diff
@@ -0,0 +1,39 @@
+<div class="col4 simpleFlexCol">
```
rmisio: It's needed to make the fields line up if there are any errors:

image

jjeffryes: Ah, ok.

PR #352, opened 2017-02-07
Update moderation tab

### LINKED ISSUE #182: Shipping Option Title
We need to give users a couple easy to select defaults here in addition to the free-form. It's obnoxious to have to think of a name for this and not super clear. 

image

jjeffryes:
- when the fee type in the moderation tab is percentage, the amount value in the fixed fee object is set to zero.
- when the fee type is fixed, the value of the percentage is set to zero.
- this clears unused values.
- the order of the percentage and fixed fee inputs are switched, to match the name of the type.

Closes #182


js/templates/modals/settings/moderation.html
```diff
@@ -87,6 +77,16 @@ <h2 class="h3 clrT">Moderation</h2>
                   <% }); %>
                 </select>
               </div>
+              <div class="col2 js-feePercentageInput <% if (ob.fee && ob.fee.feeType === 'FIXED') { print('visuallyHidden') } %>">
+                <div class="inputPercentWrapper clrBr clrSh2">
+                  <% var perc = ob.fee.percentage || 0; %>
```
rmisio: Why is this necessary? Isn't this happening anyway with the default set to 0 in the model?

js/models/profile/Profile.js
```diff
@@ -185,12 +185,24 @@ export default class extends BaseModel {
...
       if (method !== 'delete') {
-        // convert the amount field
-        if (options.attrs.modInfo && options.attrs.modInfo.fee &&
-          options.attrs.modInfo.fee.fixedFee && options.attrs.modInfo.fee.fixedFee.amount) {
-          const amount = options.attrs.modInfo.fee.fixedFee.amount;
-          const isBTC = options.attrs.modInfo.fee.fixedFee.currencyCode === 'BTC';
-          options.attrs.modInfo.fee.fixedFee.amount = decimalToInteger(amount, isBTC);
+        if (options.attrs.modInfo) {
+          // set the percentage to zero if the type is FIXED so old values aren't kept
+          if (options.attrs.modInfo.fee.feeType === 'FIXED') {
```
rmisio: Would it be better to "reset" the non used value in save of the view, rather than here in the model?

The problem with doing it here is that your model and, consequently the UI, are out of sync with what's set on the server. So, while the server will, for example, have a percentage of 0, the global profile model is still retaining that old value.

I think it would be cleaner to make the switch in the save() method of the Moderation view, e.g:

```
 
    // change the formData as necessary before saving it in the model   
 
    this.profile.set(formData);
```

Also, rather than resetting the values to hard-coded ones, I would reset them to whatever the defaults are in the model. There are default values for both fields, but if you wanted to make it future proof, you could check for an 'undefined' default value. So... if there is a default, you set it to the default. If there isn't a default, you would both delete it from the formData object and unset() it in the model.

PR #353, opened 2017-02-08
Encrypted Updated Signing Certs

hoffmabc:
This is so Travis can build using the new signing certs for Windows binaries.


hoffmabc:
This one is good until 2020 so shouldn't have to do this again anytime soon.

hoffmabc:
Ok I think this one should be working properly now @jjeffryes let's watch for the OSX build to eventually complete.

PR #355, opened 2017-02-08
Btn spinner

rmisio:
Updating buttons which have a loading / processing state to use our SVG spinner (as opposed to the old placeholder of having the text 'processing' overlay the button).


morebrownies:
@rmisio should the spinner be working on the Follow button? When I hit follow, it doesn't do anything for a few seconds and then "Unfollow" appears.

rmisio:
@morebrownies As of now, it's not in place on the Follow button on the User Page (home tab). We'll work that in later since it requires a slight refactor since that's not using our common User Card view.

It does, however, work on User Cards in other sections of the app, e.g Connected Peers, Followers tab, etc...

PR #356, opened 2017-02-08
Add color previews to demystify hex RGB values.

### LINKED ISSUE #196: Settings: Color Options
screen shot 2016-12-13 at 9 44 39 am

Suggestion: display the colors in a shape (circle or square) under the 'primary, secondary, or text color' headings... because unless you're a developer, no one knows what these codes mean.

srhoulam:
Resolves #196

### DIFF
--- a/styles/modules/modals/_settings.scss
+++ b/styles/modules/modals/_settings.scss
@@ -96,4 +96,12 @@
       }
     }
   }
+
+  .previewBox {
+    min-width: 1em;
+    min-height: 1em;
+    width: auto;
+    height: 60%;
+    margin: 20%;
+  }
 }
--- a/styles/modules/modals/_settings.scss
+++ b/styles/modules/modals/_settings.scss
@@ -98,10 +98,9 @@
   }
 
   .previewBox {
-    min-width: 1em;
-    min-height: 1em;
-    width: auto;
-    height: 60%;
-    margin: 20%;
+    width: $formElementHeight;
+    height: $formElementHeight;
+    border-style: solid;
+    border-width: 1px;
   }
 }
--- a/js/templates/modals/settings/page.html
+++ b/js/templates/modals/settings/page.html
@@ -207,7 +207,7 @@
             <div class="clrT2 txSm"><%= ob.polyT('settings.pageTab.helperPrimaryColor') %></div>
           </div>
           <div class="col1">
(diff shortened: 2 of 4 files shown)


srhoulam:
Thanks for the feedback @rmisio. I've made the requested changes and verified them.

styles/modules/modals/_settings.scss
```diff
@@ -96,4 +96,12 @@
...
     }
   }
+
+  .previewBox {
+    min-width: 1em;
+    min-height: 1em;
+    width: auto;
+    height: 60%;
+    margin: 20%;
+  }
```
jjeffryes: Since input elements are set to have 40px line heights, and this class won't be used in any other situation, it would be best to just set this element to have the same for height and width, like this:

```
height: $formElementHeight;
width: $formElementHeight;
```

If you also remove the margin, I think it looks quite nice with the above size (see screenshot). 

You can also set the border-width and border style here, as this will always have a border, rather than add a border class to the element.

image

srhoulam:
Thanks @jjeffryes. How are the latest changes?

jjeffryes:
Very nice, @srhoulam 

Once @rmisio marks his review as accepted, and we hear from @morebrownies we can merge this.

srhoulam:
Good to hear. Feel free to let me know if anything else needs to be done.

srhoulam:
There seems to be an issue with the CI script generating certificate(s) and environment variables. Is there a commit I should rebase these changes on so it can pass that check?

jjeffryes:
@srhoulam if you rebase it to master, it should be fine. We updated the cert a while ago.

jjeffryes:
It looks like this still needs to be rebased so Travis will run correctly?

Also, we need @morebrownies to take a look.

morebrownies:
Sorry about the delay @jjeffryes, just got around to looking at this now.

I think it looks pretty good. In the future, we'll want to make the color square clickable, but we can definitely layer that in later.

Nice work @srhoulam 🍻

srhoulam:
Yes, I did rebase it to the head of master a couple times. It took care of the build script invoking a missing command/binary, but it didn't solve the openssl invocation exiting with an error code.

Thanks for the feedback, everyone.

@morebrownies I can create the issue for that improvement if it doesn't already exist.

morebrownies:
Yeah I don't think we have an issue logged for that one yet @srhoulam

PR #359, opened 2017-02-09
userData command line option

rmisio:
This change-set allows the setting of the userData path to be passed in as a command line option. This path controls to which folder localStorage data (and possibly other things) is stored as is critical if you want to run multiple clients on the same machine.

The option can be passed in as follows:

```
NODE_ENV=development ./node_modules/.bin/electron . --userData /Some/path/here
```

Keep in mind, on Windows, there is a different syntax there to set the environment variable.

Also, keep in mind with the above command you are bypassing all our dev "sugar" (watchers, sass compiler, etc...) that is otherwise present with `npm start`. Unfortunately, in order to pass a command line option to 'npm start' and then have it ultimately be passed into the 'electron' process would require some tweaks to our npm scripts. Since it's unlikely you would need to run a second client concurrently and develop on it, we're going to leave that alone for now.

### DIFF
--- a/js/start.js
+++ b/js/start.js
@@ -1,4 +1,4 @@
-import { remote, ipcRenderer, app as electronApp } from 'electron';
+import { remote, ipcRenderer } from 'electron';
 import $ from 'jquery';
 import Backbone from 'backbone';
 import Polyglot from 'node-polyglot';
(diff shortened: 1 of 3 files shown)


jjeffryes:
The way to do it on Windows is 
```
set node_env=development&&electron . --userData pathToDirectory
```

It's worth noting you have to have electron installed globally for that to work.

rmisio:
On windows, this can be done via:

```
set node_env=development
node_modules\electron\dist\electron.exe . --userData pathToDirectory
```

rmisio:
An additional point to consider is that starting the app this way will not automatically kick off the various watcher processes which compile your resources (e.g. Sass to CSS). So, if you'll be making any changes to the source when running the client this way (e.g. you pull in some updates), you'll want to run `npm run build` first before launching with the command above.

Issue #374, opened 2017-03-01
Should chat timestamps update in real-time?


rmisio:
Since we are using relative timestamps for chat... do they need to update in real (or semi-real) time?

![image](https://cloud.githubusercontent.com/assets/794517/23472778/d3171656-fe62-11e6-8dee-6b0762e65468.png)

For example, after a minute, 'a few seconds ago' would become 'a minute ago'.

rmisio:
I would think that, yes, they should update. But... I would probably mark this as a nice-to-have and worry about it post release. If we keep trying to implement every little chat detail, it's going to keep dragging on a while.

jjeffryes:
This is a tough call. If it's easy to do, we should do it. If it's hard... it could be really weird if we defer it, as you could get a bunch of messages that all say "a minute ago" even though the older ones aren't a minute old any more.

rmisio:
I don't think it's super hard to do... I just need to be mindful of performance, since there could potentially be dozens or hundreds of messages in view. I need to do a little research to find out how expensive multiple setIntervals are and also be smart about on re-rendering a message if the timestamp actually changed.

My guess is it's probably 4 hours of work.

An alternative is just using the absolute date / time, like in v1 (e.g. Mar 1st, 2017 9:33pm) which would be a few minute switch and then later on circling back and doing the relative time properly.

Just depends on how important having the relative time is right now.

jjeffryes:
Let's just use absolute time, and add an issue to go back and make it relative later.

morebrownies:
Would it be useful to mark some of these smaller tasks with "Help Wanted" @jjeffryes? This seems like a good one for the community to work on.

hoffmabc:
Great suggestion.

jjeffryes:
@morebrownies yes, that would be good.

rmisio:
This was completed.

PR #375, opened 2017-03-01
Chat

### LINKED ISSUE #354: Add Chat Functionality to the Chat Sidebar
The chat sidebar should show your conversations and allow the continuation of any past conversation. New conversations should appear in the chat sidebar.

### Chat Sidebar Conversation List
- [ ] Show a vertical stack of avatars. Each avatar represents a conversation with the other user whose avatar is shown.
- [ ] The stack is ordered by most recent message.
- [ ] There should be an indicator of how many unread messages are not visible (scrolled off screen).
    - [ ] If the conversation has unread messages, a badge with a number equal to the number of unread messages is shown on top of the avatar.
    - [ ] Clicking on the avatar or the first message to the right of the avatar, will load the conversation in the Current Conversation Section using the ob/chatmessages api (send it peerID and an empty subject, with pagination).
    - [ ] When the conversation is active, the unread message badge is removed, and the unread messages are marked as read (using the markchatasread endpoint).
    - [ ] When the conversation is active, the message to the right of the avatar is emphasized with a darker/thicker border.
- [ ] To the right of each avatar is a truncated handle or guid, and the first few words of the last message from that user (the actual last message may be from you). The message is truncated. Truncation can use ellipses.
    - [ ] If the last message is unread, it should be bold or otherwise emphasized.
- [ ] When the chat bar is closed, the messages should not be visible.
- [ ] As new messages come in through the web socket, the list is updated.

### Chat Sidebar Current Conversation Section
- [ ] When a new conversation is started, or an old conversation is active, the Current Conversation Section is shown.
- [ ] The avatar and handle or guid of the person being chatted with is shown.
    - [ ] The avatar and handle/guid should be clickable, and will load that user's user page.
- [ ] The messages in the conversation are shown, with the sender's avatar next to each message. The user's own messages should be aligned to the right, the other participant's to the left.
    - [ ] Each message shows how long ago it was sent, based on its timestamp.
        - if possible, messages that are less than 1 minute apart should have a single timestamp for the entire group, below the group.
    - [ ] If a handle with an @ is in a message, it should automatically be formatted as a clickable link to that handle
    - [ ] If a link (http, https, ftp) is in a message, it should automatically be formatted as a clickable external link.
    - [ ] If a ob link is in a message, it should automatically be formatted as a clickable internal link.
    - [ ] The avatar and handle/guid should be clickable, and will load that user's user page.
    - [ ] New messages will come in through a web socket with a peerID (guid), and should immediately be shown in the current conversation. 
    - [ ] An api call to markchatasread should be sent immediately when each message appears n the Current Conversation Section.
- [ ] Messages scroll if they don't fit in the Current Conversation Section.
- [ ] The messages are fetched in batches, and old messages will lazy load if the user scrolls up.
- [ ] There is a message input field. It should be limited to 20,000 characters.
    - [ ] When the user starts typing, a message should be sent with a blank message value. This will trigger a "typing" indicator for the other user. This should be on a timer, so it shows for X seconds, and at the end of that time another message is sent if the other user is still typing.
- [ ] There is an emoji button next to the message input field. It will show an emoji selection box when clicked, allowing the user to insert an emoji into their conversation.
- [ ] There is a close button, it will close the Current Conversation Section. 
- [ ] There is a more menu, with the option to delete the conversation. 

### User Page Message Button
- [ ] When the message button on the user page is clicked, it should start a new conversation with that user.

## Data
/ob/chatconversations - get all chat conversations as an array, with guid and unread count
/ob/chatconversation - delete one conversation
/ob/chatmessages - get messages for one conversation
/ob/chatmessage - delete one message
/ob/chat - POST the message
/ob/markchatasread

This is the format of messages on the websocket that are relevant to chat:

### Message:
```
{
    "message": {
        "messageId": "KwLwyVgEjP+HHA==",
        "peerId": "QmfCXCN4x7S8A2Ka6ZQisF3kU5bPH8vzV44odLo4mZccaw",
        "subject": "",
        "message": "Hello World!",
        "timestamp": "2017-02-08T15:47:25.562600073-05:00"
    }
}
```

### Message has been read:
```
{
    "messageRead": {
        "messageId": "FHi6CCAjxE59OQ==",
        "peerId": "QmfCXCN4x7S8A2Ka6ZQisF3kU5bPH8vzV44odLo4mZccaw",
        "subject": ""
    }
}
```

### Other user is typing:
```
{
    "messageTyping": {
        "peerId": "QmfCXCN4x7S8A2Ka6ZQisF3kU5bPH8vzV44odLo4mZccaw",
        "subject": ""
    }
}
```

## ToDos
- add a delete button to the conversation to permanently delete the conversation
- change chatconversations api to order the data by latest message (timestamp)
- change chatconversations to include the last message
- design the offscreen unread messages indicators
- design a typing indicator

## ToDo Later (post 2.0)
- design a way to delete a specific message in the chat (locally)
- design a way to delete the entire conversation locally
- consider a way to access an expanded view for the conversation

### LINKED ISSUE #384: Chat typing handling
The typing pop up should probably use the handle instead of the guid if one is available.

### LINKED ISSUE #386: When chat is closed, clicking on a chat head should both open chat and the convo window for the clicked head.

rmisio:
closes #354 
closes #386 
closes #384 

This doesn't include Emoji support. I'm starting on that now, so you'll still see commits coming into here, but wanted to have the review on this PR start, since it's a decent size feature.

Also @jjeffryes, some CSS tweaks will be needed to account for the close button. I'm not sure the best (least painful) way to approach those. Basically, the close button has different states depending on what's behind it:

1.) Overlay is behind it.
image

2.) Page with a tab bar behind it:
image

2.) Page without a tab bar behind it (e.g. Connected Peers page). The Connected Peers is a dummy page. You may want to check the designs / @morebrownies to find out if there are any real pages without tab bars that we need to account for:
image

Ideally, the chat wouldn't have to be aware of what's behind it... but unless we come up with something clever, that might just have to be the case. We already have a root level class that tells us if a model is open, which would account for 1.

For 2 and 3, since the presence of a tab bar is determined by individual changes, each page would need to manage a top level class, which is not ideal, but if there's no other way, I suppose we could work a flag into the base view or create a pageView which all pages descend from.

Anyhow... good luck!

### DIFF
--- a/js/utils/serverConnect.js
+++ b/js/utils/serverConnect.js
@@ -54,19 +54,27 @@ export function getDebugLog() {
  * Returns an object with data about the current connection, incuding
  * the status (e.g. 'connecting', 'connected'...) and any relevant data
  * (e.g. the Socket instance, the Server Configuration model...).
- *
- * If you're looking to bind to Socket events, you would call this method
- * and then bind to the Socket instance that's returned. Keep in mind that
- * until the first connection is made, this method will return null, in which
- * case you'd wanted to wait until the 'connected' event to do your binding.
- *
- * But, since the app won't proceed out of Start.js, until a connection is
- * made, it's really only in Start.js that you need to wait for the event
- * and beyond Start.js you should just be able to get the socket via this
- * method.
  */
 export function getCurrentConnection() {
   return currentConnection;
+}
+
+/**
+ * Call this method to obtain the socket instance in order to bind socket events.
+ * If we are not currently connected to a server, this method will return false.
+ * For the most part, you could no-op in that case since as it is now the
+ * Connection Modal will overlay the app when there is no connection and any
+ * subsequent reconnection will result in the app re-starting.
+ */
+export function getSocket() {
+  const serverConnection = getCurrentConnection();
+  let returnVal = false;
+
+  if (serverConnection && serverConnection.status !== 'disconnected') {
+    returnVal = serverConnection.socket;
+  }
+
+  return returnVal;
 }
 
 function socketConnect(socket) {
--- a/js/views/chat/Chat.js
+++ b/js/views/chat/Chat.js
@@ -201,6 +201,19 @@ export default class extends baseVw {
   _openConversation(guid, profilePromise) {
     this.open();
 
+    if (this.chatHeads) {
+      this.clearActiveChatHead();
+
+      // If there's a chat head, mark it as active
+      const chatHeadModel = this.collection.get(guid);
+
+      if (chatHeadModel) {
+        this.chatHeads.views[this.collection.indexOf(chatHeadModel)]
+          .$el
+          .addClass('active');
+      }
+    }
+
     if (this.conversation && this.conversation.guid === guid) {
       // In order for the chat head unread count to update properly, be sure to
       // open before marking convo as read.
@@ -278,7 +291,15 @@ export default class extends baseVw {
   }
 
   closeConversation() {
+    this.clearActiveChatHead();
     if (this.conversation) this.conversation.close();
+  }
+
+  clearActiveChatHead() {
+    if (this.chatHeads) {
+      this.chatHeads.views
+        .forEach(chatHead => chatHead.$el.removeClass('active'));
+    }
   }
 
   open() {
--- a/js/views/chat/Chat.js
+++ b/js/views/chat/Chat.js
@@ -9,7 +9,6 @@ import Profile from '../../models/profile/Profile';
 import baseVw from '../baseVw';
 import ChatHeads from './ChatHeads';
 import Conversation from './Conversation';
-import moment from 'moment';
 
 export default class extends baseVw {
   constructor(options = {}) {
@@ -118,6 +117,17 @@ export default class extends baseVw {
 
     this.listenTo(this.conversation, 'convoMarkedAsRead',
       () => this.onConvoMarkedAsRead(guid));
+
+    this.listenTo(this.conversation, 'deleting',
+      (e) => {
+        e.request.done(() => {
+          this.collection.remove(e.guid);
+
+          if (this.conversation && this.conversation.guid === e.guid) {
+            this.conversation.close();
+          }
+        });
+      });
 
     this.$chatConvoContainer
       .append(this.conversation.render().el);
@@ -328,6 +338,7 @@ export default class extends baseVw {
 
       this.listenTo(this.chatHeads, 'chatHeadClick', this.onChatHeadClick);
       this.listenTo(this.chatHeads, 'rendered', this.onChatHeadsRendered);
+      this.listenTo(this.chatHeads, 'click')
 
       // It is important that both of the following occur before the chatHeads
       // view is rendered:
--- a/js/views/chat/Chat.js
+++ b/js/views/chat/Chat.js
@@ -1,0 +1,67 @@
+// import $ from 'jquery';
+// import _ from 'underscore';
+import loadTemplate from '../../utils/loadTemplate';
+import baseVw from '../baseVw';
+
+export default class extends baseVw {
+  constructor(options = {}) {
+    if (!options.collection) {
+      throw new Error('Please provide a chat conversations collection.');
+    }
+
+    super(options);
+
+    // this._state = {
+    //   status: 'not-connected',
+    //   ...options.initialState || {},
+    // };
+  }
+
+  className() {
+    return 'chat';
+  }
+
+  events() {
+    return {
+      // 'click .js-btnConnect': 'onConnectClick',
+    };
+  }
+
+  // getState() {
+  //   return this._state;
+  // }
+
+  // setState(state, replace = false) {
+  //   let newState;
+
+  //   if (replace) {
+  //     this._state = {};
+  //   } else {
+  //     newState = _.extend({}, this._state, state);
+  //   }
+
+  //   if (!_.isEqual(this._state, newState)) {
+  //     this._state = newState;
+  //     this.render();
+  //   }
+
+  //   return this;
+  // }
+
+  // remove() {
+  //   super.remove();
+  // }
+
+  render() {
+    loadTemplate('chat/chat.html', (t) => {
+      this.$el.html(t({
+        // ...this.model.toJSON(),
+        // ...this._state,
+      }));
+
+      // this._$deleteConfirm = null;
+    });
+
+    return this;
+  }
+}
--- a/js/views/chat/Chat.js
+++ b/js/views/chat/Chat.js
@@ -1,7 +1,9 @@
-// import $ from 'jquery';
+import $ from 'jquery';
 // import _ from 'underscore';
+import { getBody } from '../../utils/selectors';
 import loadTemplate from '../../utils/loadTemplate';
 import baseVw from '../baseVw';
+import ChatHeads from './ChatHeads';
 
 export default class extends baseVw {
   constructor(options = {}) {
@@ -15,6 +17,14 @@ export default class extends baseVw {
     //   status: 'not-connected',
     //   ...options.initialState || {},
     // };
+
+    this._isOpen = false;
+
+    this.listenTo(this.collection, 'sync', () => {
+      console.log('boom selecta');
+      window.boom = this.collection;
+      this.render();
+    });
   }
 
   className() {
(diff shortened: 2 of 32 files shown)


jjeffryes:
@rmisio I'll take a look at the CSS as soon as I can.

js/utils/serverConnect.js
```diff
@@ -54,21 +54,29 @@ export function getDebugLog() {
...
 }
 
+/**
+ * Call this method to obtain the socket instance in order to bind socket events.
+ * If we are not currently connected to a server, this method will return false.
+ * For the most part, you could no-op in that case since as it is now the
+ * Connection Modal will overlay the app when there is no connection and any
+ * subsequent reconnection will result in the app re-starting.
+ */
+export function getSocket() {
```
rmisio: Added this as a convenience method to get the socket instance. Should be easier than having to get it from getServerConnection().

You could use it as so:

```
const socket = this.getSocket();

if (socket) {
  this.listenTo(socket, 'message', () => {});
}
```

js/views/chat/Chat.js
```diff
@@ -0,0 +1,457 @@
...
+    // determine if a certain portion of the el is out of view rather than the
+    // whole element
+
+    if (firstUnreadChatHead) {
+      const firstUnreadIndex = this.collection.indexOf(firstUnreadChatHead);
+
+      if (!isScrolledIntoView(this.chatHeads.views[firstUnreadIndex].el)) {
+        this.$el.addClass('outOfViewUnreadsAbove');
+      } else {
+        this.$el.removeClass('outOfViewUnreadsAbove');
```
rmisio: I need to know whether firstUnreadChatHead is true to determine even if I want to set the firstUnreadIndex variable. I could skip setting firstUnreadIndex and do:

```
this.$el.toggleClass('outOfViewUnreadsAbove, !!firstUnreadChatHead && !isScrolledIntoView(this.chatHeads.views[this.collection.indexOf(firstUnreadChatHead)].el));
```

But I think this is a case where a few extra lines is worth making things more readable.

hoffmabc:
I'm not sure if this is defined in our requirements (a problem in itself), but when you click Message on a user profile it should then lead to an on focus of the chat message window. Right now this PR just pops open the window but doesn't focus on it.

rmisio:
> when you click Message on a user profile it should then lead to an on focus of the chat message window

@hoffmabc Ok, done.

morebrownies:
@rmisio I noticed the conversation list and chat window are different widths. Is it possible for the conversation list to expand to match the list of the chat window? I know there was a potential issue with this, but I can't remember what it was.

screen shot 2017-03-07 at 12 55 21 pm

rmisio:
@morebrownies The problem is that the convo list would overlap most modals.

morebrownies:
Doesn't that mean the chat window would too then at its current width?

morebrownies:
We may need to reduce the width on the chat window to prevent it from covering modals?

rmisio:
Up to you. Things would be kind tight. I'm wondering if we should maybe live with the overlap on the convo?

image

FWIW, the fact that the convo width doesn't line up with the convo list width doesn't bother me at all... but maybe my design eye isn't as discerning ;)

Anyhow, your call.

jjeffryes:
I noticed it too, it would be good if they were the same.

For the issues with the close button, it's hard sometimes to tell if the close button is associated with the chat, one simple solution (which I tried briefly in code, it should be easy to do) is to give the chat a solid background when it's open, and make the close button connect to the top of it like a tab. But that makes the width difference really noticeable.

morebrownies:
I'm starting to lean towards a background color when the chat panel is opened too @jjeffryes. I agree the close button does get a bit lost in the interface.

Here's a comp of what it would look like with a background color and the width of the conversation list and chat window aligned:

screen shot 2017-03-08 at 9 43 24 am

Thoughts??

rmisio:
@morebrownies FWIW, That design is not really representative of the width available to us when modals are open in the background. I think you're roughly looking at a 20% decrease.

You're looking at roughly 225px (imagine a scroll-bar there):

image

It will take me a bit of finicky little css work to shrink things down... so I just want to make sure you are fine with things getting more cramped (particularly in the convo window)?

morebrownies:
Is it simple to keep the user page aligned when the chat panel is opened? The navigation currently stays fixed, while the body container slides left. See screenshot

align

morebrownies:
Let me double check to make sure that width is usable @rmisio

morebrownies:
.msgContentBox needs a max-width and word wrap. Pasting a link results in content getting hidden:

screen shot 2017-03-08 at 10 15 11 am

jjeffryes:
Messing around with it while fixing that missing border, it turned out to be pretty easy to add a solid border with a nice transition when you open the chat.

The width not matching the conversation box is the only sticking point.

morebrownies:
@rmisio I just tested #chatContainer at 225px width (on a 13" screen and 30" screen) and it feels very useable to me. I think we should proceed with the change. 

screen shot 2017-03-08 at 10 24 01 am

morebrownies:
We talked about showing "3 minutes ago" under messages instead of the basic timestamp. @rmisio did moment.js not give us what we needed out of the box to achieve this? I'm mostly just curious. If it was a lot of additional work it is definitely worth punting on.

screen shot 2017-03-08 at 10 26 21 am

jjeffryes:
Rob can explain about the timestamp, but it turned out to be a lot of extra work.

hoffmabc:
Having a solid color behind the chat heads to me seems not very nice. Are we sure we want to change it to that?

morebrownies:
Can you elaborate on that @hoffmabc? Would like to understand better why you're not feeling it. 

I feel the background color could be helpful when the conversation list starts to compete with the user's hero image:

**No background color**
screen shot 2017-03-08 at 10 45 25 am

**w/ background color**
screen shot 2017-03-08 at 10 46 53 am

hoffmabc:
I'm much more of a fan of the transparency look than the stark blocky solid color. Even something more like a slight transparency seems like it would be better than just a stark white block.

image

morebrownies:
That's cool. I like that approach too.

hoffmabc:
You guys agree?

http://i.giphy.com/l3q2K5jinAlChoCLS.gif

morebrownies:
Seems like a good middle ground to me 👍

rmisio:
> You guys agree?

BHoff busting out the UX design mad skillz. You best recognize!

rmisio:
> did moment.js not give us what we needed out of the box to achieve this? I'm mostly just curious. If it was a lot of additional work it is definitely worth punting on.

The relative timestamp is really easy to do. It's the fact that it needs to be updated regularly to make any sense that added some complication (`a few seconds ago` needs to change to `a minute ago` after a minute, then to `2 minutes ago` after another minute, eventually to `an hour ago`, etc... otherwise as you're chatting, you're just going to have a bunch of messages perpetually showing `a few seconds ago`). So, we punted on that for now. (#374).

@wolf ^

morebrownies:
ok thanks @rmisio.

jjeffryes:
Something like this?

image

jjeffryes:
Neutral background with conversation open: 

image

morebrownies:
Yeah I think that should work @jjeffryes. We may need to fine tune it a bit once we get customize page in place.

jjeffryes:
It's set to be a transparent version of the primary color, and should automatically adjust when custom page colors is coded.

jjeffryes:
I should say, I coded that in the patch PR I did, which I need to update. ;)

morebrownies:
We're going to need a "ding" sound to trigger when a chat message is received. The sound we have in 1.0 should be good.

rmisio:
> We're going to need a "ding" sound to trigger when a chat message is received. The sound we have in 1.0 should be good.

@morebrownies Could we please get an issue for that so we could prioritize and track it?

jjeffryes:
@rmisio and @morebrownies I put in an issue to track the sounds we'll need. We should do something more unified than we did in 1.0, I think. #387

rmisio:
FWIW, I think that chat background looks quite crappy. It looks OK on a very colorful photo, like Brian's, but on something more subtle like Wolf's ski shop photo or on our default photo or when the modal overlay is in the background... I think it looks pretty bad.

Just my two cents... I'll live with it, but IMHO, no BG at all looked much better.

image

rmisio:
@jjeffryes Do you want to respond and / or handle this one: https://github.com/OpenBazaar/openbazaar-desktop/pull/375#issuecomment-285117588

jjeffryes:
I can take a look at sliding the tabs too. It should be possible to put in a universal fix for all tabs.

morebrownies:
Do you feel the legibility is good enough without a background @rmisio? On some pages, I found it  difficult to focus on text when there is an overly distractive cover photo in the background (see screenshot).

screen shot 2017-03-08 at 6 07 14 pm

PR #376, opened 2017-03-01
Adding some bottom padding to the connected peers page

rmisio:
Adds some bottom padding to the Connected Peers page so the Load More button isn't obscured by the Testing badge.

### DIFF
--- a/styles/modules/_userPage.scss
+++ b/styles/modules/_userPage.scss
@@ -155,4 +155,8 @@
       }
     }
   }
+
+  &.connectedPeersPage {
+    padding-bottom: 70px;
+  }
 }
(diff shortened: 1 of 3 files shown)


styles/modules/_userPage.scss
```diff
@@ -155,4 +155,8 @@
       }
     }
   }
+
+  &.connectedPeersPage {
```
jjeffryes: Since we're going to get rid of the connected peers page, would it be better to just switch the floR class on the button to floL?

Then it will be visible and we won't need to remember to remove this style rule later.

rmisio:
ok, @jjeffryes removed the float so the button now appears on the left.

PR #377, opened 2017-03-02
Store tab

### LINKED ISSUE #255: Add Store Tab to Settings Modal
image

The Store tab sets data used by vendors.

## Functional Requirements
- [x] A control to set the store to on or off. This control shows the current state of this setting. 
- [ ] A list of moderators that have currently been selected for the store. 
    - The design of the selected list is assumed to be similar to the browsing list (TBD)
    - Each moderator in the list has a button to remove them from the list.
- [x] A button to browse available moderators.
- [ ] When clicked, a list of available moderators is shown.
- [ ] Each moderator has the following available interactions:
    - [ ] A select button. When clicked, a popup/tooltip is shown with the terms of service, a cancel button and an agree button. Clicking the cancel button closes the tooltip, clicking agree the moderator is marked as selected. 
    - [ ] A button to add a moderator by handle or ID. When clicked, it will show a moderator card, with the option to select that moderator (identical to a moderator in the browse list). If the handle/guid profile comes back as not a moderator, a message should be shown that they are not a moderator.
- [ ] A save button. Clicking this button will save the current changes. 

## Data
- moderators, an array of guids, from the Settings API (stored in the settings model).
- the list of available moderators the get moderators API.
- vendor, a boolean, from the Profile API (stored in the profile model).
- each moderator loaded will need to make a call to IPFS to get it's profile data

## TODOs
- update the designs to indicate the store is off when the user is looking at their own listings
- update the designs to show what happens when a moderator is added by handle/guid.
- on backend, store the listings in a non-public folder when the store is deactivated, so the user can still look at and modify their own listings.

jjeffryes:
This adds the Store Tab to the Settings modal.

A Moderators collection was created, which accepts several parameters on creation.
apiPath = an alternate api to use for the URL, used in this case to switch from moderators to fetchprofile
async = can be set to true to fetch profiles asynchronously
include = sets the query parameter for include, in this case to "profile" 
excludeCollection = can accept another Modertors collection to exclude. Used to prevent the available moderators collection from adding moderators already in the selected moderators collection

On fetch, a list of guids can be given as a fetchList parameter, that will switch the fetch to POST and it will send the list to the server, expecting the server to send back a set of profiles. This also sets the notFetchedYet property of the collection, which is used to tell when all the expected moderators have been retrieved.

Add has been overridden to make sure profiles returned are valid moderators (their moderator attribute is true, and they have a modInfo attribute), and they are not on the excluded list. As moderators are added they are removed from the notFetchedYet list.

Parse has been overridden to handle asynchronus loading. If the collection is set to be async, the fetch will receive an id for the websocket instead of profiles, and will listen for messages on the websocket with the same id, adding the socket profiles as they arrive.

I wasn't able to write tests for the new Moderators collection because in my environment it throws an error when a Moderators collection is created by the test:
```
C:\Users\J\ob-j\openbazaar-desktop\node_modules\backbone.localstorage\backbone.localStorage.js:58
Backbone.LocalStorage = window.Store = function(name, serializer) {
                        ^
ReferenceError: window is not defined
```
This seems like it could be related to same issue that prevents jsDOM from working in tests.

The Store view handles loading the various collections, and creating the Moderator child views. Each of the list areas has a processing state (added to the containers file, it's similar to the one used on buttons), for the selected moderators processing lasts until all of them are loaded. For the available moderators processing ends when the first one is loaded, or 15 seconds elapses with no moderators returned.

Selecting and de-selecting moderators works through events fired on the Moderator or ModeratorDetails views, which the Store tab listens for. Since selected or de-selected moderators can be in any of the three lists, the Store view keeps an array of the current list of moderator guids to save, and updates it as needed. 

Closes #255


js/views/modals/Settings/Store.js
```diff
@@ -0,0 +1,325 @@
...
+      this._$modListByID = null;
+      this._$modListAvailable = null;
+      this._$browseMods = null;
+      this._$submitModByIDInput = null;
+      this._$submitModByID = null;
+      this._$submitModByIDInputError = null;
+      this._$submitModByIDInputErrorText = null;
+    });
+
+    if (this.currentMods.length) {
```
rmisio: It also seems weird to me that when I click save, the available mod list reverts to empty with a browse button, forcing me to re-browse. I would think we could just leave in place what's there with maybe a refresh button.

js/views/modals/moderatorDetails.js
```diff
@@ -0,0 +1,83 @@
...
+export default class extends BaseModal {
+  constructor(options = {}) {
+    const opts = {
+      removeOnClose: true,
+      ...options,
+    };
+
+    super(opts);
+    this.options = opts;
+    this.model = this.options.model;
```
rmisio: > this.model = this.options.model;

Backbone does this by default.

rmisio: Can we ensure that the model is provided in the constructor? The view will bomb in places if it's not provided.

js/views/moderatorCard.js
```diff
@@ -0,0 +1,64 @@
...
+      this.listenTo(modModal, 'addAsModerator', () => {
+        this.changeSelectState('selected');
+        this.trigger('changeModerator', { selected: true, guid: this.model.id });
+      });
+    } else if (this.cardState === 'selected') {
+      this.changeSelectState('deselected');
+      this.trigger('changeModerator', { selected: false, guid: this.model.id });
+    } else if (this.cardState === 'deselected') {
+      this.changeSelectState('selected');
+      this.trigger('changeModerator', { selected: true, guid: this.model.id });
```
rmisio: For example, there's no reason we can't be passing the full model along in these cases and then other views don't need to re-fetch it.

js/views/modals/Settings/Store.js
```diff
@@ -0,0 +1,325 @@
...
+      this.$submitModByIDInputError.removeClass('hide');
+      this.$submitModByIDInputErrorText.text(blankError);
+    }
+  }
+
+  processIDorHandle(modID) {
+    if (modID.charAt(0) === '@') {
+      // if the id is a handle, get the guid
+      const handle = modID.slice(1);
+      getGuid(handle)
```
rmisio: There needs to be some cleanup on any async calls like this one (or `this.modsSelected.fetch({ fetchList: this.currentMods });`).

1.) In remove of this view, the calls should be aborted.
2.) In the handlers, the first line should be something to the effect of `if (this.isRemoved) return;`).

Otherwise, you'll have the situation where the user navigates away from this view and the handlers of those calls still execute down the line. At the very least, their keeping this view from being garbage collected. On a worse level, they may trigger Dialogs that are no longer applicable.

js/views/modals/moderatorDetails.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  messageClick() {
+    // TODO: wire this in after chat is working
+    console.log('Message button clicked');
+  }
+
+  followClick() {
+    this.$followBtn.addClass('processing');
+    followUnfollow(this.model.id, 'follow')
```
rmisio: Same as the previous comment... asyncs need to be cleaned up in case this view has been removed.

js/views/modals/moderatorDetails.js
```diff
@@ -0,0 +1,83 @@
...
+
+  unfollowClick() {
+    this.$unFollowBtn.addClass('processing');
+    followUnfollow(this.model.id, 'unfollow')
+        .always(() => (this.$unFollowBtn.removeClass('processing')));
+  }
+
+  addAsModerator() {
+    this.trigger('addAsModerator');
+    super.close();
```
rmisio: I would call `this.close`, so if this method ever overrides close, we won't be bypassing it. If it's not implemented (like now), then the parent will be called anyways.

rmisio: In this view `super.close` would only make sense in an overridden close method -or- in a rare case where you overrode close, but still want some part of this view to use the parents instead.

js/collections/Followers.js
```diff
@@ -7,18 +7,20 @@ module.exports = Collection.extend({
   /* we have to use the older style for this collection, the ES6 style creates a bug where models
```
jjeffryes: None of these changes to followers should have been in the PR, that was an accidental inclusion of a half-done refactor. I'll take it out, it should be done as a separate PR.

jjeffryes:
@rmisio I made a lot of changes to this PR, both to address issues you noticed in the code review, and to take care of various other things I noticed while I was doing that.

js/collections/Moderators.js
```diff
@@ -36,15 +45,17 @@ export default class extends Collection {
...
     // remove any returned profiles that are not valid moderators
     if (models) {
       filteredModels = filteredModels.filter((mod) => {
-        // don't add excluded ids
-        const notExcluded = excludeList.indexOf(mod.id) === -1;
+        // don't add excluded ids or your own id
+        const notExcluded = excludeList.indexOf(mod.id) === -1 && mod.id !== app.profile.id;
         // don't add if not a mod or the mod data is missing
-        return mod.moderator && mod.modInfo && notExcluded;
+        return mod.get('moderator') && mod.get('modInfo') && notExcluded;
```
rmisio: The profile model has a 'isModerator' getter which would handle the first two conditions.

js/views/moderatorCard.js
```diff
@@ -7,8 +7,12 @@ import { launchModeratorDetailsModal } from '../utils/modalManager';
 
 export default class extends BaseVw {
   constructor(options = {}) {
-    super(options);
-    this.options = options;
+    const opts = {
```
rmisio: Did you have some default options and then removed them here?

As it is now, this assignment is pointless since it results in the same thing, so you could have left the code as it was before.

rmisio:
There's a couple of errors I'm noticing:

1.) If I forget to preface a handle with '@' and type, for example, 'chrispacia', clicking submit doesn't seem to do anything.

2.) After selecting an available mod and saving, there's an exception which is causing the view to hang:

image

jjeffryes:
Clicking an available mod and saving works for me, I'll have to do some experimenting and see if I can reproduce it. 

Good catch on the @, I didn't consider that.

rmisio:
> Good catch on the @, I didn't consider that.

You could check if it's a multihash and consider it to be a guid if so:

```
import multihashes from 'multihashes';

let isMultiHash = true;

try {
  multihashes.validate(multihashes.fromB58String(term));
} catch (exc) {
  isMultiHash = false;
}
```

That's how the address bar is determining if it's a GUID or not. Sounds like isMultiHash() would make a really good utility function.

jjeffryes:
@rmisio  This is ready for review again.

@morebrownies I added something not in the design. When you browse the available moderators, after at least on moderator is added to the list of available ones, the browse button disappears. But the user may want to be able to browse some more (it's possible more will come on line while the tab is open). 

So I added a link that appears next to the headline over the available list that allows the user to browse again. It only appears when the first browse button is not visible.

morebrownies:
Ok @jjeffryes. I'll start testing this PR out this afternoon.

morebrownies:
looking good @jjeffryes! 

I found a few things that may need a bit of clean up:

1) Searching by GUID/handle seems a bit flaky. I entered a GUID, hit submit, and was stuck on spinner for a long time....
screen shot 2017-03-10 at 4 24 41 pm
I got impatient and proceeded to hit "Save" at the bottom of the page, and I noticed the spinner in the Search by GUID/handle was still spinning, even after the page successfully saved. 

2) When I search for Moderator by GUID/handle, I noticed the value in the .js-submitModByIDInput input gets cleared out. I would prefer for the input value to stay populated, especially since it can take a while to locate them.  On that note, we also may need to provide a Cancel button when the process is running. 

3) Small thing... I like the "Browse Moderators" link you added, but I feel the copy should read "Reload List" or "Refresh List". 
screen shot 2017-03-10 at 4 32 27 pm

4) If I have moderators selected in the "Available Moderators on the Network" section, and then click the "Browser Moderators" link, I lose the moderators I selected (e.g. the green check goes away and I have to re-select them).
screen shot 2017-03-10 at 4 34 28 pm

I'm going to keep hammering on it a bit, but so far, that's all I've found 👍

jjeffryes:
@rmisio this has been updated (a lot) and the issues you found should be corrected, along with a variety of other things I found.

js/models/profile/Profile.js
```diff
@@ -113,6 +113,11 @@ export default class extends BaseModel {
       addError('website', app.polyglot.t('profileModelErrors.provideValidURL'));
     }
 
+    if (typeof attrs.vendor !== 'boolean') {
+      // this error should never be visible to the user
+      addError('vendor', `The vendor is invalid: ${attrs.vendor}`);
```
rmisio: It would be nice if the error message informed you why it's invalid... something like 'Please provide a vendor as a boolean".

js/templates/modals/settings/store.html
```diff
@@ -0,0 +1,119 @@
...
+                       <%= ob.polyT('settings.storeTab.browseModerators') %>
+                    </button>
+                  </div>
+                  <button class="btnTxt tx6 clrT2 txU browseMore js-browseMods">
+                    <%= ob.polyT('settings.storeTab.browseMoreModerators') %>
+                  </button>
+                </div>
+              </div>
+            </div><% // end of flex inside col9 %>
+            <% if (ob.errors.storeModerators) print('<div class="row"></div>  ' + ob.formErrorTmpl({ errors: ob.errors.storeModerators })) %>
```
rmisio: No need to change this, but as an FYI you can pass a class (or classes) into the error template in case you want to give it a custom style (like some margins):

```
ob.formErrorTmpl({
  errors: ob.errors.storeModerators,
  class: 'class1 class2',
}))
```

Which would result in this:

```
<ul class="unstyled errorList class1 class2">
  <li><i class="ion-alert-circled"></i>You must provide the salamander.</li>
</ul>
```

js/views/modals/moderatorDetails.js
```diff
@@ -0,0 +1,102 @@
...
+    const followAsync = followUnfollow(this.model.id, 'follow')
+        .always(() => {
+          if (this.isRemoved()) return;
+          this.$followBtn.removeClass('processing');
+        });
+    this.asyncCalls.push(followAsync);
+  }
+
+  unfollowClick() {
+    const unfollowAsync = this.$unFollowBtn.addClass('processing');
```
rmisio: Looks like you're assigning the async variable to the wrong line. I think it should go to the one below.

js/views/userCard.js
```diff
@@ -4,19 +4,38 @@ import loadTemplate from '../utils/loadTemplate';
...
 export default class extends BaseVw {
   constructor(options = {}) {
     super(options);
     this.options = options;
-    this.guid = options.guid || this.model.get('guid');
+
+    if (this.model instanceof Profile) {
+      this.guid = this.model.id;
+      this.fetched = true;
+    } else if (this.model) {
```
rmisio: What type of model would it be, if not profile?

js/views/userCard.js
```diff
@@ -4,19 +4,38 @@ import loadTemplate from '../utils/loadTemplate';
...
+      this.guid = this.model.get('guid');
+      this.fetched = false;
+    } else {
+      this.model = new UserCard({ guid: options.guid });
```
rmisio: Isn't it a profile that we want to fetch here? The UserCard doesn't have a url function implemented, so fetching it will just be a failed call. Also, the profile model has all the functionality (parse, etc...) to handle the profile data (which I believe is what we want here).

rmisio: Oh, I see you are creating the profile model below in loadUser. What is the point of the UserCard model here?

rmisio: Since we're expecting the guid to have been passed in as an option in this block, can we please check that happened and give a descriptive error message if not? Otherwise, the profile fetch will fail because it will be trying to call `ob/profile/undefined` and it's harder for a dev to understand from that that they need to pass in a guid to this view.

js/views/modals/moderatorDetails.js
jjeffryes: Weird. Copy paste error maybe.

js/views/userCard.js
jjeffryes: When the follwers collection creates userCards, the model is userCard.

rmisio:
When I search for a mod by handle and enter a handle that doesn't correspond to a real user, the last part of this message is very confusing to me.

image

Why would they "still be saved as one of (my) moderators"?

rmisio:
When I search for the guid of a valid moderator it doesn't end up putting the moderator in the Selected Moderators list even though I see that it is fetching that profile via 'fetchprofiles' and the profile does come back from the socket.

I searched for 'QmfQ2wdMyEowdQd4ezwk2QSGuVFFh7WwCe2atYaWnk3jJs'.

rmisio:
Also, Browse Moderators does not result in any moderator showing up in the list, even though the ensuing socket returns the profile of 'QmfQ2wdMyEowdQd4ezwk2QSGuVFFh7WwCe2atYaWnk3jJs'.

rmisio:
Would you mind fixing the merge conflicts with master?

js/views/userCard.js
jjeffryes: When userCard views are created by the followers collection, the model on construction is userCard, which triggers the creation and fetching of the profile model. If only a guid option was passed in, this keeps that structure. 

We could probably refactor all of it to always use profile models instead, but that might be better as a separate task.

rmisio: Since all the userCard model does is give a guid, I think we could simplify the userCard view to either accept a guid or a profile model. If the creator of the userCard view has a userCard model, then it could just pass in the guid.

jjeffryes:
I tested this just now, both searching and browsing are working correctly. 

Moderators aren't added by the browse button if they already exist in the selected list, or they aren't a valid moderator, maybe the one you saw returned met one of those criteria?

Good catch on the error, I had a todo on my list to split that error into two different use cases and forgot to. Searched moderators shouldn't use the same error message as invalid moderators on your selected list.

rmisio:
@morebrownies Were you still going to check this PR out or are you all good?

jjeffryes:
@hoffmabc can you test this on Mac and see if you can replicate the issues Rob saw? I wasn't able to replicate them on Mac or Windows.

Also, Travis is bombing again on openSSL.

hoffmabc:
I'm testing and will report as soon as possible. 

In the meantime does it seem weird that we have "Submit" here and not something like "Find"? Are we submitting the moderator if there are search results that come back?

hoffmabc:
Ok so:

1) Enter 'QmfQ2wdMyEowdQd4ezwk2QSGuVFFh7WwCe2atYaWnk3jJs' into Look up a moderator... input and I find the moderator.
2) View him and then Add as Moderator
3) Returns me to store and I now have a checkbox.
image
4) Hit Browse Moderators button shows this:
image
5) I can check/uncheck him in the bottom box now.
6) Save adds him to the top
image

Does that help @jjeffryes ?

hoffmabc:
The OpenSSL will always bomb on branches that aren't on the main repo. Sorry! It's due to the fact that Travis won't use those encrypted variables on other repos.

jjeffryes:
@hoffmabc thanks, it seems to be working right for you.

I fixed that issue on the home tab you noticed, I missed updating a css class.

rmisio:
Ok... it still doesn't work for me, but if it's not reproducible by everybody, we could leave it be for now and if others find it, we could log a bug.

An unrelated nice-to-have would be to be able to submit a handle / guid via the enter key.

PR #378, opened 2017-03-03
New CSS to match the designs, with the color of the chat close button…

jjeffryes:
…. Also makes improvements to the test notice, and prevents the chat from covering the scrollbar.

Note that this does not match the designs when the close button is over a background with no tab bar, in that case the button is the color of the background instead of the primary color. But after reviewing the designs, I can't think of a situation in which that would actually happen, there should always be a tab bar present.


index.html
```diff
@@ -1,6 +1,6 @@
 <!-- This file will be run from ./.tmp/, so make paths relative to there -->
 <!DOCTYPE html>
-<html id="ov1">
+<html id="ov1" class="pre-release"><!-- remove pre-release class when released -->
```
jjeffryes: I've added the pre-release class here to make it easy to test what things look like if the feedback button and milestone warning aren't present.

styles/_base.scss
```diff
@@ -117,48 +117,107 @@ ul, ol {
...
+        position: fixed;
+        top: $bar;
+        left: 0;
+        z-index: 1;
+        width: 100%;
+        height: $barLg;
+        overflow: hidden;
+
+        &:before {
+          // mask for the chat close button
```
jjeffryes: This is a hole through which the page background can show, the exact same size and shape as the close button.

styles/_base.scss
```diff
@@ -117,48 +117,107 @@ ul, ol {
...
+        }
 
-        .userIcon {
-          border-width: 2px;
-          border-style: solid;
-          width: 44px;
-          height: 44px;
-          margin-right: $pad;
+        &:after {
+          // false bottom border for the pageTab
```
jjeffryes: To allow the page tab border to move when the chat is open, so it doesn't go through the transparent close button, it's been replaced with a 1px high div.

styles/_base.scss
```diff
@@ -209,8 +268,8 @@ ul, ol {
 
   #chatConvoContainer {
     position: fixed;
-    right: 0;
-    bottom: 42px;
+    right: $scrollbar;
+    bottom: 1px;
```
jjeffryes: The designs have the bottom border of the chat conversation being visible, so I bumped it up a pixel.

styles/_mixins.scss
```diff
@@ -14,8 +14,12 @@
...
 }
 
-@mixin chatOpeningTransition($property) {
-  transition: $property 0.5s ease;
+@mixin chatOpeningTransition($props...) {
+  $transitions: ();
+  @each $prop in $props {
+    $transitions: append($transitions, #{$prop 0.5s ease}, comma);
+  }
+  transition: $transitions;
```
jjeffryes: I used multiple properties with the transition mixin at one point, but took them out. I left the enhancement to the mixin in place, it will be useful if later we need to apply this transition to more than one property.

styles/_theme.scss
```diff
@@ -26,6 +26,11 @@ button,
   input[type="radio"]:checked + label:before {
     box-shadow: inset 0 0 0 3px $secondary;
   }
+
+  &#pageTabBar:before {
+    // the chat button cutout shadow fills the rest of the tab bar with color
+    box-shadow: 0 0 0 200vw $primary;
```
jjeffryes: The peak through window for the close button fills the page tab with it's drop shadow. 

I made a demo of the effect here: https://jsfiddle.net/jjeffryes/8ew04tz5/

rmisio:
In the case when an overlay is present, it looks like the bottom border is missing on the button (also on the hover state):

image

rmisio:
Other than the above issue, this LGTM.

jjeffryes:
Well, nuts. The missing bottom bar on hover and overlays isn't going to be fixable with this technique.

I think we may need to live with it always having a bottom border, instead of how it is in the design.

rmisio:
@morebrownies Can we just give that chat button a bottom border at all times, at least for now? It's problematic for it to selectively have a bottom border because chat really doesn't know what's behind it. I'm also mildly concerned about it having to line up with pixel perfect precision with the bottom border of the tab bar.

If it's a big deal, I think we could circle back to it post launch, but IMHO it doesn't seem like something we should get stuck on now, especially since a lot of hours were already spent on it.

image

image

morebrownies:
@rmisio it should be fine to revisit some of the finer details later on. I'm going to test the Chat PR today.

rmisio:
@jjeffryes Given the fact that the chat button (at least for now) will always have a bottom border, are there any tweaks you need to make to this PR?

If you like, I could do the minimal CSS to get that bottom border back in... I'm just wondering if there are any other changes you need to make (e.g. is everything in this PR still needed)?

jjeffryes:
Yes, there are some changes I want to make, I should be able to do them today.

styles/_base.scss
```diff
@@ -117,48 +117,53 @@ ul, ol {
...
-          height: 44px;
-          margin-right: $pad;
-        }
+    .chatOpen &:before {
+      right: 0;
+    }
+
+    #pageContainer {
+      padding-top: $barLg;
+
```
jjeffryes: Previously the pageTabBar style was nested in contentFrame instead of pageContainer, this fixes the hierarchy, and puts the padding-top on pageContainer which makes layout more consistent.

styles/_base.scss
```diff
@@ -209,8 +214,8 @@ ul, ol {
 
   #chatConvoContainer {
     position: fixed;
-    right: 0;
-    bottom: 42px;
+    right: $scrollbar;
```
jjeffryes: Don't cover the scrollbar with the chat container.

styles/_base.scss
```diff
@@ -286,22 +291,43 @@ ul, ol {
...
-  border-style: solid;
-  z-index: 10;
-
-  .milestone-inner {
-    position: relative;
-    top: 10px;
+  display: none;
+
+  .pre-release & {
+    display: block;
```
jjeffryes: Adjusts the milestone callout on the bottom to not overlap the chat sidebar, and only be visible when the pre-release class is present. This makes it simpler to test the layout for release by turning that class off.

styles/_base.scss
```diff
@@ -286,22 +291,43 @@ ul, ol {
...
+  }
+
 }
 
 #doorbell-button {
-  right: 8px;
+  display: none;
+
+  .pre-release & {
+    display: block;
```
jjeffryes: The feedback button is also adjusted, and dependent on the pre-release class, so you can hide it and the milestone callout together.

styles/modules/_chat.scss
```diff
@@ -8,11 +8,10 @@
   border-radius: 0;
   font-size: $tx1;
   border-right-width: 0;
-  border-bottom-width: 0;
```
jjeffryes: Always show the bottom border.

styles/modules/_chat.scss
```diff
@@ -8,11 +8,10 @@
...
   @include chatOpeningTransition(right);
 
   .chatOpen & {
-    right: 1px;
+    right: $scrollbar;
```
jjeffryes: Don't cover the scrollbar.

jjeffryes:
@rmisio I've updated this PR to add the bottom border, and removed the changes we don't need. I left in the changes to make sure the scrollbar isn't covered, and to improve the placement of the milestone and feedback elements.

rmisio:
A couple things:

1.) The tab bar no longer has a bottom border.

image

Is that what we want? I thought the compromise (at least for now) was just to give the chat button a bottom border at all times and move it up a bit so it's not overlapping the bottom border of the tab bar, for example:

image

@morebrownies what do you think? I suppose getting rid of the bottom border on the tab bar also solves the issue.

2.) A right border is missing on the chat close button:

image

jjeffryes:
No, that's just a mistake. I forgot to revert that part of the code with the rest of it.

rmisio:
Another thing is that the chat was positioned so that it would be above the testing badge. Since that badge is moving now when chat opens, it leaves a gap.

image

I, personally, preferred the badge staying on one place. With it moving over, there's a good chance it will overlap critical elements on different pages which would require changes to potentially multiple pages.

I would have it not move over and leave the chat positioned on top of it. Once the badge no longer makes any sense, we could simply remove it and re-position chat so it goes all the way to the bottom.

morebrownies:
Haven't tested this PR yet. I'll fire it up shortly. 

@rmisio Could we possibly move the "Milestone 1 testing" badge to the left size of the viewport? It would be nice to give chat the full height so that we don't need to go back and clean it up later.

jjeffryes:
I just updated this to include the solid border on open with a transition, to see what it would look like. It doesn't change the width. 

The background/border can be removed by deleting a few lines of css.

There are a number of other small tweaks in there now too, to make sure everything lines up everywhere (I spotted a few things that were off by a pixel on my screen). 

The milestone tag doesn't move now, and I changed the feedback button to be on the left side of the milestone tab, so it's still visible when chat is open if you want to leave feedback about the chat.

morebrownies:
that's cool @jjeffryes 👍  

definitely starting to feel better

jjeffryes:
We don't want the milestone badge on the left, because it will overlap the status messages for saving and publishing.

morebrownies:
👍

jjeffryes:
This is updated now per our conversation, and applies a non-opaque background to the chat sidebar when it is open, and makes the conversation area match the width of the chat sidebar.

rmisio:
We're still missing a bottom border on the chat close button when there's an overlay behind it:

image

Wolf seemed fine with styling the button as so. If it's taking so long to make it some other way... I think what's below is fine (we could always circle back):

image

jjeffryes:
No, that's intentional. When the chat has a solid background, the button looks like a tab so it's clearly associated with the chat.

@morebrownies what do you think, should it be like that, or detached?

rmisio:
Wait, just to confirm, you are saying this is correct?

image

It looks absolutely broken to me.

jjeffryes:
Yes, that is correct. 

Making it stay tabby-looking with the transparent background is tough, but I think it needs to be visually associated with the chat sidebar in some way, just being a second close button next to the modal close button and below the app close button, with nearly the exact same appearance, is really not great.

rmisio:
Ok, I'll have to take your word for it that that looks like a tab. Looks like nothing more than a button without a bottom border to me. Maybe my contrast / brightness settings are making it look different to me than to others.

Anyhow, if @morebrownies is fine with this PR, then from my side, it's ready to merge.

morebrownies:
What looks broken to you @rmisio? It looks like some borders need to be added, but beyond that, are there additional elements that appear off? Please explain... 

We're making very quick decisions and it is likely some details may still be out of whack. Definitely point them out if you notice anything.

rmisio:
image

> What looks broken to you @rmisio? It looks like some borders need to be added, but beyond that, are there additional elements that appear off? Please explain...

In that image, the close button to me looks broken. It looks like a button with a missing border. Josh is claiming it's tabby-looking and that it looks like it should.

That's my point. If it also looks OK to you, I'll chalk it up to me seeing something others are not and I'll happily let it go.

morebrownies:
I definitely think adding a border to the chat panel would help connect the button a lot better. The transparency difference makes the visual a bit weird, but it looks better than having no border.

screen shot 2017-03-08 at 5 57 12 pm

rmisio:
Yeah... that looks like a tab. The whole issue though was that it was problematic to have the button line up with the top of the chat.

@jjeffryes Is the border do-able in that way? With the close button being separate from chat, it requires some pixel perfect precision.

morebrownies:
Yeah @rmisio, it seems we kind of looped back to the pixel perfection scenario. If you guys can get it close, I feel I can tighten it up a bit after we get these PRs merged. I have it working pretty nicely in my DOM, but I'd rather wait to open a PR since there are already a few active ones. 

Let's wait and see what @jjeffryes comes up with too. Maybe he figured out a simple work around.

morebrownies:
Great work guys. This is looking a lot better. I see 2 minor things left to clean up:

1) A right border on the chat panel would be nice. See the screenshot below. When the scrollbar moves lower on the page, it exposes the missing border more clearly.

2) We discussed having an active state on the conversation that is currently active. This would be helpful if we could sneak it in. 

screen shot 2017-03-09 at 12 30 16 pm

rmisio:
> We discussed having an active state on the conversation that is currently active. This would be helpful if we could sneak it in.

@jjeffryes I think I'l get that one.

jjeffryes:
@morebrownies this is ready for another look

jjeffryes:
Oh, nuts. More comments came in while I was working on this. I'll add the right border and active state shortly.

morebrownies:
Nice! Loving that the tab bar slides over now 👍 

Ok, cool. Was about to call the border out again hehe

This is looking good to me. There are a few typography things I'm planning to clean up after you guys get this merged into master. Otherwise, looking good. 👍

morebrownies:
There are some extra lines displaying next to the avatars too @jjeffryes 

screen shot 2017-03-09 at 2 22 15 pm

jjeffryes:
The chat head area now has borders all around, and some other tweaks were made. 

Known issue: on Windows 10 retina monitors, which any amount of zoom set, the top of the chat head area is off by 1 pixel. This appears to be a bug in Chrome and how it rendered fixed or absolute positioned elements when the screen is zoomed (the issue disappears if zooming is turned off, but then the app is tiny and unreadable. Zooming is set to 150% by default). 

image

jjeffryes:
@morebrownies in what situation did you see those extra lines to the right of the avatars? I wasn't able to reproduce it on Windows or Mac.

rmisio:
> We discussed having an active state on the conversation that is currently active. This would be helpful if we could sneak it in.

This is done in #375.

morebrownies:
False alarm @jjeffryes. I'm not seeing them anymore.

morebrownies:
@jjeffryes deleting a conversation results in the chat panel height getting stuck at the condensed size.

screen shot 2017-03-10 at 12 01 17 pm

rmisio:
> deleting a conversation results in the chat panel height getting stuck at the condensed size.

@morebrownies Can we please log a bug for that and I'll fix that later? I'll do it in the main chat PR.

morebrownies:
Sure #391

jjeffryes:
Other than #391, is this accepted? If so, we can merge this into the main chat PR, and then merge that into master if it's accepted. @rmisio @morebrownies

rmisio:
I'll do 391 in the Chat branch after this one is merged into it. It will only take a few minutes.

If @morebrownies is ok with this one, we could go ahead and merge.

morebrownies:
Doing a final pass through it now

morebrownies:
I think this is about set. 2 very small things left:

1) @hoffmabc + @cpacia wanted to see the width of .headContent fill the width of #chatContainer 
screen shot 2017-03-13 at 9 37 01 am
_how it currently looks_

2) The color of the border on #chatContainer and #chatCloseBtn shouldn't be dark. We'll want the actual chat conversation window to be emphasized a bit, instead of the entire chat panel. 
screen shot 2017-03-13 at 9 33 08 am


Making these 2 changes should result in chat looking like this:

screen shot 2017-03-13 at 9 41 08 am

morebrownies:
Scrollbar and emojis looks good @jjeffryes. Did you see my message above @jjeffryes? Otherwise looks good! Nice work on this @rmisio @jjeffryes 🍻

jjeffryes:
@morebrownies updated.

morebrownies:
Looks good to me 👍  

I'm going to make a few minor tweaks to Chat after you guys get this merged into master.

jjeffryes:
This should be ready to merge into Chat, @rmisio

js/views/chat/Conversation.js
```diff
@@ -425,7 +425,6 @@ export default class extends baseVw {
     if (this._isOpen) return;
     this._isOpen = true;
     getBody().addClass('chatConvoOpen');
-    this.$messageInput.focus();
```
rmisio: Why was this removed? Don't we want to focus the message input field when you open the convo window?

PR #380, opened 2017-03-03
Emojis

### LINKED ISSUE #251: Chat emoji
Is the intention for the chat box to provide support for Slack-like emojis, e.g. `:simple_smile:` will display a smiley face?

rmisio:
closes #251 

This PR adds in emoji support to chat using the Twemoji library which replaces Unicode emoji characters with images.

Please Note:
This PR includes the Chat PR (#375). Once that is merged in, the file diffs should will be a bit leaner here.


rmisio:
Once #375 is merged, I'll need to make a minor tweak or two to this one to account for the adjusted Chat width.

js/views/chat/Conversation.js
```diff
@@ -0,0 +1,629 @@
...
+      // Developer error - this shouldn't happen.
+      console.error('There was an error saving the chat message.');
+      console.dir(save);
+    }
+
+    $(e.target).val('');
+  }
+
+  onBlurMessageInput(e) {
+    this.lastMessageInputCursorPos = e.target.selectionStart;
```
rmisio: I think some refactor made it unnecessary. Good catch, I'll remove it.

js/utils/dom.js
```diff
@@ -19,3 +19,33 @@ export function isScrolledIntoView(element) {
   // Check its within the document viewport
   return top <= document.documentElement.clientHeight;
 }
+
+// http://stackoverflow.com/a/41426040/632806
+export function insertAtCursor(myField, myValue) {
+  if (document.selection) {
```
rmisio: I'm thinking it might be wiser to leave it in so this is more portable. Especially, since this is a generic utility function, it's very likely we'd use it in our future web-app and it seems easier to have the cross-browser stuff there rather than have to hunt it down later.

I wouldn't of made an effort to put the cross browser stuff in, but since it came for free, I brought it over.

jjeffryes:
@rmisio I merged the chat PR, this PR has some merge conflicts, can you de-conflict those?

PR #382, opened 2017-03-06
Refactor of the listing model and related code to adhere to new listing schema.

rmisio:
This is intended to be used with this PR on the server:
https://github.com/OpenBazaar/openbazaar-go/pull/392

The respective merges should be coordinated.

FWIW - the amount of areas potentially affected by this change are vast. I did my best to test, but it's possible I missed something (and / or there's some scenario the unit tests don't cover). If we find any listing related oddities, please send them my way.


jjeffryes:
This looks good to me. We can merge when @cpacia is ready with his branch.

PR #392, opened 2017-03-10
Add mod on user page and user card

### LINKED ISSUE #372: Add processing to follow/unfollow buttons on the user page
The user page should be updated to use the new processing button template on the follow/unfollow button, so it will show the processing spinner when the button is clicked.

jjeffryes:
This adds the ability to add or remove a user as one of your moderators from their user card or user page.

It also refactors the home page to use the full user card view, which solves some other issues.

This code is dependent on #377, do not merge it until that PR has been merged.

Closes #372


morebrownies:
Ok I just had a chance to play around this with PR @jjeffryes 

A few small things...

1) The "Add as Moderator" button copy should probably be changed to "Details" since it opens the overlay with more information.

2) It's not very clear what "$13,779.37 (+9) per dispute" means. I believe the 9 should be 9%

screen shot 2017-03-14 at 11 22 32 am

I also tested from the user card and it seems to work well. We'll want to make sure the % sign gets added here too:

screen shot 2017-03-14 at 11 26 15 am

Looks good otherwise 👍

jjeffryes:
@morebrownies this has been updated.

js/views/userPage/Home.js
```diff
@@ -19,33 +33,86 @@ export default class extends BaseVw {
...
       'click .js-termsLink': 'termsClick',
-      'click .js-termsClose': 'termsClose',
-      'click .js-addModerator': 'addModeratorClick',
+      'click .js-addModerator': 'changeModeratorClick',
       'click .js-guid': 'guidClick',
       'mouseleave .js-guid': 'guidLeave',
     };
   }
 
+  get ownMod() {
```
rmisio: Since we're repeating this, can we make an isOwnModerator(guid) method on the settings model?

js/views/userCard.js
```diff
@@ -0,0 +1,213 @@
...
+
+    this.ownGuid = this.guid === app.profile.id;
+    this.followedByYou = followedByYou(this.guid);
+
+    this.loading = !this.fetched;
+    this.settings = app.settings.clone();
+
+    this.listenTo(this.settings, 'sync', () => {
+      app.settings.set(this.settings.toJSON());
+      this.$modBtn.toggleClass('active', this.ownMod);
```
rmisio: For toggling of the active class, I would listen to `change:storeModerators` on app.settings. That way if the change was initiated outside of here (e.g. the store tab in the settings modal), the button will reflect the correct state.

js/views/userPage/Home.js
```diff
@@ -1,15 +1,29 @@
...
     this.ownPage = options.ownPage;
-    this.currentModerator = false; // TODO set if this user is one of the viewer's moderators
+
+    this.userCard = new UserCard({ model: this.model, hideControls: true });
+
+    this.settings = app.settings.clone();
+
+    this.listenTo(this.settings, 'sync', () => {
+      app.settings.set(this.settings.toJSON());
+      this.$addModeratorLbl.toggleClass('hide', this.ownMod);
```
rmisio: Same as the comment on the user card... if we move the adjustment of the state of the two buttons to the change:storeModerators on app.settings then we'll be good if some other area initiated the change.

PR #399, opened 2017-03-16
Master with revert reverted

rmisio:
This branch needed to fix an issue arising from the Emojis branch being merged into the chatCSSMods branch and then reverted. So, in order to get emojis into master, the revert needed to be reverted - which is what is done here.

http://stackoverflow.com/a/30341780/632806


hoffmabc:
This wins revert of the year @rmisio

PR #402, opened 2017-03-17
Variants

rmisio:
This PR implements the variants UI in the Edit Listing modal.


js/views/modals/editListing/EditListing.js
```diff
@@ -553,6 +706,8 @@ export default class extends BaseModal {
     } else {
       // client side validation failed
       this.$saveButton.removeClass('disabled');
+      console.log('no soup for you');
```
jjeffryes: Stray console.log?

js/views/modals/editListing/EditListing.js
```diff
@@ -553,6 +706,8 @@ export default class extends BaseModal {
...
+      window.soup = this.model;
```
jjeffryes: What's window.soup?

rmisio: lol... yeah that's just some debugging stuff I need to remove. I expose the model globally so that I could easily see the validation errors, since not all of them make it to the UI (some only happen if the dev messes up and isn't providing the data in the right format).

Good catch!

morebrownies:
Looking great @rmisio !

I noticed a couple of things, which I think are all very simple fixes:

1) When the price input is null, the variant table does not display. Is this intentional? Or a bug? I definitely expected the table to display and couldn't figure out why it wasn't displaying.

2) The surchage input should probably default to 0. I got hit with 30 inputs to fill out. 
screenshot 2017-03-20 16 37 12

3) The quantity input should also probably default to 0. 

4) We may want to put a max height on the variant table since it can get very tall. Maybe 500px?

That's all I found. Great work. 

Wolf

js/views/modals/editListing/EditListing.js
```diff
@@ -863,6 +1078,66 @@ export default class extends BaseModal {
...
+      this.$('.js-inventoryManagementSection').html(this.inventoryManagement.render().el);
+      this.listenTo(this.inventoryManagement, 'changeManagementType', this.onChangeManagementType);
+
+      // render variant inventory
+      if (this.variantInventory) this.variantInventory.remove();
+
+      this.variantInventory = this.createChild(VariantInventory, {
+        collection: item.get('skus'),
+        optionsCl: item.get('options'),
+        getPrice: () => this.getFormData(this.$itemPrice).item.price,
```
jjeffryes: If a user hasn't entered a price yet, this gets passed as a string with a value of zero instead of a number, which throws an error in the currency.js code. 

Casting it as a number fixed it when I was testing it.

jjeffryes: If a user hasn't entered a price yet, this gets passed as a string with a value of zero instead of a number, which throws an error in the currency.js code. 

Casting it as a number fixed it when I was testing it.

rmisio:
@morebrownies @jjeffryes Ok, all the changes are in. Assuming everything looks good, this should be ready for merge.

> We may want to put a max height on the variant table since it can get very tall. Maybe 500px?

@morebrownies I did put this one in there. See how it feels. For me, it feels a bit clunky when you have the dual scrollbars. Maybe we should increase the max-height to 700 or 800 to limit the times both would appear? Anyhow... totally up to you though. Feel free to adjust or not at your discretion.

> When the price input is null, the variant table does not display. Is this intentional? Or a bug? I definitely expected the table to display and couldn't figure out why it wasn't displaying.

That was a bug. Great catch!

morebrownies:
thank you @rmisio 

I like that the table fits comfortably in the default viewport size. It looks like we have a bit more real estate to work with. I'd be cool bumping the  max-height to 600px. Anything beyond that may start to bleed outside of the default window size. 

screenshot 2017-03-21 09 23 22

jjeffryes:
@morebrownies is this PR accepted at this point?

morebrownies:
Yes @jjeffryes

jjeffryes:
@rmisio you ready for this PR to be merged? I just want to confirm there was nothing else you wanted to do with it.

rmisio:
image

PR #405, opened 2017-03-20
Adjust css so table has even margins and full width inputs.

jjeffryes:
A few minor tweaks to the CSS so the variant inventory table has even margins, and the inputs fill their cells. It seemed more expedient to make this a PR than to put in a comment, since I'd already made the change in the web dev panel when checking why the right padding was 0.

Before:
image

After:
image

### DIFF
--- a/styles/modules/modals/_editListing.scss
+++ b/styles/modules/modals/_editListing.scss
@@ -30,7 +30,7 @@
         position: absolute;
         top: 0;
         left: 0;
-      }      
+      }
 
       .btn {
         width: 61px;
@@ -69,7 +69,7 @@
 
             &.addElement {
               cursor: default;
-              
+
               .imagesIcon {
                 font-size: 50px;
                 @include center;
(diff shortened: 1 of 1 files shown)


jjeffryes:
Closing this because I noticed an issue.

jjeffryes:
This patch is optional, @morebrownies should weigh in on it vs. the main PR.

morebrownies:
Looking great @rmisio !

I noticed a couple of things, which I think are all very simple fixes:

1) When the price input is null, the variant table does not display. Is this intentional? Or a bug? I definitely expected the table to display and couldn't figure out why it wasn't displaying.

2) The surchage input should probably default to 0. I got hit with 30 inputs to fill out. 
screenshot 2017-03-20 16 37 12

3) The quantity input should also probably default to 0. 

4) We may want to put a max height on the variant table since it can get very tall. Maybe 500px?

That's all I found. Great work. 

Wolf

morebrownies:
I like that change @jjeffryes

morebrownies:
Just realized I added my comment to the wrong PR.

morebrownies:
@jjeffryes there probably needs to be a min width on the inputs to prevent this from happening

screenshot 2017-03-20 19 06 00

PR #407, opened 2017-03-21
Tweaks to the table layout in the combos.

jjeffryes:
I found a bug where a long tag in the variant can break the layout. This patch adds a wrapper class to be used when a select2 tag input is used in a flex container, which will prevent it from growing beyond the size of the container.

This also has a few small tweaks to the combo table that will even out the columns a bit.

**Long tag before:**
image

**Long tag after:**
image

**Tweaked combo table:**
image

### DIFF
--- a/js/templates/modals/editListing/variant.html
+++ b/js/templates/modals/editListing/variant.html
@@ -1,20 +1,22 @@
-<div class="col6 simpleFlexCol">
+<div class="col6">
   <% if (ob.errors['name']) print(ob.formErrorTmpl({ errors: ob.errors['name'] })) %>
   <input type="text" class="clrBr clrP clrSh2 variantNameInput js-variantNameInput" name="name" value="<%= ob.name %>" placeholder="<%= ob.polyT('editListing.variants.titlePlaceholder') %>" maxlength="<%= ob.max.nameLength %>">
 </div>
-<div class="col6 simpleFlexCol">
+<div class="col6">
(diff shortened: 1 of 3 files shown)


styles/modules/modals/_editListing.scss
```diff
@@ -261,13 +261,17 @@
             border-style: solid;
             padding: $pad;
             word-break: break-word;
-            min-width: 90px;
+           // min-width: 90px;
```
rmisio: Can we please keep the min-width? It allows the table to get a horizontal scrollbar when it has a lot of content. Otherwise things get very squished:

image

js/templates/modals/editListing/variant.html
```diff
@@ -1,20 +1,22 @@
-<div class="col6 simpleFlexCol">
```
rmisio: Can we please keep the simpleFlexCols? They are need so that fields remain bottom aligned when only one has an error:

image

styles/modules/modals/_editListing.scss
morebrownies: +1 we need a min width on the columns. Otherwise the legibility gets pretty bad

morebrownies: screenshot 2017-03-21 09 41 34

jjeffryes:
Closing this for now, it can be applied later to master as only a patch to fix the variant overflow, and the other changes removed.

PR #415, opened 2017-03-22
Cleaned up typography in chat

morebrownies:
Cleaned up the typography, borders and spacing a bit. Also removed the "bookmark" button from the address bar since we're no longer supporting that functionality.


jjeffryes:
This looks good.

PR #416, opened 2017-03-24
Clean up

morebrownies:
**Changes**
- Changed the discover icon from the "eye" to the OpenBazaar tent
- Cleaned up some copy
- Cleaned up styling in chat

I didn't end up making changes to variants. I tried, but couldn't get them working. I will log a few issues that we can clean up later (after 2.0 release)


jjeffryes:
This looks good. @rmisio if you sign off as well we can merge it in.

PR #424, opened 2017-04-06
Wallet send receive

rmisio:
Implements the Send & Receive Money sections in the wallet.


js/models/wallet/Spend.js
```diff
@@ -0,0 +1,145 @@
...
+
+/**
+ * Use this function anytime you want to spend money from the internal wallet.
+ * This function will ensure the balance is properly updated and the Wallet view
+ * is informed so that it's UI could reflect the transaction.
+ * @param {Object} fields An object containing the arguments to the spend call -
+ * (address, amount and optionally currency, memo and feeLevel).
+ * @returns {Object} The jqXhr representing the POST call to the server.
+ */
+export function spend(fields) {
```
rmisio: @jjeffryes You'll want to use this function in the Buy Wizard to pay from the internal wallet. It's important that any payments from the internal wallet go through this function since this function will update common state (e.g. wallet balance) as well as relevant state in the wallet.

morebrownies:
> A new address should be generated each time this Receive Money button is clicked.

@rmisio I don't seem to be generating a new address each time I hit receive

morebrownies:
^ maybe that is a server thing?

morebrownies:
> The Cancel button will return to the Send Money section.

Also the Receive money screen should have a cancel button

screen shot 2017-04-06 at 11 05 33 am

morebrownies:
I'm attempting to send some testnet coins to my wallet now. Waiting for confirmations :)

rmisio:
> A new address should be generated each time this Receive Money button is clicked.
> @rmisio I don't seem to be generating a new address each time I hit receive

That is an incorrect requirement. Any time a new transaction comes in, then we will generate a new address. If no transaction has come, it's pointless to call the server, because the same address will be returned.

rmisio:
> Also the Receive money screen should have a cancel button

I didn't include that because it seemed inconsistent to me that the Receive section has a Cancel, but not the Send. If you think it's important for it to be there, I could it add it in.

morebrownies:
> I didn't include that because it seemed inconsistent to me that the Receive section has a Cancel, but not the Send. If you think it's important for it to be there, I could it add it in.

The Send screen doesn't really need a Cancel button because that is the default screen. Clicking "Cancel" when viewing Receive was intended to take you back to Send. I know there is a "Send" button but it felt better to also have a "Cancel" button.

morebrownies:
> That is an incorrect requirement. Any time a new transaction comes in, then we will retrieve a new address. If no transaction has come, it's pointless to call the server, because the same address will be returned.

@cpacia you in agreement with this too?

morebrownies:
@rmisio when I click the "refresh" icon in the address bar while I have the Wallet open, the Wallet goes away, and then when I re-open it I see this:

screen shot 2017-04-06 at 1 15 32 pm

rmisio:
> @rmisio when I click the "refresh" icon in the address bar while I have the Wallet open, the Wallet goes away

This is because the Wallet is a modal and therefore does not have a history state. When you click Refresh, you're really refreshing the underlying page.

> ... and then when I re-open it I see this (blank wallet):

Good catch! Fixed.

jjeffryes:
@rmisio can you fix the merge conflicts?

PR #425, opened 2017-04-06
Search

### LINKED ISSUE #203: Search View
The Search View is a view that is shown when the user is searching for listings on the OpenBazaar network. It is used to view, filter, and open those listings.

The Search View should be shown when:
- [ ] A search term is in the address bar and the enter key is pressed (a search term is identified by not being a valid route).
- [ ] A search address is in the address bar and the enter key is pressed.
- [ ] The search view is activated programatically.

The Address Bar
- The first time the address bar is focused, show a tooltip explaining how search works.
- The first time search results are returned, show a tooltip with the name of the search service and an input to change to a different search service.
- The address shown in the address bar should include the url of the search service and the filters being searched for, so it can be copy and pasted to another client and send the same query.
- If a search term is entered into the address bar, it is replaced by the full address of the search query.
- It might look something like: ob2://search.ob1.io?q=shoes&type=physical&reviews=4

The Search API
- The client will connected to an external search service to send and receive search data. 
- The search service will be set to the default search service when the client is run for the first time, and can be changed at any time.
- Changing the search service will probably be done in the settings modal (TBD).
- All search services will use the same format for data. This view should be work with any service that uses the correct data formats.

The Search View should show the following static data:
- The log of the search service.
- The name of the search service in the alt value of the logo.
- The standard filters that are available for the search service.
- The dynamic filters available for the search service (built dynamically from the data returned).
- The list of sortable parameters in a select list. It is not shown if 1 or less parameters are returned.

The following elements are interactive:
- Filters
    - Each filter is applied when a selection is made (radio or checkbox clicked). If the filter is a freeform text field, it will have a submit button.
     - Each filter shows up to 5 options. If there are more options, show a "X more..." link. 
     - Clicking on the X more... link will show the additional options, along with a "Show less" link that will return the filter to showing the first 5 filters.
     - Any filter with type checkbox will have a link to select all of the options. Clicking it a second time will de-select all of the options.
- The Shipping Filter
    - The Shipping Filter is always the first filter, and is always present (even if type is a filter, and the type selected is digital). 
    - The Shipping Filter has a select box to select a country based on the user's saved addresses.
    - The country from the user's first saved address is always pre-selected.
    - The Shipping Filter has a link to add a new address. 
    - Clicking the add address link will open the Settings Modal with the Addresses tab active.
    - When the Settings Modal is closed, the Shipping Filter will update the list of selectable addresses.
    - The Shipping Filter has a checkbox for selecting a ships to country, and for only showing listings with free shipping to the selected country.
    - if the free shipping checkbox is checked but no country is selected, all listings with free shipping to any country are shown.
- The Listings
    - When clicked on, each listing will open the listing detail modal for that listing
    - If the avatar of the seller is hovered over, a small modal with information about that seller should appear (To Be Designed). At minimum it should show the data included in the listing for that seller. 
- The grid/list button will toggle the design between a list and a grid.
- The width button will toggle the display between normal width and 100%.
- The Sort By select list will send a new query with the sort by parameter.
- Pagination (To Be Designed).

The Results
- The results are ordered by the currently selected sort by parameter. If only one parameter was returned by the search service, the sort by select list is not shown, and the results are sorted by that parameter. If no sort by parameter was returned, the results are shown in the same order as they are in the data.

## Data

**Initial load**
When the client starts, or the search provider is changed, the client should call the search API, and it should return the following data:
- The name of the search service.
- The address of the search service logo.
- A list of standard filters that are available.
- A list of dynamic filters that are available, with data to construct the filters.

**Search Query**

The search query should be sent with parameters for the query.

The parameters should be:
- The query (one or more words)
- All the filter parameters (filter name, value)
- The listing to start with (return listings starting with X)
- The number of listings to return (Y listings total, starting with X)
- The sort by parameter.

**Search Data Returned from the Search Query**

The data returned from the search query should match the data from the listings API, with the following exceptions:

- Data required by the available filters should be present in each listing if applicable (ie: if condition is an available standard filter, condition must be returned in the data. If size is an available custom filter, listings that have a size should have that returned in their data)
- Data required by a custom sort option should be returned in each listing if applicable. 
- Additional data about the seller may be returned (To Be Designed). 

**Designs**

1. Search View after receiving search data
image

2. No results found
image

3. First time focusing on the address bar
image

4. First time search results are recieved
image

jjeffryes:
This PR adds search and discovery to the client.

Tests will be added in a patch or a separate PR in order to not hold up the review.

@morebrownies I added a default image (the same one used in the address bar) for the on-page search if the logo the search provider gives fails to load, which is happening currently with the default search provider.

I also added error modals if the search provider call fails. If the call is to a different provider than the current default (ie: you pasted in a http address), the error message will give you the option to try the same search term with your default provider.

If the error occurs with the user's default provider, they are given the option to open the settings modal and change it. On doing so, the search view will reload and try to use the corrected url.

You can test this by putting a bogus url into the address bar, and by changing the search provider in settings/general to something incorrect.

Closes #203


morebrownies:
Should the search page be functional @jjeffryes? I'm not seeing any listings.

screen shot 2017-04-06 at 8 16 23 pm

jjeffryes:
@morebrownies it looks like you don't have an up to date copy of the PR? That, or something is breaking during the page load. Do you see any errors in the console?

I just tried it on a new machine, and it is working normally. You should see this when clicking discover:

image

rmisio:
When there's no results, the previous arrow shouldn't be active:

image

jjeffryes:
The issue Mike saw was he had run this branch at least once before when the url for the default was different, and that was saved to his settings.

I've updated the code to account for that, I'll push the fix after reviews are in.

jjeffryes:
I've added a state for no results found, and fixed the issue with the previous page button being active if no results are found. I'll push that code after reviews are completed so I don't disrupt anything in progress.

image

morebrownies:
Ok thanks @jjeffryes I'm now seeing results. 

None of the listings seem to load when I click them though and the filtering/sorting also doesn't seem to work.

morebrownies:
Also, we probably shouldn't display the suggestions label if there aren't any suggestions

screen shot 2017-04-07 at 9 16 58 am

morebrownies:
In settings, I have my currency set to USD, yet the search results display in BTC.

screen shot 2017-04-07 at 9 18 36 am

screen shot 2017-04-07 at 9 19 37 am

jjeffryes:
Re: suggestions, those are just stubbed in, I don't think they exist yet in the search server code.

I can comment them out for now.

morebrownies:
Also @jjeffryes is it possible to display the user's avatar? Right now it appears you're double outputting the listing image. 

screen shot 2017-04-07 at 9 31 40 am

jjeffryes:
@morebrownies good catch on the currency, there was an issue with the server returning 'btc' instead of 'BTC'. I have that fixed, will push it when Rob's review is done.

The avatar being a listing image is a server issue, Tyler is working on it.

jjeffryes:
Tyler's working on suggestions, when he has them available I'll code them in. That might be a later PR.

morebrownies:
Ok cool I found a listing that loads @jjeffryes. I noticed when I click the listing, the overlay does not display the Store name / view store link. We should definitely get that outputting otherwise it's not clear what store you're looking at.

**Current**
screen shot 2017-04-07 at 9 51 03 am

**Expected**
screen shot 2017-04-07 at 9 52 47 am

js/collections/Results.js
```diff
@@ -0,0 +1,54 @@
+import { integerToDecimal } from '../utils/currency';
```
morebrownies: > Seems like some spacing is needed between the icon and text:

Glad you bought it up. I feel the same way. It looks like the spaces are in the DOM, but they're not outputting for some reason. 

> Should we be allowing search within this drop down?

Yeah, not super useful. In most cases, the dropdown would probably be under 5 options.

> The left border of the search button is not changing on hover:

If it's easier to put a gap between the input and button, that should be fine too.

jjeffryes:
The spacing on the icons in the filters looks correct on both my Mac and my Windows machine, can you think of a reason it would be different on yours, Rob?

screenshot 2017-04-07 15 39 17

jjeffryes:
In Windows the flag for Germany renders as just "DE" and the USA as "US". I'm not sure how much we can do about that, it shows up that way in the json when I look at the same text in a browser. The spacing is the same issue, the icons and text are whatever the search provider sends us. 

image

jjeffryes:
Any filters shown are what the server returns, right now I believe it's returning a default set with a blank search request. The client shouldn't intervene in any way in the filters, whatever the server sends should be assumed to be correct, since there's no way to know what filters a search provider will want to send.

jjeffryes:
All of the issues noticed in the review should be corrected now. Let me know if I missed anything.

morebrownies:
Ok testing now

morebrownies:
Store name not displaying @jjeffryes 

screen shot 2017-04-07 at 2 33 51 pm

jjeffryes:
@morebrownies that's correct. The search provider doesn't send the store name currently, until it does it won't display (I mentioned that in one of the 300 comments above. ;) )

morebrownies:
ah, ok definitely missed that. my bad

morebrownies:
Hey @jjeffryes sorry found one more small detail. On a user's store, when no search results are found, we display the following:

screen shot 2017-04-07 at 3 31 49 pm

In search/discover, we display:
screen shot 2017-04-07 at 3 33 02 pm

It would be nice if those could match up.

tyler-smith:
As noted above, there are a few places the search server isn't working as expected, however I've reviewed the related code in the client and everything appears okay, so when I get the backend issues shored up it should be all good to go.

js/views/search/Search.js
```diff
@@ -0,0 +1,248 @@
...
+    // initial render to show the loading spinner
+    this.render();
+
+    // query the search provider
+    this.callSearch = $.get({
+          url: searchURL,
+          dataType: 'json',
+        })
+        .done((data, status, xhr) => {
+        // make sure minimal data is present
```
rmisio: Could we also only do the done() handler logic if the call isn't aborted?

js/views/search/Search.js
```diff
@@ -0,0 +1,248 @@
...
+
+  get filterQuery() {
+    // return all currently active filters in the form of a query string
+    return this.$filters && this.$filters.length ? `&${this.$filters.serialize()}` : '';
+  }
+
+  processTerm(term) {
+    this.term = term || '';
+    // if term has spaces, replace them with +
+    // if term is false, search for *
```
rmisio: I think you want to be running these variables through encodeURIComponent() - spaces are just one thing to be handled as there are other non-url friendly characters that need to be encoded. To be safe I would apply this to all these vars, not just term, e.g:

```
  const query = `q=${term ||  '*'}`;
  const page = `&p=${this.serverPage}&ps=${this.pageSize}`;
  const encodedQueryString = encodeURIComponent(`${query}${this.sortByQuery}${this.filterQuery}${page}`);
  const searchURL = `${this.sProvider}?${encodedQueryString}`;
```

js/views/search/Results.js
```diff
@@ -0,0 +1,167 @@
...
+      // update the address bar
+      app.router.navigate(newURL, { replace: this.firstRender });
+    } else {
+      // show the loading spinner
+      this.$el.addClass('loading');
+
+      const newPageCol = new ResultsCol();
+      this.pageCollections[page] = newPageCol;
+
+      newPageCol.fetch({
```
rmisio: Can we give the same "abort" treatment here as in the callSearchProvider call. Basically, we only want the most recent loadPage to be happening, so:

```
if (this.newPageFetch) this.newPageFetch.abort();

this.newPageFetch = newPageCol.fetch()...
```

Then we'd want to "opt-out" of the handlers if the call has been aborted.

Finally, in this View's remove can we please abort any this.newPageFetch calls?

js/views/search/Search.js
rmisio: Can we also abort the fetch in this View's remove?

js/utils/templateHelpers.js
```diff
@@ -4,11 +4,17 @@ import {
   isHiRez, isLargeWidth, isSmallHeight, getAvatarBgImage,
 } from './responsive';
 import { upToFixed } from './number';
+import twemoji from 'twemoji';
 
 export function polyT(...args) {
   return app.polyglot.t(...args);
 }
 
+export function parseEmojis(msg) {
```
rmisio: Minor, but can we change the argument to 'text'. 'msg' was appropriate in the chat view because it was parsing a chat message, but since we've added this to be used generically, I think the argument name should follow suit.

js/views/search/Search.js
```diff
@@ -0,0 +1,248 @@
...
+
+    const errorDialog = new Dialog({
+      title,
+      msg,
+      buttons,
+      showCloseButton: false,
+    }).render().open();
+    this.listenTo(errorDialog, 'click-changeProvider', () => {
+      this.changeProvider();
+    errorDialog.close();
```
rmisio: Looks like some indentation is off here and also the `errorDialog.close` a few lines down.

styles/components/_listingsGrid.scss
```diff
@@ -1,15 +1,65 @@
 .listingsGrid {
```
rmisio: Do you think it's worth it to put a comment in this file that these styles are applied to both the Search and Store pages, so please be mindful of both when making any changes? Your call.

js/views/search/Search.js
```diff
@@ -0,0 +1,229 @@
...
+      this.serverPage = options.serverPage || 0;
+      this.pageSize = options.pageSize || 12;
+      // if the term was not a url, process the term before calling the search provider
+      this.processTerm(term);
+    }
+
+    this.usingDefault = this.sProvider === app.localSettings.get('searchProvider');
+
+    // if not using a passed in URL, update the default provider if it changes
+    this.listenTo(app.localSettings, 'change:searchProvider', (_, provider) => {
```
jjeffryes: Sure. This was just a copy paste from something else that used an underscore there.

js/views/search/Search.js
jjeffryes: Is it possible for done to fire and have a statusText of abort? If the call was aborted, wouldn't it fail instead?

jjeffryes: (it looks like it's possible in the W3C spec, though it seems like it would be kind of a rare thing to happen)

rmisio: Yeah, i didn't think that through. We could probably do without the abort check in the done handler.

PR #429, opened 2017-04-09
Fixes div closed in wrong place in settings/store

jjeffryes:
Before:
image

After:
image


rmisio:
@jjeffryes I feel like I'm at the pub playing 'Spot the differences' and I'm failing. What is the difference between those two screenshots?

jjeffryes:
@rmisio the before one is missing the gap between the save box and the box above it.

PR #437, opened 2017-04-18
Wallet transactions

rmisio:
This PR brings in the transactions list section of the wallet.


morebrownies:
Any chance you can send me some testnet coins @rmisio? 1PGhFabkipk6uHF1Y3m7bWaaM8kPREuwrn

I'm going to try sending some in too.

hoffmabc:
Hey @morebrownies do you think it might be good to have a little icon or something to indicate this button takes you out of the app to your browser?
image

Something similar to this after the words View Details?
image

Or is that too much?

morebrownies:
Yeah @hoffmabc I typically like displaying that icon on links/button when leaving the app. good catch

morebrownies:
Looking good overall @rmisio I'm waiting for some transactions to confirm, but otherwise looks good! nice work 👍

morebrownies:
Just curious, how frequently does the transaction data update within the client? I notice on one transaction I have (0 confirmations) within the client, but when I click "View details" and view the transaction details on the web via  https://testnet.blockexplorer.com, I see 2 confirmation.

rmisio:
> Any chance you can send me some testnet coins @rmisio? 

@morebrownies Just sent some.

rmisio:
> Just curious, how frequently does the transaction data update within the client? 

The client updates based on info coming from the server. With regard to confirmations, there is a walletUpdate socket that comes in every 15 seconds. I don't know how often the server checks to get that data... so you'll have to check in with @cpacia on that.

The litmus test if it's a client or server issue is if you refresh the app and the confirmations have gone up, unless a new block happened to come in at that precise moment, it's probably a client issue that I need to look into. Otherwise, it's the server not having the most up to date data.

rmisio:
> Yeah @hoffmabc I typically like displaying that icon on links/button when leaving the app. good catch

@morebrownies is that an ionic icon (i'm not seeing it there)? If not, could you provide it?

morebrownies:
Thank you. I did a hard refresh on the app and confirmations are still stuck at 0.
screenshot 2017-04-18 11 56 43

https://testnet.blockexplorer.com/tx/792814ad0b37fe9c13a1e93173439df500d138b750b3418c1a1ec8476820ef70 shows 4 now

morebrownies:
> @morebrownies is that an ionic icon (i'm not seeing it there)? If not, could you provide it?

Sure, try: ion-android-open @rmisio

rmisio:
> Thank you. I did a hard refresh on the app and confirmations are still stuck at 0.

Yeah... you could ask Chris than what the issue is... although you might get the standard 'testnet is messed up' response ;)

morebrownies:
paging @cpacia any idea why confirmations are not displaying correctly within the client?

hoffmabc:
When I send funds it goes right to 0.0
image

rmisio:
> When I send funds it goes right to 0.0

Did you send all your funds? If so, that's what it's designed to do. On the creation of an outgoing payment, the funds are immediately deducted from your confirmed balance.

hoffmabc:
No I had $120 and I sent $.70

morebrownies:
Update: I'm up to 6 confirmations now (according to blockexplorer), client displays 0 confirmations and the funds are not usable. It seems my funds are completely locked up. Is testnet really this unstable? If so, that really sucks! 

screenshot 2017-04-18 12 34 08

rmisio:
> No I had $120 and I sent $.70

Hmmm... that's not right. The balance is mostly based off a walletUpdate socket that comes every 15 seconds. Check on yours and if it's indeed coming in with zero for the confirmed amount, then you'll want to ping Chris and see what the issue could be:

image

rmisio:
> Is testnet really this unstable? If so, that really sucks!

I don't know. It hasn't been perfect for me, but it's been useable... not what you're experiencing.

morebrownies:
Thanks for adding the icon @rmisio looks good

cpacia:
One request I would make is to make the txid copy to clipboard on click. The view details button tries to load firefox for me (not my default browser) and it grinds away and is very slow. I'd really just copy and use chrome.

rmisio:
> One request I would make is to make the txid copy to clipboard on click. The view details button tries to load firefox for me (not my default browser) and it grinds away and is very slow. I'd really just copy and use chrome.

@cpacia Are you sure that Chrome is set-up as your default browser? We're using the method that the Electron docs suggest to open an external url and the docs claim it will open the URL in the user's default browser.

cpacia:
image

js/views/modals/wallet/Transaction.js
```diff
@@ -0,0 +1,198 @@
...
+    });
+  }
+
+  get $retryPmtConfirmedBox() {
+    return this._$retryPmtConfirmed ||
+      (this._$retryPmtConfirmed = this.$('.js-retryPmtConfirmed'));
+  }
+
+  remove() {
+    if (this.retryPost) this.retryPost.abort();
```
rmisio: They're aborted in the remove() of the Transactions (plural) view, which is where they originate from. I don't want the child Transaction views to abort them because, other views could theoretically be waiting on the same calls, since they are cached and potentially shared by different views.

jjeffryes:
@morebrownies once you've approved your review, we can merge this.

hoffmabc:
The issue I was seeing with the wallet is not occurring with a fresh db @rmisio @jjeffryes

morebrownies:
@jjeffryes interface wise everything is looking good. I haven't been able to get inbound payments to confirm on my end, which means I also haven't been able to test outbound payments. It might have something to do with my local setup. If you guys have those parts working on your end, then I think we should be good.

PR #443, opened 2017-04-20
Customize Checkboxes

jjeffryes:
Creates a custom style for checkboxes, and changes radio buttons and checkboxes to be visually hidden, to prevent potential DOM bugs from using display: none.

This can be further customized (it's built entirely with CSS, and is meant to match the current radio button style), but be careful with the measurements, they must use ems so they scale correctly with the text size of the label the pseudo elements that create the checkbox are in. 

**Before:**
image

**After:**
image

### DIFF
--- a/styles/_mixins.scss
+++ b/styles/_mixins.scss
@@ -78,11 +78,11 @@
   }
 
   input[type="checkbox"] + label:before {
-    color: $col2;
+    border-color: $col1;
   }
 
-  input[type="checkbox"]:checked + label:before {
-    color: $col1;
+  input[type="checkbox"] + label:after {
+    border-color: $col1;
   }
 
   .tab.active {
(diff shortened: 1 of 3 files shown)


rmisio:
@jjeffryes This PR does horrible things to scrolling in settings. Please see the attached video.
[broken-scroll.zip](https://github.com/OpenBazaar/openbazaar-desktop/files/945699/broken-scroll.zip)

morebrownies:
I'm not crazy about the style/feel of them either. Maybe we should hold off on this until we have more time.

rmisio:
> I'm not crazy about the style/feel of them either. Maybe we should hold off on this until we have more time.

I agree. This really doesn't seem like something we should be spending our cycles on at this time.

jjeffryes:
@rmisio I can't replicate the scrolling issue you're seeing, and I can't think of a way the css changes would affect the page like that. 

@morebrownies we don't have to use this. But if we want to swap out the checkbox design from the current icon based one, we'd need to do something like this (personally, I like the current icon one, and it's much more robust than this one in terms of sizing and positioning).

PR #444, opened 2017-04-20
Transactions shell

rmisio:
This pr accomplishes:
- Centralizes the mini profile found on the transactions and user page into its own view.
- Creates the initial tab structure of the Transactions page.


js/templates/miniProfile.html
```diff
@@ -0,0 +1,16 @@
+<div class="flexVCent">
+  <div>
+    <div class="userIcon disc clrBr2 clrSh1" style="<%= ob.getAvatarBgImage(ob.avatarHashes || {}) %>"></div>
+  </div>
+  <div class="flexExpand">
+    <h1 class="h2 txUnb txUnl"><%= ob.name %></h1>
+    <div class="txt5b gutterHSm">
+      <span class="clrT">
+        <%= ob.parseEmojis('📍', '', { style: 'width: 10px' }) %> <%= ob.location || ob.polyT('userPage.noLocation') %>
```
rmisio: Similar to what I've done here... we're going to want to adjust other areas of the app that use a hard-coded emoji to pipe it through twemoji. Otherwise we risk that certain OSs / browsers will not render the emoji properly.

js/views/transactions/Transactions.js
```diff
@@ -0,0 +1,113 @@
...
+        app.router.navigate(`transactions/${targ}${subRoute ? `/${subRoute}` : ''}`);
+      }
+
+      this.$('.js-tab').removeClass('clrT active');
+      this.$(`.js-tab[data-tab="${targ}"]`).addClass('clrT active');
+
+      if (this.currentTabView) this.currentTabView.$el.detach();
+
+      if (!tabView) {
+        if (this[`create${capitalize(targ)}TabView`]) {
```
jjeffryes: I don't see any create functions that would be called from this, is it just a leftover from the userpage view?

js/views/transactions/Transactions.js
```diff
@@ -0,0 +1,113 @@
...
+    let tabView = this.tabViewCache[targ];
+
+    if (!this.currentTabView || this.currentTabView !== tabView) {
+      if (opts.addTabToHistory) {
+        // subRoute is anything after the tab in the route, which is something
+        // we want to maintain, e.g:
+        // transactions/<tab>/<slug>/<blah>
+        // the subRoute is '/<slug>/<blah>'
+        const curRoute = location.hash.startsWith('#ob://') ?
+          location.hash.slice(6) : location.hash.slice(1);
```
rmisio: The address bar will always show ob://, but internally we are handling routing on either. The main reason is that I didn't want to go back in the code and have to update all our links or our manual navigates to include the preface. Do you think that would be the better way to go?

The downside of supporting both is that in places where we do need to manage / parse the route, we need to account for both of them.

js/views/transactions/Transactions.js
rmisio: Yes and I left it in because I may implement some create functions and since it gracefully handles them being there or not, I don't see too much harm in it being there.

js/views/transactions/Transactions.js
jjeffryes: I think it's fine, I was just curious about it.

js/views/transactions/Transactions.js
jjeffryes: Ok. If it's intentional then no worries.

PR #446, opened 2017-04-20
Purchasing

### LINKED ISSUE #408: Add Message Functionality to Moderator Details Modal
Now that chat is ready, wire in the message button in the Moderator Details Modal.

jjeffryes:
This adds the initial code for the purchase screen, and the moderation section. 

The moderation section has it's own view and collection, separate from the code used in the settings/store tab. The plan is to use this new code in a future refactor of the store tab. 

The moderation section has the following functionality:
- loads the listing's moderators asynchronously
- if a moderator is invalid or an error is returned for that id, the moderator is loaded in an "error" state
- the first valid moderator to load is selected
- the moderator cards have been changed to have separate view and select buttons. The buttons can have a selected, unselected (faded select) or deselected (red X) state.
- when one moderator is selected, the other moderators are deselected.
- when the moderated transaction checkbox is unchecked, the moderators are deselected and the moderator is unset from the order model.
- when the moderated transaction checkbox is checked again, any selected moderator from before it was checked is rechecked

This PR also includes css to set emojis to the same size as any parent text size class.

This PR adds in the code to make the message button in the moderator details modal work, which closes #408


jjeffryes:
@morebrownies you'll want to take a look at this, both when purchasing something and in the settings/store tab, for the tweaks to how selecting a moderator work.

js/models/Order.js
```diff
@@ -0,0 +1,51 @@
...
+    return {
+      shipTo: '',
+      address: '',
+      city: '',
+      state: '',
+      postalCode: '',
+      countryCode: '',
+      addressNotes: '',
+      moderator: '',
+      items: [{
```
rmisio: Are you planning on keeping this flat, or are we going to use a nested collection?

js/models/listing/Listing.js
```diff
@@ -253,6 +256,8 @@ export default class extends BaseModel {
     const parsedResponse = response.listing;
 
     if (parsedResponse) {
+      // set the hash
+      parsedResponse.listingHash = response.hash;
```
rmisio: Why call this listingHash and not just leave it as hash?

js/templates/moderatorCard.html
```diff
@@ -1,30 +1,57 @@
-<div class="moderatorCardInner js-moderatorCard flexRow gutterH clrP">
+<%
+   // check to see if the card was created with at least minimum data, not just a peerID, which would indicate a server error.
+   loaded = !!ob.name;
```
rmisio: Could we declare this variable with var, const or let, please?

js/utils/modalManager.js
```diff
@@ -57,6 +59,23 @@ export function launchSettingsModal(modalOptions = {}) {
   return aboutModal;
 }
 
+export function launchPurchaseModal(modalOptions = {}) {
```
rmisio: Can you just launch this modal from the listing details view? Modal manager is more designed for singleton modals where an instance is shared app wide.

This guy could really just be launched from wherever it's CTA is. Please be sure to remove the modal in the remove of the view that launches it.

moderatorDetailsModal would fall into the same category, but I think we could leave it be for now. It wasn't a 100% clear to me the exact role of modalManager, so I didn't flag it earlier.

js/views/modals/listingDetail/Listing.js
```diff
@@ -307,6 +308,22 @@ export default class extends BaseModal {
...
+    const selectedVariants = [];
+    this.variantSelects.each((i, select) => {
+      const variant = {};
+      variant.name = $(select).attr('name');
+      variant.value = $(select).val();
+      selectedVariants.push(variant);
+    });
+
+    this.purchaseModal = launchPurchaseModal({
+      listing: this.model.toJSON(),
```
rmisio: I don't have enough context to give a strong opinion on this... but why are you passing in the flat JSON of the listing as opposed to the whole listing model? The listing model has some utility functions and I would think having access to its full API would be more efficient.

Am I missing something.

test/models/listing/listing.js
```diff
@@ -198,83 +207,86 @@ describe('the Listing model', () => {
...
 
     expect(valErr && valErr.termsAndConditions &&
-      !!valErr.termsAndConditions.length || false).to.equal(true);
+        !!valErr.termsAndConditions.length || false).to.equal(true);
   });
 
   it('fails validation if, for a physical good, at least one shipping options is not provided',
-    () => {
-      const listing = new Listing();
+      () => {
```
rmisio: Why all this extra spacing? Is there some convention I'm missing?

test/models/listing/listing.js
```diff
@@ -198,83 +207,86 @@ describe('the Listing model', () => {
...
-        },
-        coupons: [
-          {
-            discountCode: Date.now() + Math.random(),
-            priceDiscount: 500, // should not fail validation
+        listing.set({
+          item: {
+            price: 500,
           },
-          {
```
rmisio: I don't understand what's happening here? Is spacing really resulting in such a dramatic diff?

js/models/Order.js
jjeffryes: I'll put in a nested collection. It's a little weird now, the protocol allows multiple items but there's no way to actually do that in the UI design.

js/models/listing/Listing.js
jjeffryes: I like to be explicit. I can keep it as hash if you want.

js/templates/moderatorCard.html
jjeffryes: Yes, that was just a typo.

js/utils/modalManager.js
jjeffryes: Ah, ok. I thought it was better to run everything through the modal manager, so all modals were handled the same way. That way we could always handle positioning and layering the same.

js/views/modals/listingDetail/Listing.js
jjeffryes: I only needed the data for the template. But you're right, there could be a case where the other functions are useful, and it can be converted to JSON in the render.

test/models/listing/listing.js
jjeffryes: Yeah, I think this is all spacing. My spacing settings must be different than yours, and my IDE is updating everything when I edit a file. Did yours change? I don't think I've changed mine, unless it happened when WebStorm updated.

js/models/listing/Listing.js
jjeffryes: That, and it's called listingHash in the order.item, which is the only thing it's used for here.

test/models/listing/listing.js
rmisio: ok, I guess it's fine.

test/models/listing/listing.js
rmisio: I couldn't quite grasp what it was changing there... but if it's just spacing it's fine. Our linter is ok with it, so I could live with it as well.

js/models/listing/Listing.js
rmisio: ok, it's not such a huge deal, leave it be.

But for consistency can we preface all the other listing attributes with listing as well (listingCategory, listingSlug, etc...)? ;)

rmisio:
The modal needs to handle the case of a listing having no moderators, whether that's not making the Moderated Payment checkbox and Moderator box visible or offering some type of messaging (or something else).

image

jjeffryes:
@rmisio this is ready for re-review, all the requested changes have been made, plus I've broken out the models with sub-collections and models, and added some CSS so emojis are sized according to the text class that affects them automatically.

jjeffryes:
@morebrownies we still need you to take a look at this, specifically the changes to how the moderator cards work.

morebrownies:
@jjeffryes I'm attempting to add a moderator to my store and I'm running into this issue when I hit save:



Do you have a store setup with moderators already added that I can test?

jjeffryes:
@morebrownies your screenshot doesn't look like it loaded. What issue are you hitting with save?

morebrownies:
screenshot 2017-04-24 09 49 59

jjeffryes:
@morebrownies well, that's not a very informative error. Have you cleared your data recently? A number of things have changed. 

You can try clicking purchase on ob://QmYcZjTpcpqnMpdPSyz27zVNkfWNsGaFWByAAu1pUBBQaN/store/test

morebrownies:
I cleared my db this morning. I'll clear it again and see if I get the same error.

morebrownies:
Ok, issue fixed after reseting my DB. 

Anyways, I think it's mostly working pretty well now. I do find it a bit strange that there's a big checkbox button next to each moderator in the moderator list. That checkbox button is really functioning like a radio button and would probably be less confusing if we used a radio button instead. 

@jjeffryes I know you mentioned you coded it this way so that you could reuse logic between views. That's probably fine in the short term, but we should circle back at some point to clean this 

**Current**
screenshot 2017-04-24 14 33 42

**Design**
screenshot 2017-04-24 14 38 34

jjeffryes:
@morebrownies If we switched it to a radio button, it would be a radio button in the settings/store tab too.  Is that ok?

js/views/modals/listingDetail/Listing.js
```diff
@@ -307,6 +309,32 @@ export default class extends BaseModal {
...
+  startPurchase() {
+    const selectedVariants = [];
+    this.variantSelects.each((i, select) => {
+      const variant = {};
+      variant.name = $(select).attr('name');
+      variant.value = $(select).val();
+      selectedVariants.push(variant);
+    });
+
+    if (this.purchaseModal) {
```
rmisio: Since we're removing it on close, I don't think this case will ever be hit... but up to you if you want to leave it in.

rmisio:
> @morebrownies If we switched it to a radio button, it would be a radio button in the settings/store tab too. Is that ok?

@jjeffryes I think Mike was fine with circling back to it. When we do, there's no reason we can't re-use the same view, but have slight differences in look / feel and behavior. Assuming there's enough commonality in the behavior that we would want to re-use (which I guess there would be based on how the content is loaded... is that correct?), we could just pass in some flags to use a radio or a checkbox. A radio selected would send a different event than the checkbox being selected, so the consumer could customize the behavior based off of that.

morebrownies:
> @morebrownies If we switched it to a radio button, it would be a radio button in the settings/store tab too. Is that ok?

no that wouldn't work. Settings should support selecting many (checkboxes) while checkout should only allow you to select one (radio buttons)

PR #447, opened 2017-04-21
Category bug fix

jjeffryes:
When the data for a listing is returned from the server, the categories can have a value of null, which breaks code that is expecting an array. 

This will check the categories and use a blank array if the value is falsey.

### DIFF
--- a/js/models/listing/ListingShort.js
+++ b/js/models/listing/ListingShort.js
@@ -22,6 +22,15 @@ export default class extends BaseModel {
     return this.get('shipsTo').indexOf(country) !== -1;
   }
 
+  parse(response) {
+    const parsedResponse = response;
+
+    parsedResponse.categories = Array.isArray(parsedResponse.categories) ?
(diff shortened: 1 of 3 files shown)


jjeffryes:
This fix has been updated to use parse, which is much better. 

This PR only needs to be merged if the server can't be updated to not send null for the categories value.

js/models/listing/ListingShort.js
```diff
@@ -22,6 +22,15 @@ export default class extends BaseModel {
     return this.get('shipsTo').indexOf(country) !== -1;
   }
 
+  parse(response) {
+    const parsedResponse = response;
```
rmisio: There's no point in setting parsedResponse to response. Objects are assigned by reference, so later when you change parsedResponse, you're also changing response.

If you want to modify a separate object (at least at the top level), then you would want:

```
const parsedResponse = { ...response };
```

Otherwise, just modify response directly.

PR #450, opened 2017-04-24
Keep Radio Button Labels Same Size

### LINKED ISSUE #434: Vertical Spacing in Filter
image

jjeffryes:
If the radio button label on a filter has no text, use min-height to keep it the same height as one with text.
Closes #434


jjeffryes:
Requesting @hoffmabc for a review here, since you spotted the bug.

Issue #463, opened 2017-04-26
Shorter default names


JustinDrake:
When a new node is created and no name is specified, it looks like the reference client is creating a random name of the from `ob ` + a random string.

![screen shot 2017-04-26 at 13 18 26](https://cloud.githubusercontent.com/assets/731743/25434116/f13848fc-2a82-11e7-882e-962cd5f9c1d9.png)

The random string is very long, so much so that we had to clip it on www.openbazaar.chat. We would appreciate it if the random part could be kept short, e.g. 6 characters long. The random string uses lowercase letters and digits, so 6 characters yields (26+10)^6 = 2,176,782,336 possibilities (two billion!).

hoffmabc:
I think that's probably reasonable barring any technical reason for the extra long string. @cpacia?

JustinDrake:
Cool. While we're at it maybe replace `ob ` by `OB ` (uppercase) 👍

jjeffryes:
There's no particular technical reason for it, it's arbitrary. We picked something long because people can type in long names, and any system showing names should be able to display a name at least that long, or handle the overflow gracefully (clipping is a perfectly acceptable solution for that).

JustinDrake:
I think this has been fixed. Thanks :)

PR #464, opened 2017-04-26
Change random profile name algo

hoffmabc:
Accomodates #463

### DIFF
--- a/js/models/profile/Profile.js
+++ b/js/models/profile/Profile.js
@@ -13,7 +13,7 @@ export default class extends BaseModel {
       handle: '',
       location: '',
       moderator: false,
-      name: `OB ${Math.random().toString(36).slice(1).substring(1,7)}`,
+      name: `OB ${(Math.floor(Math.random() * 2116316159) + 60466176).toString(36)}`,
       nsfw: false,
       shortDescription: '',
       avatarHashes: new Image(),
--- a/js/models/profile/Profile.js
+++ b/js/models/profile/Profile.js
@@ -13,7 +13,7 @@ export default class extends BaseModel {
       handle: '',
       location: '',
       moderator: false,
-      name: `ob ${Math.random().toString(36).slice(2)}`,
+      name: `OB ${Math.random().toString(36).slice(1).substring(1,7)}`,
       nsfw: false,
       shortDescription: '',
       avatarHashes: new Image(),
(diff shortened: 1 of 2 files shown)


JustinDrake:
Thanks @hoffmabc. Technically `Math.random()` is 0 inclusive, so this doesn't quite work. Should be `OB ${Math.round((Math.pow(36, 6 + 1) - Math.random() * Math.pow(36, 6) - 1)).toString(36).slice(1)` (with a ` - 1`)

hoffmabc:
I decided that a simpler approach might be to just substring out this original algo for the first 6 chars. Do you guys see an issue with that?

JustinDrake:
@hoffmabc Doesn't quite work. To guarantee a minimum length of 6 characters you could add padding

```
(Math.random().toString(36)+'0000000').slice(2, 8)
```

js/models/profile/Profile.js
```diff
@@ -13,7 +13,7 @@ export default class extends BaseModel {
       handle: '',
       location: '',
       moderator: false,
-      name: `ob ${Math.random().toString(36).slice(2)}`,
+      name: `OB ${Math.random().toString(36).slice(1).substring(1,7)}`,
```
jjeffryes: You don't need the slice and the substring, the slice is cutting off the first character (which is a 0), then the substring is cutting off the first character again (which is now a .). 

You could do the same thing with `OB ${Math.random().toString(36).substring(2,8)}`

There is a chance the random number won't have enough decimals (0.2, for instance), if we care about that we could do a little more here to fix it.

jjeffryes: A more robust way to do it would be:
```
`OB ${(Math.random() * 1000000000).toString(36).split('.')[0]}`
```
That should provide a 6 character random string, with no period.

JustinDrake: @jjeffryes Your solution fails. E.g. `(0 * 1000000000).toString(36).split('.')[0]` returns "0"

jjeffryes: "OB 0" is a valid result, so that's working as intended. I suppose I should have said "should usually give you 6 characters, but has a small chance of being less."

If we want to guarantee 6 characters, we can do this:

```
`OB ${(Math.floor(Math.random() * 2116316159) + 60466176).toString(36)}`
```

Floor here is nicer than cutting it at the period, and this should generate a string between zzzzzz (the highest 6 character number in base 36) and 100000 (the lowest).

That's just off the top of my head, so suggestions on improving it are welcome.

PR #467, opened 2017-04-27
Profile url updates

rmisio:
- Updates the profile model to only use our local url. Previously it had been getting other's profiles via ipns.
- Updates the async profiles fetch in chat to include the usecache=true flag (which is ignored now, but will be used once the [profile caching server PR](https://github.com/OpenBazaar/openbazaar-go/pull/519) is merged)

### DIFF
--- a/js/models/profile/Profile.js
+++ b/js/models/profile/Profile.js
@@ -26,7 +26,7 @@ export default class extends BaseModel {
   }
 
   url() {
-    return app.getServerUrl(`ob/profile/${app.profile.id}`);
+    return app.getServerUrl(`ob/profile/${this.id}`);
   }
 
   // todo: set peerId instead of ID when setting ID.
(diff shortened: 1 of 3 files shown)


js/models/profile/Profile.js
```diff
@@ -26,9 +26,7 @@ export default class extends BaseModel {
   }
 
   url() {
-    // url is handled by sync, but backbone bombs if I don't have
-    // something explicitly set
-    return 'use-sync';
+    return app.getServerUrl(`ob/profile/${app.profile.id}`);
```
jjeffryes: Shouldn't this be ${model.id}?

Currently, it looks like it gets your own profile for all profile calls.

rmisio: Yup, good catch. Fixed.

I also updated the Listings collection url.

PR #469, opened 2017-04-28
Purchasing Update

jjeffryes:
This PR stubs in shipping and makes a call to the purchase API for testing purposes.
- shipping is just an empty view and template at this point
- the order model is working
- the call to purchase only works for non-physical listings, since the shipping code isn't done. It outputs the results in the console.
- an update was made to the listing model url so it uses the right API to retrieve a listing hash.
- the moderators view as updated to have a method for retrieving the IDs of currently selected moderators. This allows the purchase view to pull that data when the order is finalized instead of tracking changes made to which moderator is selected. This will also be very useful if the settings store tab is refactored with the new moderators view.
- the confirmation tooltip was added.

### DIFF
--- a/js/models/listing/Listing.js
+++ b/js/models/listing/Listing.js
@@ -136,7 +136,7 @@ export default class extends BaseModel {
           app.getServerUrl(`ob/listing/${slug}`);
       } else {
         options.url = options.url ||
-          app.getServerUrl(`ipns/${this.guid}/listings/${slug}.json`);
+          app.getServerUrl(`ob/listing/${this.guid}/${slug}`);
       }
     } else {
       options.url = options.url || app.getServerUrl('ob/listing/');
--- a/js/views/modals/purchase/Purchase.js
+++ b/js/views/modals/purchase/Purchase.js
@@ -81,6 +81,8 @@ export default class extends BaseModal {
       'click .js-payBtn': 'clickPayBtn',
       'click .js-pendingBtn': 'clickPendingBtn',
       'change #purchaseQuantity': 'changeQuantityInput',
+      'click .js-confirmPayConfirm': 'clickConfirmBtn',
+      'click .js-confirmPayCancel': 'closeConfirmPay',
       ...super.events(),
     };
   }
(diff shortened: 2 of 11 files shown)


js/models/listing/Listing.js
```diff
@@ -136,7 +136,7 @@ export default class extends BaseModel {
           app.getServerUrl(`ob/listing/${slug}`);
       } else {
         options.url = options.url ||
-          app.getServerUrl(`ipns/${this.guid}/listings/${slug}.json`);
+          app.getServerUrl(`ob/listing/${this.guid}/${slug}`);
```
jjeffryes: The url was updated here so it would add the listing hash to the model.

js/views/modals/purchase/Purchase.js
```diff
@@ -128,28 +122,37 @@ export default class extends BaseModal {
...
-    console.log(this.order.attributes);
+  closeConfirmPay() {
+    this.$confirmPay.addClass('hide');
+  }
+
+  purchaseListing() {
+    // set the moderator
+    this.order.set('moderator', this.moderators.selectedIDs[0]);
+
+    $.post({
```
jjeffryes: This section is to test whether the order model data can be used to call the purchase api.

js/views/modals/purchase/Shipping.js
```diff
@@ -0,0 +1,43 @@
...
+    this.options = options;
+
+    if (!this.model || !(this.model instanceof Listing)) {
+      throw new Error('Please provide a listing model');
+    }
+
+    this.listenTo(app.settings.get('shippingAddresses'), 'update',
+      (cl, updateOpts) => {
+        if (updateOpts.changes.added.length ||
+          updateOpts.changes.removed.length) {
```
rmisio: As an FYI...this condition will account for a shippingAddress being added or removed. If you also want to account for an existing shippingAddress changing... I believe that information would be in `updateOpts.changes.merged`. It may be easier to just re-render the view on update depending on how the view keeps track of it's own state and how grace-fully it re-renders (it also may be easier to do the updates with jQuery - all depends on how you architect the view).

jjeffryes: I don't think I will call save on the order model, since the server returns data that shouldn't go into the order model. 

But I will be using the order model to render the receipt (right side bar with the prices), so it seemed like it was worth it to make it a model, so it's easy to rerender the receipt on changes etc.

jjeffryes: Good tip on the address changes, I'm working on that now.

PR #470, opened 2017-05-01
Purchasing - Shipping

jjeffryes:
This PR adds shipping to the purchase view.


js/views/modals/Settings/Settings.js
```diff
@@ -30,6 +30,8 @@ export default class extends BaseModal {
       Moderation,
     };
 
+    this.initTab = this.tabViews.hasOwnProperty(opts.initTab) ? opts.initTab : 'General';
```
jjeffryes: The initTab option allows another view to open the settings modal to a specific tab. It's used here to open the addresses tab.

js/views/modals/purchase/Shipping.js
```diff
@@ -12,15 +14,11 @@ export default class extends BaseModal {
...
 
-    this.listenTo(app.settings.get('shippingAddresses'), 'update',
-      (cl, updateOpts) => {
-        if (updateOpts.changes.added.length ||
-          updateOpts.changes.removed.length) {
-          // update the shipping section with the changed address information
-          // TODO: add shipping code here
-        }
-      });
+    this.listenTo(app.settings.get('shippingAddresses'), 'update', () => this.render());
```
jjeffryes: If any changes are made to the shippingAddresses collection, this view will re-render. This means the state of the selection is lost, but that is unavoidable without adding some complexity since the user might have deleted the currently selected address, or changed the order of the addresses.

js/models/purchase/Order.js
```diff
@@ -23,8 +23,21 @@ export default class extends BaseModel {
     };
   }
 
+  addAddress(sAddr) {
+    // this will convert and set an address from the settings
```
jjeffryes: This allows an address from the settings to be converted and added to the order model in the proper format.

jjeffryes:

jjeffryes:
I noticed a bug in the shipping option template.

jjeffryes:
The missing price formatting in the shipping option is fixed now.

js/views/modals/purchase/Purchase.js
```diff
@@ -140,19 +148,37 @@ export default class extends BaseModal {
...
-        console.log(data);
+        .done((data) => {
+          console.log(data);
+        })
+        .fail((data) => {
+          console.log(data);
+        });
+    } else {
+      let errMsg = '';
+      _.each(this.order.validationError, (obj) => {
```
rmisio: This approach is a bit of a deviation from how we display errors elsewhere, which is inline above the respective field that has an error. I suppose if the errors here are quite minimal, we could probably get away with displaying them all in a modal. But on more traditional forms, for consistency (and IMHO, better UX), I would do them inline.

Anyways... have you tested this with some actual errors? I don't think this would work because the error object is not a list, it's an object, for example:

```
{
  fieldName: ['error1', 'error2'...],
  fieldName2: ['error1'],
}
```

You'd probably want something like:

```
Object.keys(this.order.validationError)
  .forEach(errKey => {
    this.order.validationError[errKey].forEach(err => {
      errMsg += `${errKey} - ${err}<br>`;
    });
  });
```

rmisio: Why use underscore utility functions when native JS has a well supported equivalent (in this case forEach, which is supported all the way back to IE9)?

rmisio:
If I select a Shipping Address with no name, the ShippingOptions view bombs:

image

jjeffryes:
@rmisio how are you using a shipping address with no name? The name is required when you save the address.

rmisio:
> how are you using a shipping address with no name? The name is required when you save the address.

Oh sorry, it did have a name. I just wiped my data, so I don't have the details, but I'm pretty sure I gave it the minimum data. Anyhow, the error is bombing because whatever object name is supposedly on is undefined.

I think what I was doing was purchasing when I had a single address - click add new address - add a new one and then back on the Purchase modal, select the new one I just created. That was causing it to bomb.

jjeffryes:
> I think what I was doing was purchasing when I had a single address - click add new address - add a new one and then back on the Purchase modal, select the new one I just created. That was causing it to bomb.

I changed some things related to that in my current branch, which may have fixed it, as I can't replicate it with my current code (which will be in the next PR shortly). 

We can probably close this one when I submit the next PR, since it polishes some things along with adding the receipt code.

jjeffryes:
re: the errors, my thinking was a) other than quantity, it would be really unusual to actually have an error, and b) I was being a little lazy.

The _.each is because it's an object instead of an array, it worked as expected when I tested it. I'll tinker with getting the errors into places that make more sense/

PR #471, opened 2017-05-02
WIP - Transactions purchases

rmisio:
This PR brings in the functionality for the Purchases tab on the Transactions page.

Notes:
- This should be run in conjunction with this [server PR](https://github.com/OpenBazaar/openbazaar-go/pull/515), although that PR is likely in flux. When that PR changes, this PR will need to be adjusted.
- I've hidden the sort icon on the Date column since it looks like we'll be moving to a Sort drop down.
- The listing column is not clickable at this moment. I'm waiting for some back-end data, after which it will link to the listing detail page.

PR #472, opened 2017-05-03
Purchasing - Shipping and Receipt

jjeffryes:
- this adds the receipt, in addition to the shipping code from the closed #470 
- the receipt is shown and updated in real time.
- errors have been refactored to show in the appropriate spot in the DOM when a model error is thrown.
- the disclaimers and notes were added under the action button in the right sidebar.
- validation was added to check if the user has a shipping option when it is needed, to address a situation where the user has no addresses, adds one, then opens settings again and deletes it.
- the way the shipping options are applied to the order model was refactored to happen whenever a selection is made, so the shipping price can be added to the receipt immediately.


js/templates/modals/purchase/purchase.html
```diff
@@ -19,6 +19,8 @@
     <div class="flexColRow gutterVMd">
       <% // to support multiple items in a purchase, loop this section %>
       <section class="contentBox pad clrP clrBr clrSh3">
+        <div class="js-errors"></div>
```
jjeffryes: This is the generic holder for errors that aren't associated with a specific section of the template.

js/models/purchase/Order.js
```diff
@@ -23,30 +23,31 @@ export default class extends BaseModel {
     };
   }
 
+  addAddress(sAddr) {
```
jjeffryes: The address from settings is not in the same format as the address needed here, so this method will translate it and add it to the order.

js/views/modals/purchase/Purchase.js
```diff
@@ -140,19 +166,51 @@ export default class extends BaseModal {
   }
 
   purchaseListing() {
+    // if the listing has shipping, set it
+    if (this.listing.get('shippingOptions') && this.listing.get('shippingOptions').length) {
```
jjeffryes: It's possible for the user to delete all of their addresses in settings before trying to make a purchase, so there's a check here to make sure that doesn't happen. The order model can't do the checking because it doesn't know if the order needs an address or not, so it always treats the address data as optional.

js/views/modals/purchase/Purchase.js
```diff
@@ -140,19 +166,51 @@ export default class extends BaseModal {
...
+        .done((data) => {
+          console.log(data);
+        })
+        .fail((data) => {
+          const response = data && data.responseText || '';
+          const errMsg = response ? JSON.parse(response).reason : '';
+          const errTitle = app.polyglot.t('purchase.errors.orderError');
+          openSimpleMessage(errTitle, errMsg);
+        });
+    } else {
```
jjeffryes: This code loops through any model errors, and either displays them in the appropriate container or displays them in a generic container.

js/views/modals/purchase/Receipt.js
```diff
@@ -0,0 +1,56 @@
...
+    this.listenTo(this.model.get('items').at(0), 'change', () => this.render());
+    this.listenTo(this.model.get('items').at(0).get('shipping'), 'change', () => this.render());
+  }
+
+  className() {
+    return 'receipt flexColRows gutterVSm';
+  }
+
+  get prices() {
+    // create an array of price objects that matches the items in the order
```
jjeffryes: This is an array because it's an array in the order model, to support future code that has multiple items in each order.

morebrownies:
I've been poking around trying to generate some test transactions with this PR, but none of them seem to appear in my Purchases section :(

rmisio:
> I've been poking around trying to generate some test transactions with this PR, but none of them seem to appear in my Purchases section :(

Once you create the purchases using this PR, are you switching over to the [other PR](https://github.com/OpenBazaar/openbazaar-desktop/pull/471) to try and view them?

morebrownies:
Ohhhh no I didn't switch. Let me give that a try.

morebrownies:
Still no luck. I may need to blow away my db again. Trying that now.

morebrownies:
Still no luck after blowing away my db. Sorry guys

js/views/modals/purchase/Purchase.js
```diff
@@ -38,10 +42,14 @@ export default class extends BaseModal {
     // add the item to the order.
     this.order.get('items').add(item);
 
+    this.receipt = this.createChild(Receipt, {
```
rmisio: The problem with declaring your child views in the constructor (essentially caching them between renders) is that when you re-render, the event handlers don't work anymore. At the top of my head, I don't remember exactly the reason why.

Anyways... if you change the currency in settings and get the 'Data change <refresh>' popup, after clicking refresh, the CTAs from any child views won't work (e.g. selecting a different shipping option will not update the receipt).

I don't run into this issue because I declare the child views in render (be sure to remove them first in case they already exist).

jjeffryes: It happens because the events need to be re-delegated when you append a new render, which I forgot about. 

I'm refactoring the action (pay, etc.) button to be a separate view and ended up moving all the sub-views to being created in the render, which should fix this.

jjeffryes:
@rmisio I've updated this PR, with fixes from the code review, and some bug fixes.

The biggest change was I broke the action buttons (pay, pending, close) into a sub view to make a cleaner solution to some of the things that came up in the code review.

Changes from the code review:
- improved the error validation in the item model.
- clear errors each time the purchase is validated.
- moved the sub view creation into the render.
- fixed the quantity being set as a string.

New Changes:
- added a check to make sure the order has a shipping sub model set if it is needed.
- refactored the action button to be a sub view.
- added the various notes based on the action button state.
- removed validation from the Option model, because options are optional and the attributes may be blank.
- removed validation from the Shipping model, because shipping is optional, and the attributes may be blank.
- fixed copy/paste error where the BaseModal view was used instead of the BaseVw in the sub views.
- fixed an issue where if no valid shipping option was available, the old shipping option would be retained.

rmisio:
Clicking the Close button doesn't do anything. Is that coming later?

Also, looks like there's an extraneous space after the quantity in the receipt.

image

rmisio:
May very well be working as expected, but quantity doesn't effect the shipping price. Is that correct that it would be the same price to have 10 items shipped versus 1?

js/views/modals/purchase/Purchase.js
```diff
@@ -123,40 +107,84 @@ export default class extends BaseModal {
...
+          dataType: 'json',
+          contentType: 'application/json',
+        })
+          .done((data) => {
+            this.state.phase = 'pending';
+            this.actionBtn.render();
+            console.log(data);
+          })
+          .fail((jqXHR) => {
+            const errMsg = jqXHR.responseJSON ? jqXHR.responseJSON.reason : '';
```
rmisio: I proposed we use this:
`errMsg = jqXhr.responseJSON && jqXhr.responseJSON.reason || '';`

and you put in this:
`errMsg = jqXHR.responseJSON ? jqXHR.responseJSON.reason : '';`

It's not a massive deal, but the benefit of the first version is that if responseJSON or reason are not defined, it will set the error message to an empty string.

In the second version, if reason is not defined, you will set the error message to undefined. I think openSimpleMessage will handle that gracefully, so it's up to you if you want to change this.

FWIW, 'reason' is a custom field provided by the server. If the request fails before getting to the server, reason won't be there.

jjeffryes:
The close button is non-functional currently.

You're right on shipping, the api multiplies it on purchase.

js/views/modals/purchase/ActionBtn.js
```diff
@@ -0,0 +1,93 @@
...
+      (this._$payBtn = this.$('.js-payBtn'));
+  }
+
+  get $confirmPay() {
+    return this._$confirmPay ||
+      (this._$confirmPay = this.$('.js-confirmPay'));
+  }
+
+  remove() {
+    $(document).off('click', this.onDocumentClick);
```
rmisio: This is a bug I need to also fix in Configuration.js (where you got the code), but bind() returns a new function and that is actually what need to be unbound:

// in the constructor
this.boundOnDocClick = this.onDocumentClick.bind(this);
$(document).on('click', this.boundOnDocClick);

// in remove
$(document).off('click', this.boundOnDocClick);

otherwise this doc click handler is just going to perpetually live and a new one will keep getting applied (and keep on living) every time the Purchase modal is opened.

rmisio:
> You're right on shipping, the api multiplies it on purchase.

Shouldn't we multiply it on the receipt then?

jjeffryes:
> Shouldn't we multiply it on the receipt then?

Yes, I was agreeing with you. ;)

I have it fixed, along with a formatting issue I noticed, I'll push it when you are done reviewing.

PR #473, opened 2017-05-04
WIP - Transactions - Purchases and Sales tab

rmisio:
This PR implements the Purchases and Sales tab on the Transactions page.

Notes:
- This should be run in conjunction with this [server PR]
(https://github.com/OpenBazaar/openbazaar-go/pull/515).
- You will need Purchase / Sales data to properly test this PR. It can be obtained via directly using the ob/purchase API or via buying a listing via this [client PR](https://github.com/OpenBazaar/openbazaar-desktop/pull/472).
- Since there are API changes coming, this PR is a WIP... It's being posted so any reviewers who want a head start can do so.


js/collections/Transactions.js
```diff
@@ -1,22 +1,83 @@
...
+      timestamp: '2017-04-25T09:28:20-07:00',
+      title: 'try again',
+      total: 156606,
+      unreadChatMessages: 0,
+      vendorHandle: '',
+      vendorId: 'QmbSYKcypVVZNVQ4tqgG6oE8kNBapjgoU5676fBp62hcWE',
+    };
+  }
+
+  // temporary to test pagination with dummy data
```
rmisio: Leaving this in temporarily so that I could easily create dummy data to test pagination.

js/views/transactions/table/Table.js
```diff
@@ -0,0 +1,475 @@
...
+        this.indexedViews.byUser[view.model.get('vendorId')] = view;
+        this.indexedViews.byUser[view.model.get('buyerId')] = view;
+      }
+
+      this.indexedViews.byOrder[view.model.id] = view;
+    });
+  }
+
+  get transactionsPerPage() {
+    return 5;
```
rmisio: For testing I've kept this quite low, but will increase before final merge.

js/templates/transactions/table/row.html
```diff
@@ -0,0 +1,115 @@
...
+%>
+
+<td class="clrBr orderCol noOverflow">
+  <div class="unreadBorder clrE1"></div>
+  <%= ob.orderId %>
+</td>
+<td class="clrBr dateCol"><%= ob.moment(ob.timestamp).format('l LT') %></td>
+<td class="clrBr listingCol">
+  <div class="flexVCent gutterHSm">
+    <span class="thumb" style="background-image: url(http://localhost:4002/ipfs/Qmc2R67eBoY7u7Y5oV7yjVy8ivTxhVi443yJXr1dkbQhLo), url('../imgs/defaultItem.png')"></span>
```
jjeffryes: The hard coded thumbnail is just a placeholder for now?

js/utils/currency.js
```diff
@@ -153,7 +153,7 @@ export function formatCurrency(amount, currency, options) {
     // out with the appropriate Bitcoin symbol
     const formattedAmount = new Intl.NumberFormat(opts.locale, {
       minimumFractionDigits: 0,
-      maximumFractionDigits: 8,
+      maximumFractionDigits: 6,
```
rmisio: I think using 8 decimal places is problematic in a lot of layouts, especially tabular ones with columns with limited space. @morebrownies was even suggesting going down to 4 (that's what a lot of the designs have). I think 4 is little too low, so I though 6 could be a happy medium.

Another alternative, is leaving the default as 8 and then for certain situations were space is at a premium, passing in a smaller value. But... wouldn't we want to be consistent across the app? (plus the vast majority of places it's used in the UI have space constraints)

js/views/transactions/Tab.js
```diff
@@ -0,0 +1,217 @@
...
+  events() {
+    return {
+      'change .filter input': 'onChangeFilter',
+      'keyup .js-searchInput': 'onKeyUpSearch',
+    };
+  }
+
+  onChangeFilter() {
+    let state = [];
+    this.filter.state = this.$filterCheckboxes.filter(':checked')
```
rmisio: good catch. this went through a refactor or two, so that was probably lost in the shuffle.

js/templates/transactions/table/row.html
```diff
@@ -0,0 +1,116 @@
...
+%>
+
+<td class="clrBr orderCol noOverflow">
+  <div class="unreadBorder clrE1"></div>
+  <%= ob.orderId %>
+</td>
+<td class="clrBr dateCol"><%= ob.moment(ob.timestamp).format('l LT') %></td>
+<td class="clrBr listingCol">
+  <div class="flexVCent gutterHSm">
+    <% const imageHashes = { small: ob.thumbnail, tiny: ob.thumbnail } %>
```
rmisio: This is a little ugly, but IMO getListingBgImage needs to be extracted to be used for different listing image sizes. Maybe it would accept a target width and height and then it would select the appropriate image hash based off of that as opposed to what it's doing now, which is flat out using small for hiRez and tiny otherwise.

It would mean the responsive module would need to know what the different sizes are (e.g. small is X by X, whereas tiny is X by X, etc...), but I think that's ok.

In any case, I think this dirty (super easy) approach I used, for now, is probably ok.

PR #476, opened 2017-05-05
added a confirm box to the deletion of a listing from the listing de…

### LINKED ISSUE #414: Delete Button Needs a Confirm Step/State
This delete button just goes ahead and nukes the listing with no confirmation. Probably not a great approach unless we can back out of it somehow.

image

rmisio:
…tail modal

closes #414

### DIFF
--- a/js/templates/modals/listingDetail/listing.html
+++ b/js/templates/modals/listingDetail/listing.html
@@ -21,12 +21,12 @@
     </div>
   </div>
   <div class="js-deleteConfirmedBox confirmBox deleteConfirm tx5 arrowBoxTop clrBr clrP clrT hide">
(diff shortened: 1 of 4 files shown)


js/templates/modals/listingDetail/listing.html
```diff
@@ -20,6 +20,15 @@
       }) %>
     </div>
   </div>
+  <div class="js-deleteConfirmedBox confirmBox deleteConfirm tx5 arrowBoxTop clrBr clrP clrT hide">
+    <div class="tx3 txB rowSm">Are you sure?</div>
+    <p>Once it's deleted, it's gone forever.</p>
```
jjeffryes: The text in the confirmation box should be translated.

PR #481, opened 2017-05-11
Purchasing coupons

jjeffryes:
This PR adds coupons, the memo field, and the alternate contact field.

It also refactors the handling of shipping errors to take place in the models.

This PR builds on #472, it will remain a WIP until that PR is merged.


jjeffryes:
@rmisio this PR is ready.

jjeffryes:
Found a bug.

js/views/modals/purchase/Purchase.js
```diff
@@ -114,29 +130,46 @@ export default class extends BaseModal {
...
+  blurEmailAddress(e) {
+    this.order.set('alternateContactInfo', $(e.target).val());
+  }
+
+  blurMemo(e) {
+    this.order.set('memo', $(e.target).val());
+  }
+
+  changeCoupons() {
+    this.receipt.coupons = this.coupons.couponHashes;
```
rmisio: Since this is the only place you're using the couponHashes, a more robust way to get them here would be to pass them into the event rather than relying on a variable attached to the instance.

```
this.trigger('changeCoupons', this.couponHashes);

changeCoupons(couponHashes) { ... }
```

js/views/modals/purchase/Coupons.js
```diff
@@ -0,0 +1,119 @@
...
+        }
+      } else {
+        this.invalidCode = { type: 'Invalid', code };
+      }
+      this.render();
+    });
+  }
+
+  findCoupon(hashedCode, code) {
+    let coupon = this.coupons.findWhere({ hash: hashedCode });
```
rmisio: Why not make this all one line?

```
return this.coupons.findWhere({ hash: hashedCode }) ||
  this.coupons.findWhere({ title: code });
```

rmisio:
Looks like we could use some padding above the memo:

image

js/languages/en-US.json
```diff
@@ -839,14 +852,15 @@
...
+      "coupon": "Coupon",
       "shipping": "Shipping",
       "total": "Total"
     },
     "errors": {
       "moderatorsTitle": "Moderator Data Failted to Load",
       "moderatorsMsg": "This listing has availble moderators, but they could not be loaded.",
       "orderError": "Your order can not be processed due to the following errors.",
-      "missingAddress": "You must choose a shipping address to complete your purchase."
+      "missingAddress": "Please choose a valid shipping address."
```
jjeffryes: Yep. Good catch.

jjeffryes:
The latest PR fixes the issues raised in the code review, and also:
- shows an error modal if the user attempts to buy their own listing.
- allows a user to test their own coupons.
- adds a moderated boolean to the order model, if true, the order will validate if it has a value for the moderator
- fixes a bug where not all the errors were cleared when purchasing.
- refactors the error codes in the coupons to be cleaner.
- clears valid results from the coupon input, but not invalid results

@rmisio this is ready to be re-reviewed.

js/languages/en-US.json
```diff
@@ -806,6 +806,7 @@
   },
```
rmisio: I don't think it's properly removing coupons internally.

- add a coupon for 50% off
- add a coupon for 51% off

Correctly, gives you an error that coupons cannot bring a price below zero.

- remove the 50% off coupon
- attempt to add the 51% off coupon

Expectation: It should work now.
Observed: Gives you an error that coupons cannot take the price below zero.

cpacia:
I'm not sure how you've been testing this all this time but the purchase button is broke because /js/templates/modals/purchase/Receipt.html is capitalized while the code that calls it uses lower case `receipt.html`

js/views/modals/purchase/Purchase.js
```diff
@@ -159,16 +208,12 @@ export default class extends BaseModal {
             console.log(data);
           })
           .fail((jqXHR) => {
-            if (jqXHR.statusText === 'abort') return;
-            const errMsg = jqXHR.responseJSON && jqXHR.responseJSON.reason || '';
+            const errMsg = jqXHR.responseJSON ? jqXHR.responseJSON.reason : '';
```
jjeffryes: I'm not sure why this reverted. Maybe a merge issue?

js/languages/en-US.json
jjeffryes: I can't reproduce this.

I do see that after the invalid coupon is applied, but before the new one is added, the error message isn't cleared. I've fixed that.

rmisio: @morebrownies would you mind testing this? It's not happening for Josh, but consistently happens for me.

In it's simplest form you:

```
- add coupon for 50% off
- remove the above coupon
- try adding it again.
```

For me it fails with an error that the coupon amount cannot exceed 100% of the price, which is incorrect, because the coupon is only 50% of the price. It's as if it's still considering the removed 50% coupon in the calculation.

morebrownies: Sure thing. Will test out soon.

morebrownies: @rmisio @jjeffryes I tested the scenario above and it worked correctly for me.

To be clear, I did the following:

1) Applied a 50% off coupon
2) Attempted to apply an additional coupon for 51% off (it triggered an error)
3) I removed the initial 50% off coupon
4) Attempted to apply the 51% off coupon (it worked!)

rmisio: Cool. Thanks for that @morebrownies. Since it looks like just me with the issue, I won't hold up the merge.

Oops... sorry Wolf, I just merged without checking in with you. If there are any problems on your end, I'm sure Josh has another daily purchasing PR coming today or tomorrow that you could comment... lol.

morebrownies: Yeah no worries. I haven't tested thoroughly yet, but so far looking pretty good! I'll test more closely once we feel checkout is near feature complete.

PR #482, opened 2017-05-11
Transactions cases

rmisio:
This PR implements the Cases tab on the transactions page.

This PR also adjusts to the updated order states and should be tested and merged in conjunction with this [server PR](https://github.com/OpenBazaar/openbazaar-go/pull/526).


js/utils/currency.js
```diff
@@ -153,7 +153,7 @@ export function formatCurrency(amount, currency, options) {
     // out with the appropriate Bitcoin symbol
     const formattedAmount = new Intl.NumberFormat(opts.locale, {
       minimumFractionDigits: 0,
-      maximumFractionDigits: 6,
+      maximumFractionDigits: 8,
```
rmisio: Ended up restoring this to ensure we're showing prices in their full precision. Had a talk with Wolf and the problem with rounding up, which likely results in an overpayment is that, while the amount it round up is for the most part insignificant (a fraction of a cent currently)... that is variable based on the current BTC / fiat exchange rate and also dependent on the purchasing power were you spend money.

In other words, a 1/5 cent overpayment may turn into a 1 cent overpayment if the BTC price hits a jet and a 1 cent overpayment may be a big deal to people who make a dollar a day.

PR #491, opened 2017-05-22
Order detail discussion

rmisio:
This PR completes the Discussion tab on the Order Detail overlay.

It should be test and merged in conjunction with this [server PR](https://github.com/OpenBazaar/openbazaar-go/pull/534).


rmisio:
@morebrownies One note on this one... we've decided to punt on the message read indicator for the discussion tab, since there are complications with this being a group chat. Basically, the server would have to be refactored to keep track of which chat participant read the message. Also, the design would need to be refactored as well... for example, let's say you're the buyer and the message has the little green border... does that mean it was read by the moderator, vendor or both.

Anyways, I'm not sure such an indicator works on a group chat.

morebrownies:
>  Basically, the server would have to be refactored to keep track of which chat participant read the message. 

ah ok :(

>  Also, the design would possibly need some adjustment as well... for example, let's say you're the buyer and the message has the little green border... does that mean it was read by the moderator, vendor or both.

FWIW the unread indicators were intended to display which messages were previously unread by you. So in your example above, if you're the buyer, and the vendor already read the most recent message, you (the buyer) would still see the green indicator because they haven't read it yet.

styles/modules/modals/_orderDetail.scss
```diff
@@ -1,5 +1,223 @@
...
+        transform: translate(-50%, 0);
+      }
+    }
+
+    .convoMessagesWindow {
+      padding: $padLg;
+      overflow-x: hidden;
+      overflow-y: auto;
+      max-height: 375px;
+      min-height: 375px;
```
jjeffryes: That's interesting. Why set the min and max height to the same number, instead of using height?

rmisio: Yeah, I'll switch it to just height. The min / max height approach was needed on tables since the table only occasionally honored the non min / max value... but that's probably a tables only quirk.

styles/modules/modals/_orderDetail.scss
```diff
@@ -1,5 +1,223 @@
...
+
+      .chatConvoMessages {
+        padding: 10px 0;
+
+        .convoMessage {
+          .msgContentBox {
+            padding: 13px;
+            margin-top: 3px;
+            display: inline-block;
+            max-width: 186px;
```
rmisio: Yes, good catch. That other value came from the regular chat css.

PR #495, opened 2017-05-23
Various fixes

rmisio:
- Fixing issue where the transactions table was not correctly sorting by date ('newest' would sort by 'oldest' and vice versa). 
- Fixing a bug that was preventing the Edit / Create Listing modal from validating that a coupon with a fixed price cannot exceed the listing price.
- Ensuring the Discover button is not enabled when the app is not "navigable" (e.g. during start-up, no connection to the server, etc...)
- Changing the timestamps in chat from absolute dates to relative (e.g. 3 minutes ago).


jjeffryes:
@rmisio looks like the coupon test is failing, can you fix that?

PR #502, opened 2017-05-31
Purchasing payment

### LINKED ISSUE #422: Add Purchasing
Functional Requirements

When a user clicks the buy button in a listing, the purchase modal will appear, and the user will be able to use it to complete their purchase of the listing.

## Modal
- [x] When the "Buy" button is clicked in a listing, the Purchase modal will appear.
- [x] The Purchase modal is closed by clicking the close button in the top right of the modal, or by clicking the "Return to Listing" link at the top left.
- [x] The avatar of the seller is shown next to the "Return to Listing" button.

## Summary
- [x] A thumbnail of the listing is shown.
- [x] The title of the listing is shown, limited to one line. It is truncated if it is too long with an ellipses.
- [x] The variants selected on the listing screen are shown, stacked vertically.
- [x] A quantity input is present. It is used to set the quantity being purchased of the listing.
    - [x]  When the quantity input value changes, the quantity, listing price and total price in the sidebar should change.
    - [x] When the Payment section is visible, the quantity input is hidden or removed.
- [x] The total price of the listing is shown.

## Shipping
- [x] If the listing has shipping, the shipping section is shown.
- [x] If the user has at least one address in their settings:
    - [x] There is a drop down to select an existing address.
    - [x] There is a link to add a new address.
- [x] If the user has no addresses in their settings, the Add New Address Form is shown.
- **QUESTION:** to reduce scope, can this open the settings/addresses modal instead, and when it is saved the list of addresses in the dropdown updates? If no addresses are shown, there could be a link to add an address which would open the settings/addresses modal. [YES}
- [x] A list of all the shipping options that can deliver to the chosen address is shown.
- [x] Changing the address in the dropdown should update the list of shipping options.
- [x] If an address the listing can't be shipped to is selected, a message that the address can't be shipped to is shown.
- **QUESTION:** would it be better to disable unshippable addresses?
- [x] The shipping options are sorted by least to most expensive.
- [x] The shipping options are radio buttons, the first one should be selected by default.
- [x] Selecting a shipping option should update the shipping cost and total in the sidebar.

## Add New Address Form
- [x] When the add a new address link is clicked, the add new address form is shown. It is the same as the address form in settings/addresses.
- [x] If the user has at least one address saved, the address form has a "View Addresses" link, that will hide the new address form.
- [x] Clicking the Add Address button will save the current address.
- [x] When a new address is saved, it is added to the user's settings, which will need to be saved and published.

## Payment Type
- [x] the user can select Bitcoin or altcoins (via ShapeShift).
- [x] the user can select or deselect the Moderated Payment option.
- [x] if the user selects the Moderated Payment option, the Moderators section appears. It disappears if the option is deselected.
-[ ] if moderators are available, Moderated Payment is checked by default.
- **NOTE:** Remove the radio button
- **QUESTION:** have we ruled out timeout payments for v2.0 at this point? If so, the disclaimer text should be modified appropriately.

## Moderators
- [x] show a list of available moderators for this purchase.
- [x] the list uses radio buttons, the first one is selected by default if the list is visible
- [x] show a static version of the fee (don't re-calculate it on the fly based on the listing price)
- [x] the first moderator to load is selected by default
- [x] the Pay button can't be clicked if Moderated Payment is checked but no moderator is selected (or loaded)

## Additional Information
- [x] there is an input for an email address
- [x] if the listing has coupon codes, show the Coupon Code Input.
- [x] if a coupon code has been applied that matches an existing code, show a message showing it has been added.
- [x] when a coupon is added, the code will be hashed, and compared to the list of hashes in the contract. If it matches one of them, the coupon is applied. If it does not match, an error is shown.
- [x] any number of coupons can be added, they are listed below the input (they are sent to the server as an array of coupon codes). Duplicates are disregarded. 
- [x] there should be a X next to each coupon code 

## Sidebar
- [x] if possible, the sidebar should have a fixed position, and move with the page when it is scrolled.
- [x] when the payment is confirmed, the Pay button becomes a Close button and a confirmation message is shown below it.
- [x] show the quantity
- [x] show the listing total price (price x quantity)
- [x] if the product is physical, shipping should be displayed. If no shipping option has been selected yet (if the user has no addresses, for instance), it should be 0.
- [x] if there is a coupon code applied, it should display here
- [x] the total should be displayed, adding all the price numbers
- [x] **NOTE:** do not disabled the Pay button, instead show an error message (and scroll to it if necessary) 

## Payment
- [x] the current Paid or Unpaid status is shown. **NOTE:** remove this 
- [x] when the Refresh button is clicked, the payment status is refreshed. NOTE: remove the refresh button
- [x] a QR code is shown that can be used to pay for the transaction.
- [x] the amount that needs to be paid in BTC is shown (apply the BTC denomination from their settings)
- [x] when the copy link next to the amount (or the amount) is clicked, the amount is copied to the clipboard and a copied tooltip is shown.
- [x] the payment address is shown
- [x] when the copy link next to the address (or the address) is clicked, the address is copied to the clipboard and a copied tooltip is shown.
- [x] when the Pay from wallet button is clicked, a confirmation box is shown.
    - [x] if the user has sufficient funds (based on the total, ignoring fees), the confirmation box shows the amount to be paid, a cancel button, and a confirm button.
    - [x] show the current total in the wallet also (not in the designs)
    - [x] when the confirm button is pressed, it will have a spinner.
- [ ] when the Pay with altcoins button is clicked, the Shapeshift modal is shown.

## Payment Confirmed
- [x] when the payment is confirmed, the Payment Confirmed section is shown and the Payment section is hidden.
- [x] the purchases link opens the Transactions/Purchases screen when clicked.
- [x] there is an area to send a message to the seller.
- [x] **NOTE:**  drop the emojis unless they are very easy to add

jjeffryes:
This PR completes the purchasing functionality and closes #422. 

- when a purchase has been made the payment (pending) section is shown.
- a QR code is shown that can be used to pay for the purchase.
- the payment address and amount can be copied.
- the pay from wallet button shows a tooltip with the current wallet balance and a warning if the user doesn't have enough funds in their wallet.
- if the user's wallet balance changes while the tooltip is open, it will re-render.
- the pay with alt coins button will create a window for the shapeshift functionality (automatically populated by shapeshift). 
- when a purchase has been paid for, either with the internal wallet or an external source, the purchase complete section will be shown. 
- a link is shown that opens the transaction view and the transaction detail for that purchase. 
- a text area is present that can be used to send a chat message to the vendor.


morebrownies:
Will start reviewing this PR soon!

morebrownies:
screenshot 2017-06-01 11 31 03

I find this to be a strange experience @jjeffryes. All of @cpacia set moderators no longer appear to be offering moderator services, but it's still telling me to select a moderator.

If there a way to detect this scenario and eliminate the "Moderator payment" checkbox when necessary? The average person may not know how to move forward at this point.

morebrownies:
@jjeffryes here's a few issues I found that will require a bit of clean up:

1) The vertical spacing between sections should be gutterV instead of gutterVMd
screenshot 2017-06-01 13 45 21

2) I noticed the coupon input doesn't always display. Any reason why?
screenshot 2017-06-01 13 47 00

3) Radio buttons would definitely be more intuitive for selecting a moderator instead of the check action. Is this change on hold? 
screenshot 2017-06-01 13 50 29

4) Suggestion for a copy change on this error modal "You can't purchase your own listing" might be a better title. 
screenshot 2017-06-01 13 52 01

5) Can you add a tx5 class to js-receipt? It feels a bit bulky as is. 
screenshot 2017-06-01 13 57 31

6) Similar to above, can you make the values right aligned within js-receipt? Should look like this:
screenshot 2017-06-01 14 04 39

7) Clicking the Close icon should take you to the store view and clicking Return to listing should take you back to the listing detail overlay.
screenshot 2017-06-01 14 07 20
screenshot 2017-06-01 14 07 01

8) Buttons should have clrSh2 
screenshot 2017-06-01 14 42 13

9) A few things in this copy block... The link should be green and point to https://www.openbazaar.org/bitcoin/. Also some of the copy is off since we're not doing timed releases. Please update copy to: "Your payment will be held securely in escrow. If any issues arise, you’ll have the ability to open a dispute." Also, that copy should only display if/when the transaction type is multisig. 
screenshot 2017-06-01 14 44 17

I'm going to continue testing a bit more to see if anything else comes up.

morebrownies:
I'm running into this error pretty often now after clearing my database. What does it mean?

screenshot 2017-06-01 18 16 17

morebrownies:
Also, I can't highlight the Pay amount on this screen 

screenshot 2017-06-01 18 36 53

cpacia:
@morebrownies I think if you get the signature error it's because the vendor needs to update his listing. It's old code.

js/views/modals/purchase/Payment.js
```diff
@@ -0,0 +1,175 @@
...
+  constructor(options = {}) {
+    if (!options.model || !options.model instanceof Purchase) {
+      throw new Error('Please provide a purchase model.');
+    }
+
+    super(options);
+    this.options = options;
+
+    this.boundOnDocClick = this.onDocumentClick.bind(this);
+    $(document).on('click', this.boundOnDocClick);
```
rmisio: this.boundOnDocClick needs to be unbound in remove().

js/views/modals/purchase/ConfirmWallet.js
```diff
@@ -0,0 +1,68 @@
...
+  constructor(options = {}) {
+    if (!options.displayCurrency) {
+      throw new Error('Please provide a display currency.');
+    }
+
+    super(options);
+    this.options = options;
+    this.fee = false;
+
+    this.confirmedAmount = app.walletBalance.get('confirmed');
```
rmisio: Why keep track locally of the confirmedAmount instead of getting it from the walletBalance model as needed? You end up having a local variable that you have to keep in sync with that model.

js/views/modals/purchase/Complete.js
```diff
@@ -0,0 +1,126 @@
...
+import ChatMessage from '../../../models/chat/ChatMessage';
+import { getEmojiByName } from '../../../data/emojis';
+
+
+export default class extends BaseVw {
+  constructor(options = {}) {
+    super(options);
+    this.options = options;
+
+    this.processingTime = this.options.listing.get('item').processingTime ||
```
rmisio: Could we check that a listing model is passed in as an option. It makes the dependency clearer, since I don't think it would gracefully handle the scenario of the listing not being passed in.

js/views/modals/purchase/Complete.js
```diff
@@ -0,0 +1,126 @@
...
+      app.polyglot.t('purchase.completeSection.noData');
+    this.vendorPeerID = this.options.listing.get('vendorID').peerID;
+    this.orderID = ''; // supplied by pending.js when the order is paid for
+    this.state = 'pendingPayment';
```
rmisio: State doesn't seem to be used anywhere, other than to pass it into the template where it's not used.

js/views/modals/purchase/Complete.js
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  className() {
+    return 'complete';
+  }
+
+  events() {
```
rmisio: Can you prevent default on keydown?

```
  onKeyDownMessageInput(e) {
    if (!e.shiftKey && e.which === 13) e.preventDefault();
  }
```

That way, the cursor won't jump to the next line before the message is sent.

js/views/modals/purchase/ConfirmWallet.js
```diff
@@ -0,0 +1,68 @@
...
+
+    // re-render if the confirmed amount in the wallet changes
+    this.listenTo(app.walletBalance, 'change:confirmed', (md, confirmedAmount) => {
+      this.confirmedAmount = confirmedAmount;
+      this.render();
+    });
+
+    // fetch the estimated fee and rerender when it returns
+    this.fetchEstimatedFee = $.get(app.getServerUrl('wallet/estimatefee/?feeLevel=NORMAL'))
```
rmisio: Could we abort this on remove().

js/views/modals/purchase/Payment.js
```diff
@@ -0,0 +1,175 @@
...
+    spend({
+      address: this.model.get('paymentAddress'),
+      amount: this.model.get('amount'),
+      currency: 'BTC',
+    })
+      .fail(jqXhr => {
+        openSimpleMessage(app.polyglot.t('purchase.errors.paymentFailed'),
+          jqXhr.responseJSON && jqXhr.responseJSON.reason || '');
+      })
+      .always(() => {
```
rmisio: I think it's ok to not abort this call on remove(), since we want the wallet to reflect the payment. Also, it's probably not bad for the user to see any errors if the Purchase fails on the server.

But... I do think we probably want to throw a `if removed() return` in the always handler. What's in there is pretty benign if executed after the modal closes, but for future proofing it wouldn't hurt.

rmisio:
@jjeffryes Could also merge in master and fix the conflict?

rmisio:
The Purchasing modal bombs when one of the shipping options is a local pickup:

image

There are no services in that case... so maybe that's causing the issue..?

jjeffryes:
> I noticed the coupon input doesn't always display. Any reason why?

@morebrownies if the listing doesn't have coupons, the coupon input isn't shown. That was in the requirements.

> Radio buttons would definitely be more intuitive for selecting a moderator instead of the check action. Is this change on hold?

Yes. This component is also used in settings/store, which doesn't have a radio button. This is a compromise between the designs of the two, since they don't work the same.

jjeffryes:
@morebrownies and @rmisio this PR has been updated. 

@cpacia I'm going to make the updates to the notification format a separate PR.

morebrownies:
Awesome, looking forward to testing this out. I'm planning to spend some time on it tomorrow morning.

morebrownies:
There seems to be a bug here. The free shipping option is checked by default, but when I try to proceed to pay, I get hit with the "Please choose a valid shipping option" error.

screen shot 2017-06-16 at 9 48 06 am

morebrownies:
The issue above happens when trying to purcash this item ob://vintagefashion/store/zebra-dress

jjeffryes:
@morebrownies that issue should be fixed now.

morebrownies:
Alright just had a chance to play around with this some more and it's looking very good. Nice work @jjeffryes 

My 2 pieces of feedback at this point are:

1) We should consider always having the coupon input visible, regardless if the vendor has active coupons or not. As a buyer, I may have a code and not know that is has expired, but if the coupon input is hidden, there's no way for me to test the code. It would be a better experience to let the user enter a invalid coupon and receive a error message than to leave them confused. 

2) I brought this up above and I'm assuming we'll revisit it after launch, but wanted to call it out one more time in case it was missed. The checkbox experience for moderators is strange. Ideally we would use native radio buttons here instead of the custom check button. It would be best to have the radio buttons on the left side of the avatar too, as pictured here:

screen shot 2017-06-19 at 2 13 19 pm

Other than that, looking very good. We'll definitely want to have the community hammer on this and see if they get caught up anywhere too, but overall feels like a smooth checkout flow; much much better than 1.0.

morebrownies:
I spoke too soon, just ran into this issue:

screen shot 2017-06-19 at 2 19 34 pm

I can't finish purchasing because no mods are online, even when the Moderated Payment option is unchecked.

rmisio:
> Ideally we would use native radio buttons here instead of the custom check button.

@jjeffryes We could circle back to this after launch... could that mod view be abstracted to where it takes an option on whether to use a radio button or a checkbox? If it's a radio, there could be a hidden radio button that you could select when you don't want the mod selected. I'm assuming some view already handles the logic to deselect all other mods when one is selected, so I would think that could be abstracted to handle the radio button case.

morebrownies:
Sorry, one other issue I noticed... 

Hitting "BUY NOW" on a listing with a price of 0 does not do anything. We should have a error message trigger when a user clicks "BUY NOW" on a free listing.

rmisio:
> We should have a error message trigger when a user clicks "BUY NOW" on a free listing.

Are we allowing free listings?

morebrownies:
I believe the free listings are coming from the mobile app.

rmisio:
> I believe the free listings are coming from the mobile app.

If we've made the decision that we can't allow free listings because it would have to change the server architecture (since there's not BTC transaction) and we're already not allowing coupons to take a listing at or below zero.... shouldn't we follow through and have the server (and then clients) prevent a 0 amount listing from being saved?

@jjeffryes ^

jjeffryes:
No, free listings shouldn't be allowed by the server, since it's not possible for them to ever have a successful transaction. 

Weirdly, the purchase API allows you to make a purchase of a zero cost item.

I'll put in a warning so you can't buy something that costs zero, but it should be fixed on the server so it's not possible (there could still be rogue servers that allow it).

jjeffryes:
https://github.com/OpenBazaar/openbazaar-go/issues/546

jjeffryes:
@morebrownies that bug with the moderators is fixed, and I put in the radio button. I kept the view button, because the user still needs to be able to look at the details of a moderator (clicking anywhere but the radio button will show the details, but I think there should be a visual prompt for it).

The listing price will update according to the variant selected now, and an error message will be shown if the total price is zero. This would allow a zero price listing with a variant that adds a price, which is kind of weird but would still be a valid transaction.

There is also code in the purchase view that makes sure the price is positive, just in case somehow they get to that screen with a zero price listing, or a negative shipping or variant price is somehow applied.

And the coupon entry is always visible now.

morebrownies:
Hey @jjeffryes I noticed local pickup options have issues. I have one selected but can not continue to check out. 

screen shot 2017-06-20 at 3 19 36 pm

jjeffryes:
@morebrownies can you give me the link to that listing? I can't reproduce the issue.

morebrownies:
ob://QmbG9sv6Veox66CoretT614XGmXqwXPzdUFfitDQLMJWtM/store/milly-beamer

jjeffryes:
@morebrownies it works fine if your address is in albania. :)

There was a bug that was showing shipping options you can't use. That's fixed now.

jjeffryes:
@morebrownies this is just waiting for your final thumbs up before we merge it.

morebrownies:
Ok great. I'll do a final pass after I wrap up Tor designs. Shouldn't be too long.

morebrownies:
Looking good to me. All the issues I posted above seem to be resolved. 👍

Issue #371, opened 2017-02-21
Make color square clickable


srhoulam:
Brief Description: The color preview square in the settings screen can't currently be interacted with. 

Operating System (OS and version): Ubuntu 16.04
OpenBazaar version (shown on `About OpenBazaar` page in menu): 2.0.0

Observed Behavior:
Clicking the color square triggers no behavior.

Expected Behavior:
Present a color picker on click, ultimately removing the color code input.

Additional info (links, images, etc go here):
https://projects.invisionapp.com/share/RC9Q3H5X8#/screens/211105705_Openbazaar-2-0-Page-Store-Customize-Advanced-Picker

jjeffryes:
We could do a two-step implementation here. An initial, and very easy, fix, it to use the native HTML color input. 

Later, if/when we implement a custom color selector for the customize user page story, we can also use that color selector here.

rmisio:
This is done (at least a simple version using the native HTML color input).

PR #512, opened 2017-06-07
Make color square in settings clickable UI #371

jashot7:
This is the first step to implement a color picker for color values in page settings. The native HTML5 color picker is used, therefore there are limitations to how much we can control is visually. The dark grey/black border around the color square is not something we're able to change until we implement our own color picker. Also, the UI that is instantiated for picking a color is completely not controllable by our view. For example, the OS X color picker will stay open until the user closes it even if the app is closed or restarted.

### DIFF
--- a/js/views/baseVw.js
+++ b/js/views/baseVw.js
@@ -38,8 +38,8 @@ export default class baseVw extends View {
   getFormData(selector) {
     const $formFields = selector instanceof $ ?
       selector : this.$(selector ||
-        `select[name], input[name], 
-        textarea[name]:not([class*="trumbowyg"]), 
+        `select[name], input[name],
+        textarea[name]:not([class*="trumbowyg"]),
         div[contenteditable][name]`
       );
     const data = {};
@@ -148,6 +148,51 @@ export default class baseVw extends View {
     }
 
     super.delegateEvents();
+  }
+
+  /**
+   * Returns a given jQuery Element or a locally cached version for optimization.
+   * NOTE: Ensure that child views call super.render() in their render function to clear the cache
+   *       since the DOM elements in the cache will then be stale references.
+   * @param {selector} string - The jQuery selector for the Element.
+   *
+   * @return {element} JQuery The element(s) found in the View's Dom or the View's Cache
+   */
+  getCachedElement(selector) {
+    let element;
+
+    // Ensure we have our cached elements map.
+    if (!this._cachedElementMap) {
+      this._cachedElementMap = new Map();
+    }
+
+    // If the cache has the element, we shall use it.
+    if (this._cachedElementMap.has(selector)) {
+      element = this._cachedElementMap.get(selector);
+    } else {
+      // The cache does not not have the element, therefore query with jQuery and cache it.
+      element = this.$(selector);
+      this._cachedElementMap.set(selector, element);
+    }
+
+    return element;
+  }
+
+  /** Clears the cached elements map. */
+  clearCachedElementMap() {
+     // Clear the cache map.
+    if (this._cachedElementMap) {
+      this._cachedElementMap.clear();
+    }
+  }
+
+  /** It is necessary to call super.render() in child views' render methods if using
+   *  getCachedElement()
+   *  @param {this} Render requires this to be returned.
+   */
+  render() {
+    this.clearCachedElementMap();
+    return this;
   }
 
   // Will call the remove method of any child views.
--- a/js/views/modals/Settings/Page.js
+++ b/js/views/modals/Settings/Page.js
@@ -45,23 +45,23 @@ export default class extends baseVw {
 
   /** Handles when a hex color code is entered by updating color picker. */
   handleColorCodeEntered(event) {
-    const colorPickerId = $(event.target).data('color-picker-id');
-    const $colorPicker = $(`${colorPickerId}`);
-    const newHexColorCode = event.target.value;
-
-    // If the text passes a basic RegExp for a valid 6 digit hex value,
-    // update the color picker's color.
+    var colorPickerId = $(event.target).data('color-picker-id'),
+        $colorPicker = $("#" + colorPickerId),
+        newHexColorCode = event.target.value;
+    
+    // If the text passes a basic RegExp for a valid 6 digit hex value, update the color picker's color.
     if (/[0-9A-F]{6}/i.test(newHexColorCode)) {
       $colorPicker.val(newHexColorCode);
-    }
+    } 
+
   }
 
   /** Handles when a color is chosen from the color picker by updating hex color code text. */
   handleColorChosen(event) {
-    const hexInputId = $(event.target).data('hex-input-id');
-    const $hexInputEl = $(`${hexInputId}`);
-    const newColor = event.target.value;
-
+    var hexInputId = $(event.target).data('hex-input-id'),
+        $hexInputEl = $("#" + hexInputId),
+        newColor = event.target.value;
+    
     $hexInputEl.val(newColor);
   }
 
@@ -272,7 +272,7 @@ export default class extends baseVw {
       installRichEditor(this.$('#settingsAbout'), {
         topLevelClass: 'clrBr',
       });
-
+      
       const avatarPrev = this.$('.js-avatarPreview');
       const avatarInpt = this.$('#avatarInput');
       this.avatarCropper = this.$('#avatarCropper');
--- a/js/views/modals/Settings/Page.js
+++ b/js/views/modals/Settings/Page.js
@@ -30,7 +30,39 @@ export default class extends baseVw {
       'click .js-avatarRight': 'avatarRightClick',
       'click .js-headerLeft': 'headerLeftClick',
       'click .js-headerRight': 'headerRightClick',
+      'change .js-primaryColorPicker': 'handleColorChosen',
+      'change .js-secondaryColorPicker': 'handleColorChosen',
+      'change .js-textColorPicker': 'handleColorChosen',
+      'change .js-highlightColorPicker': 'handleColorChosen',
+      'change .js-highlightTextColorPicker': 'handleColorChosen',
+      'change .js-primaryColorCode': 'handleColorCodeEntered',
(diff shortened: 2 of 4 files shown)


js/views/modals/Settings/Page.js
```diff
@@ -30,9 +30,41 @@ export default class extends baseVw {
...
+      'change .js-textColorCode': 'handleColorCodeEntered',
+      'change .js-highlightColorCode': 'handleColorCodeEntered',
+      'change .js-textHightlightColorCode': 'handleColorCodeEntered',
     };
   }
 
+  /** Handles when a hex color code is entered by updating color picker. */
+  handleColorCodeEntered(event) {
+    const colorPickerId = $(event.target).data('color-picker-id');
+    const $colorPicker = $(`${colorPickerId}`);
```
jjeffryes: Also, be consistent with naming, if you end the hexInputEl with El, end colorPicker with El.

js/views/modals/Settings/Page.js
```diff
@@ -30,9 +30,41 @@ export default class extends baseVw {
...
+      'change .js-primaryColorPicker': 'handleColorChosen',
+      'change .js-secondaryColorPicker': 'handleColorChosen',
+      'change .js-textColorPicker': 'handleColorChosen',
+      'change .js-highlightColorPicker': 'handleColorChosen',
+      'change .js-highlightTextColorPicker': 'handleColorChosen',
+      'change .js-primaryColorCode': 'handleColorCodeEntered',
+      'change .js-secondaryColorCode': 'handleColorCodeEntered',
+      'change .js-textColorCode': 'handleColorCodeEntered',
+      'change .js-highlightColorCode': 'handleColorCodeEntered',
+      'change .js-textHightlightColorCode': 'handleColorCodeEntered',
```
jjeffryes: .js-highlightTextColorCode doesn't have an action associated with it. I think js-textHightlightColorCode was supposed to be that one.

js/views/modals/Settings/Page.js
jashot7: Optimization has been added - check it out. I think something like what I added would be useful in the base view; what do you think? Naming consistency was updated to match the rest of the view FTMP - good point. The '#' has been added to the ids in the data fields and also the optimization removes this goofyness. I ended up with this post-lint when I switched over to templatized strings.

js/views/modals/Settings/Page.js
```diff
@@ -30,9 +30,41 @@ export default class extends baseVw {
...
+    // update the color picker's color.
+    if (/[0-9A-F]{6}/i.test(newHexColorCode)) {
+      $colorPicker.val(newHexColorCode);
+    }
+  }
+
+  /** Handles when a color is chosen from the color picker by updating hex color code text. */
+  handleColorChosen(event) {
+    const hexInputId = $(event.target).data('hex-input-id');
+    const $hexInputEl = $(`${hexInputId}`);
```
jashot7: This doesn't work (pre-fixes) because there was no "#" in that string, so that's why I was using the string templates to add it. The fix just adds it to the data properties because that's way more efficient.

js/views/modals/Settings/Page.js
jashot7: Oh, good catch. I totally didn't need all of those anymore. Fixed and _so much_ *cleaner*.

js/views/modals/Settings/Page.js
```diff
@@ -30,9 +30,54 @@ export default class extends baseVw {
...
+  handleColorChosen(event) {
+    const hexInputId = $(event.target).data('hex-input-id');
+    const $hexInput = this.getCachedElement(hexInputId);
+    const newColor = event.target.value;
+
+    $hexInput.val(newColor);
+  }
+
+  /** Returns a given jquery Element or a locally cached version in this view for optimization. */
+  getCachedElement(elementId) {
```
rmisio: This isn't a bad we to abstract the caching of dom elements within a view. At some point, we may even want to consider integrating it into the base view.

A few thoughts:
- You need to clear the map in render() because at that point the DOM elements held in the cache are no longer in the DOM or associate with this view - the view has a fresh set of DOM elements and you don't want to get the old ones when calling this method. You could see how this happens by interacting with the color fields, clicking Save (which re-renders) and then finding the color field functionality implemented here no longer works because you're referencing cached elements that are no longer in this view.
- rather than semantically tie it to an id by using the name `elementId`, should we use something like `selector` and then you could pass in any selector, since often times we're really just working off of classes?

js/views/modals/Settings/Page.js
```diff
@@ -30,9 +30,54 @@ export default class extends baseVw {
...
+    if (!this._cachedElementMap) {
+      this._cachedElementMap = new Map();
+    }
+
+    // If the cache has the element, we shall use it.
+    if (this._cachedElementMap.has(elementId)) {
+      element =  this._cachedElementMap.get(elementId);
+    } else {
+      // The cache does not not have the element, therefore query with jQuery and cache it.
+      element = $(elementId);
```
rmisio: Could we scope the jQuery query to this views container? The Backbone.View contains a convenience method to do just that `this.$('<selector>')` which is short-hand for `$(view.el).find(<selector>)`.

It's generally the best practice to find elements of the view using that approach. Some of the benefits are:
- You're querying only a sub-set of the DOM instead of the entire DOM.
- If there's another element in the DOM outside of this view matched by the same selector, you won't unintentionally target it.

rmisio: If we implement the above change, it means this cached method won't work for dom elements outside of this view... but that's probably a tiny fraction of queries we do and such queries could probably be manually cached (if necessary).

rmisio:
@jashot7 Could you please fix the linting error as well? `npm run lint`

js/views/modals/Settings/Page.js
jashot7: Great feedback. 
- The map is now cleared in render, but the entire <getCachedElement(selector)> method now resides in baseVw. The clearing of this cache is done in subviews with a super.render() at the beginning of their render methods. 
- Selector is now used rather than Id. I like that idea.

js/views/modals/Settings/Page.js
jashot7: - jQuery queries are now scoped to the view container. Thanks for the background on this as well; that was super useful.
 > ... it means cached method won't work for dom elements outside of this view...
- I'm glad you mentioned this so that we keep it in mind going forward.

js/views/modals/Settings/Page.js
```diff
@@ -30,9 +30,33 @@ export default class extends baseVw {
...
+    // If the text passes a basic RegExp for a valid 6 digit hex value,
+    // update the color picker's color.
+    if (/^#([0-9a-f]{6})$/i.test(newHexColorCode)) {
+      $colorPicker.val(newHexColorCode);
+    }
+  }
+
+  /** Handles when a color is chosen from the color picker by updating hex color code text. */
+  handleColorChosen(event) {
+    const hexInputId = this.$(event.target).data('hex-input-id');
```
rmisio: You don't need to change this at all, but as an FYI, in the case when you already have the HTML element, it's not as critical to use `this.$` since it's not going to the DOM anyway... but as long as still works, it's OK.

js/views/baseVw.js
```diff
@@ -150,6 +150,51 @@ export default class baseVw extends View {
     super.delegateEvents();
   }
 
+  /**
```
rmisio: As an FYI, we do have dom element caching in place in most views, but it's a more manual declarative method. For example, if I want to cache the element at `.js-orderDetailsCopiedToClipboard`, I'd implement the following method:

```
  get $copiedToClipboard() {
    return this._$copiedToClipboard ||
      (this._$copiedToClipboard = this.$('.js-orderDetailsCopiedToClipboard'));
  }
```

then in render, we'd add:

`this._$copiedToClipboard = null;`

The approach you implemented involves much less configuration and, at the very least gives us another option going forward. If we find it's working well, we could eventually refactor the other views to use it, although there'd be a little work there + we'd also need to update those views to call render in the baseVw.

js/views/baseVw.js
```diff
@@ -150,6 +150,51 @@ export default class baseVw extends View {
...
+    // Ensure we have our cached elements map.
+    if (!this._cachedElementMap) {
+      this._cachedElementMap = new Map();
+    }
+
+    // If the cache has the element, we shall use it.
+    if (this._cachedElementMap.has(selector)) {
+      element = this._cachedElementMap.get(selector);
+    } else {
+      // The cache does not not have the element, therefore query with jQuery and cache it.
```
rmisio: One thing to keep an eye out for is that if the selector doesn't find an element, it will set that cache to an empty element. In theory, that shouldn't be a problem (why keep going to the DOM for an element that's not there), but I remember a situation where that wasn't ideal and I ended up switching our manual cache to only cache if it finds the element, otherwise the next time it would also go to the DOM for it.

I can't remember the specific use case for that. Maybe it was due to the fact that sometimes in render we need to do something asynchronously via setTimeout since some plugins need an element to be in the DOM before initialization... and maybe an element generated by the plugin wouldn't be there immediately, but would later. Not sure.

We could probably leave it as is for now, but just something to keep in the back of our minds.

rmisio: Actually looking through the code, in the vast majority of cases we also would cache the an empty element. It looks like it was probably some one-off weird case where caching an empty element was a problem.

js/views/modals/Settings/Page.js
jashot7: Ah, good point.

js/views/baseVw.js
jashot7: Useful illustration of current caching versus the new approach. I did see it in there and use it as I began, but the redundancy of having 10 additional methods like the one above led me to the new approach. If we ever do decide to refactor, I could handle the grunt work if you like!

js/views/baseVw.js
jashot7: Even with your second response to this, I'm glad you brought it up. Now it will be more likely to come to mind.

PR #513, opened 2017-06-08
Order detail summary

rmisio:
This PR partially implements the Summary tab of the Order Details overlay. It includes:

- the progress bar up top
- the Order Details section
- the Payment Details section
  - Includes a stubbed Pay For Order section. When the Purchasing PR is all done and merged, I'll integrate with the full Pay For Order section.
- the Order Accepted section
  - As of now the Fulfill and Refund actions are a no-op.

*** One other note... I don't think this functionality will work properly with Purchases for items set for Local Pickup. I'm not 100% clear what the flow is in that case, so I'll integrate that in a subsequent PR once I understand the details.


js/languages/en-US.json
```diff
@@ -11,17 +11,20 @@
   "testnet": "Test Mode",
   "testnetTooltip": "Test Mode uses testnet coins, which are intended for testing purposes only.",
   "tabMenuHeading": "MENU",
+  "fiatBtcPairing": "%{fiatAmount} (%{btcAmount})",
```
rmisio: I moved these three out to re-use globally. I can't imagine these being context specific and it seemed unnecessarily repetitive to keep re-declaring them.

jjeffryes:
On a digital order, I get Ship to counties.NA in the Order Details section. Is that a known issue?

image

jjeffryes:
@rmisio the details screen is still throwing a 
```

```
Error in the console on digital orders, though it no longer blocks the UI from working.

jjeffryes:
Ignore that close, hit the wrong button when leaving a comment.

rmisio:
> should the Cancel link be a different size than the other status texts? It's set to 12px vs. 14px for the status text, because it has a tx6 class on it.

I was following the style of the link below and wanted to be consistent with that. If it shouldn't be like that, I'm sure the big bad wolf can adjust it in his polish work.

image

> when an offline order hasn't been accepted yet, should the Status just have the cancel link in it? Or should it be something like "Pending" with the cancel link after the word Pending?

One thing to consider here is space issues. If we put "Pending" next to the link and the translations of either can too large, then the cancel link might break to a new line and this row will have an inconsistent height relative to the other rows. Not sure how big of a deal that is.

jjeffryes:
> One thing to consider here is space issues. If we put "Pending" next to the link and the translation of either is too large, then the cancel link might break to a new line and this row will have an inconsistent height relative to the other rows. Not sure how big of a deal that is.

Translations should be assumed to occupy more space, if the design doesn't work when the text occupies more lines, we either need to change the design, or accept it looks terrible in some languages.

js/views/modals/orderDetail/summaryTab/OrderDetails.js
```diff
@@ -0,0 +1,115 @@
...
+          .done((modProfile) => {
+            this.modProfile = modProfile;
+            this.moderatorVw.setState({
+              name: modProfile.get('name'),
+              handle: modProfile.get('handle'),
+            });
+          });
+      }
+
+      this.$('.js-moderatorContainer').html(this.moderatorVw.render().el);
```
rmisio: I actually have the N/A being output from the moderator view if there's no moderator. I could refactor to save a DOM change, but just doesn't seem worth the hassle here to save one DOM change per open of the modal.

PR #515, opened 2017-06-12
Wire in Customize and Create Listing buttons for User Page.

### LINKED ISSUE #505: When viewing my own store, I should see Customize and Create Listing buttons
screenshot 2017-06-02 10 40 31

Currently I see **Customize** and **Example Button**. It would be nice if when I click Customize, it opens Settings > Page and when I click Create Listing it opens the Create Listing modal.

### LINKED ISSUE #511: If Vendor = False, Don't Show Store Tab To Others
If a user has their vendor flag set to false, don't show their store tab on their user page to users that are not themselves (ie: they can see their own store tab, but no one else can). 

If another user uses a url to go to their store tab, show the page not found error.

jjeffryes:
- stub in block button
- hide the store tab if not looking at your own page and the user is not a vendor
- if the store tab is navigated to directly, show the home tab if the store tab shouldn't be shown
- show a warning if the user has listings but no one else will be able to see them

Note that ideally the store route would be blocked in the router, this PR defers that functionality for now because it would require more extensive refactoring, due to various actions happening that assume the store is a valid option before the value of profile.vendor is known. 

This closes #505 
This closes #511


morebrownies:
Looks good to me 👍  (i tested the buttons on the page and toggling the store on/off)

js/views/userPage/Store.js
```diff
@@ -63,6 +64,7 @@ export default class extends BaseVw {
 
     this.listenTo(app.settings, 'change:country', () => (this.showShippingChangedMessage()));
     this.listenTo(app.settings, 'change:localCurrency', () => (this.showDataChangedMessage()));
+    this.listenTo(app.profile, 'change:vendor', () => (this.showDataChangedMessage()));
```
rmisio: Can we just toggle a 'hide' class on that 'your store is currently'... box? Seems like a lot smoother update than showing this data changed pop-up and even if, on remote servers, the change happens with a lag and the user is doing something on the page... doesn't seem like it would be so distracting.

js/templates/userPage/store.html
```diff
@@ -68,6 +68,12 @@
           </div>
         </div>
       </div>
+      <% if (!ob.vendor) { %>
+        <div class="contentBox row pad clrP clrBr">
```
rmisio: Piggy backing on the comment below... maybe changing this to:

`<div class="contentBox row pad clrP clrBr <% if (ob.vendor) print('hide')%>">` and then toggling that hide class on `change:vendor`?

js/views/userPage/UserPage.js
```diff
@@ -18,12 +20,16 @@ export default class extends baseVw {
     super(options);
     this.options = options;
 
+    this.ownPage = this.model.id === app.profile.id;
+
     this.state = options.state || 'store';
+    if (this.state === 'store' && !this.ownPage && !this.model.get('vendor')) {
```
rmisio: Can we please handle this in the router by showing the Page Not Found page? It can be done by adding this to line 196:

```
      if (state === 'store' && !profile.get('vendor') &&
        profile.id !== app.profile.id) {
        this.pageNotFound();
        return;
      }
```

That's more consistent with how we're handling when you put in a route to a non-existent / active page. I've confirmed with @morebrownies that is the way we want to go.

rmisio: I saw the note in the description about some complications to this approach. If I'm missing something... please let me know.

PR #517, opened 2017-06-13
Order detail summary continued

rmisio:
The PR adds the following functionality to the Order Details overlay:
- fulfill an order
- issue a refund by the vendor

Notes:
- ensure you are on the latest server
- @jjeffryes Could you help me test something? When I do a refund (at least for a direct, online order), the refund API succeeds and a subsequent fetch of the order shows a REFUNDED state, but the refund transaction never shows up. It should look like this:

```
    "refundAddressTransaction": {
        "txid": "9b55dd7948519cda102db349165e3b1a43585adf47065fad5366a2ecb0a80461",
        "value": 149508,
        "confirmations": 1236,
        "height": 1127249,
        "timestamp": "2017-06-07T23:02:51.892599179Z"
    }
```

It never shows up for me, but according to CP, the it shows up in the unit tests. If it also does not show up for you, could you please log a server bug for it?

### DIFF
--- a/js/languages/en-US.json
+++ b/js/languages/en-US.json
@@ -1038,7 +1038,15 @@
       }
     },
     "fulfillOrderTab": {
-      "heading": "Fulfill Order"
+      "heading": "Fulfill Order",
+      "shippingCarrierLabel": "Shipping Carrier",
+      "shippingCarrierPlaceholder": "UPS, FedEx, DHL, etc",
+      "trackingLabel": "Tracking #",
+      "trackingPlaceholder": "optional",
+      "fileUrlLabel": "File URL",
+      "fileUrlPlaceholder": "Link to file(s)",
+      "passwordLabel": "Password",
+      "passwordPlaceholder": "Password to file (optional)"
     }
   },
   "orderUtil": {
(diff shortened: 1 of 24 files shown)


jjeffryes:
After fulfilling an order, the fulfill order button is still there, is that not finished yet?

image

jjeffryes:
@rmisio on the refund, it shows up for the refundee, but I don't see it in the refunder's transactions. Is that the same problem you're seeing?

(thanks to fees, the $4.67 refund was $2.31 when it arrived...)

rmisio:
> on the refund, it shows up for the refundee

@jjeffryes How are you determining that? Is the refund showing up in Order Details overlay Summary tab or in the "refundAddressTransaction" key in the API payload?

If you mean it shows up in your wallet, that's not what I was asking.

rmisio:
> After fulfilling an order, the fulfill order button is still there, is that not finished yet?

Should be finished. I'll look into it.

rmisio:
> (thanks to fees, the $4.67 refund was $2.31 when it arrived...)

Yeah... to not waste too much testnet coins, I was making purchases with low prices. For a refund of those, the refund amount is 0... lol / scream.

jjeffryes:
> @jjeffryes How are you determining that? Is the refund showing up in Order Details overlay Summary tab or in the "refundAddressTransaction" key in the API payload?

If I click the refund link, the summary tab refreshes and shows the refund in the progress bar, and the new fetch of the order has the refund in the data returned.

```
 "refund": {
            "orderID": "Qmdbsp5mhSoT4mcMsVExX6rEe3KYmgu2fjyJ7Mtu3TSmYx",
            "timestamp": "2017-06-14T21:50:21.290690900Z",
            "memo": ""
        },
```

I see two transactions in paymentAddressTransactions, but I don't see a refundAddressTransaction anywhere at all. 

image

PR #519, opened 2017-06-15
Order completion

rmisio:
This PR adds in the ability to complete an order and leave a review.

### DIFF
--- a/js/views/modals/orderDetail/summaryTab/CompleteOrderForm.js
+++ b/js/views/modals/orderDetail/summaryTab/CompleteOrderForm.js
@@ -17,7 +17,12 @@ export default class extends BaseVw {
       throw new Error('Please provide an OrderCompletion model.');
     }
 
+    if (!options.slug) {
+      throw new Error('Please provide the listing slug.');
+    }
+
     this.ratingStrips = {};
+    this.slug = options.slug;
 
     const ratings = this.model.get('ratings');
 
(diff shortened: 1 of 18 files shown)


styles/components/_misc.scss
```diff
@@ -39,3 +39,26 @@
   background-color: #000;
   border-radius: 4px;
 }
+
+.ratingStrip {
+  .ratingIcons {
+    display: inline-flex;
+    flex-direction: row-reverse;
```
rmisio: Wish we had a previous sibling selector in CSS. Instead, I'm fudging it with reversing the order in flex.

js/views/modals/orderDetail/summaryTab/CompleteOrderForm.js
```diff
@@ -0,0 +1,114 @@
...
+      overall: this.ratingStrips.overall.rating || undefined,
+      quality: this.ratingStrips.quality.rating || undefined,
+      description: this.ratingStrips.description.rating || undefined,
+      deliverySpeed: this.ratingStrips.deliverySpeed.rating || undefined,
+      customerService: this.ratingStrips.customerService.rating || undefined,
+      slug: this.slug,
+    };
+
+    this.rating.set(data);
+    this.rating.set(data, { validate: true });
```
rmisio: The reason is that I want to both validate the model and set with potentially invalid values. If you set with the validate option, the model will not set those attributes if any are invalid. But.. since I'm re-rendering the errors, I want them in the model.

Normally, this can be avoided by using Model.save(), which does set the attributes and will also validate, but in this case I don't want that because Save will kick off a request to the server if there are no errors and the request to the server will happen in the utility function.

PR #523, opened 2017-06-21
Reviews

### LINKED ISSUE #465: Add Reviews to Listings
## User Story
As a user, I want to see the reviews a listing has when I am looking at the listing. I should see a useful number of the latest reviews, and have a way to take an action to see the rest of them. I should be able to confirm a review is valid by taking an action to see the details of the transaction associated with the review.

### Review Section
The reviews should appear as a separate section within the listings screen, below the photo carousel.
- [x] If the listing has no reviews, a "This listings doesn't have any reviews yet." message is shown.
- [x] An error message could be shown in the same place if fetching the reviews fails. 
- [x] The three latest reviews should be shown.
- [x] There should be a "Load More" button at the bottom of the reviews section.
- [x] Clicking the Load More button will add 10)more reviews to the reviews section. It should automatically scroll the 1st new review into view.
- [x] The reviews section should expand to the height of its content. 
- [x] When the reviews are loading, a spinner should be shown on the Load More button.

### Review
Each review shows the review left by a buyer, and their ratings.
- [x] The date and time the review was left should be shown.
- [x] The Name of the reviewer (at the time the review was left) should be shown, unless they made the review anonymous.
- [x] When the user name is clicked on, the app navigates to that user's page.
- [x] The text of the review should be shown. If it is too long, it should be truncated with an ellipses.
- [x] There should be a show more link of some kind that will reveal the rest of the text.
- [x] The ratings for the review are shown. 

## Listing Card
- [x] update the listing card to show the review total and the rating average.

## Listing
- [x] when the review api returns, update the review total and the rating average.

## Data 
Reviews API will have pagination, with a starting review and a number of reviews to fetch. 
The Reivews API should return the total reviews and the average rating.

jjeffryes:
This adds the review display section to the listings details modal.

This PR also adds the average rating and number of ratings icons to the listing cards. 

Closes #465


morebrownies:
Nice, will review soon.

morebrownies:
Client doesn't load for me on this branch, running into this error in the JS console:

screen shot 2017-06-21 at 3 30 34 pm

jjeffryes:
@morebrownies pull latest, I forgot to commit a missing comma I fixed earlier today.

morebrownies:
Ok I found a few issues for you to revisit when you're back from vacation @jjeffryes 

1. The user card on the about screen does not overall store rating / number of ratings.
screen shot 2017-06-22 at 11 46 32 am

2. The top bar should also display the overall store rating / number of ratings (next to "No location").
screen shot 2017-06-22 at 11 55 43 am

3. I changed the format of how to output number of ratings, it should be a super simple fix. Let's remove the Chat bubble icon and display number of reviews in parentheses, as pictured below. Ratings should display like this in every location:
screen shot 2017-06-22 at 11 59 42 am

4. Rating not displaying the same between store view and search view. Store view displays 4.0, search view displays 4.2. Search view also not displaying the number of reviews.
screen shot 2017-06-22 at 12 03 52 pm
screen shot 2017-06-22 at 12 04 06 pm

5. If there are 0 reviews on a listing, **⭐ 0 (0)** should display under the BUY NOW button.  Currently nothing displays and the review section further down the listing detail screen also is hidden.
screen shot 2017-06-22 at 12 05 49 pm

6. The followers tab also doesn't display the number of ratings
screen shot 2017-06-22 at 12 08 54 pm

js/views/modals/listingDetail/Listing.js
```diff
@@ -508,6 +549,11 @@ export default class extends BaseModal {
       this.setSelectedPhoto(this.activePhotoIndex);
       this.setActivePhotoThumbnail(this.activePhotoIndex);
       this.adjustPriceBySku();
+      this.$reviews = this.$('.js-reviews');
+
+      // get the ratings data, if any
```
rmisio: Also, there's no error handling here. For example, if a user has no reviews, the endpoint will error with:

```
{
    "success": false,
    "reason": "no link named 'index.json' under QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn"
}
```

In the case of an error, the first argument to always is the xhr which you end up passing to onRatings() which subsequently bombs with:

image

I would think instead of always(), you would want to use done() and fail() so you could identify errors (not sure if we want to display them to the user or just not show reviews in that case - the latter may be fine for now, but ideally we handle it in a way that doesn't leave exceptions in the code.)

rmisio: Not sure if you want to fight this battle... I've logged many server issue today, so I'm over it, but ideally the server would return a 200 and an empty array if there are no reviews instead of:

```
{
    "success": false,
    "reason": "no link named 'index.json' under QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn"
}
```

As it is now, there's really no way to distinguish between a user having no reviews and there being an error fetching them. We may be able to punt on this if we decide to both silently ignore errors and no reviews.

(but as soon as Wolf asks for a 'This listing has no reviews' placeholder punting is not an option)

rmisio:
The first review is not showing the timestamp. It doesn't have a name (I don't think I checked the box), but there is a timestamp in the data:

image

rmisio:
Should the expanded view really have an ellipse?

image

In this case it was so long, it looks like the server truncated it, but since the client has no way of knowing this... how would it be able to put the ellipse in that case?

styles/components/_misc.scss
```diff
@@ -80,4 +85,18 @@
       }
     }
   }
-}
\ No newline at end of file
+}
+
+.orderCompleteEvent {
```
rmisio: I guess this could be removed if it's being driven by the 'clickable' state.

js/templates/ratingsStrip.html
```diff
@@ -4,10 +4,13 @@
       // Important!!! In order to simulate a previous sibling selector in
       // CSS and hover all the previos icons on hover, the icons are
       // displayed in reverse order via flex-direction. This requires index
-      // calculations to be computed from the end.    
+      // calculations to be computed from the end.
       const selected = ob.curRating >= ob.maxRating - i;
   %>
-  <a class="ion-star ratingIcon js-ratingIcon <%= ob.iconClrClass %> <% if (selected) print('selected') %>"></a>
+  <!--<a class="ion-star ratingIcon js-ratingIcon <%= ob.iconClrClass %> <% if (selected) print('selected') %>"></a>-->
```
rmisio: Could we remove the commented out line?

styles/modules/modals/_listingDetail.scss
```diff
@@ -162,6 +162,59 @@
...
+              text-decoration: underline;
+            }
+          }
+
+          .ratings {
+            td {
+              width: 50%;
+              padding-bottom: $padSm;
+
+              .emoji {
```
rmisio: I guess this could go away since you're globally applying these styles in _misc.scss?

jjeffryes:
@morebrownies and @rmisio this has been updated, including dead horses.

morebrownies:
I keep running into this error when loading a listing detail modal. Maybe it is because I'm attempting to view a listings on a dated server? 

screen shot 2017-07-03 at 10 49 08 am

jjeffryes:
@morebrownies there have been several updates, it's possible this doesn't work with old nodes any more. I'll look into it.

jjeffryes:
@morebrownies you can retest now, make sure you're using servers updated in the last 30 minutes or so.

js/views/modals/listingDetail/Listing.js
```diff
@@ -90,6 +92,29 @@ export default class extends BaseModal {
...
+    this.rating = this.createChild(Rating);
+
+    // get the ratings data, if any
+    this.ratingsFetch =
+      $.get(app.getServerUrl(`ob/ratings/${this.vendor.peerID}/${this.model.get('slug')}`))
+        .done(data => this.onRatings(data))
+        .fail((jqXhr) => {
+          if (jqXhr.statusText === 'abort') return;
+          // if there are no ratings, no response is returned.
+          if (jqXhr.status === 200) {
```
rmisio: Is this if block really needed? I'm don't think there would ever be a case where a 200 triggers the fail handler.

jjeffryes:
@morebrownies this is just waiting for a final thumbs up from you to merge.

morebrownies:
Awesome. I'll test it again shortly.

morebrownies:
Ok still seeing lots of issues on this PR, but it may be my local server/client that is causing some of the issues.

1. Search results do not show the # of ratings, only the overall rating 
screen shot 2017-07-06 at 3 28 44 pm

2. The listing detail screen does not show the # of ratings or the overall rating
screen shot 2017-07-06 at 3 38 27 pm

3. Total # of reviews shows 4 in the title bar, but it appears only 2 reviews were left on the listings within the store. This may be because a listing with reviews was deleted or maybe there is a bug? not sure, but worth calling out.
screen shot 2017-07-06 at 3 38 52 pm

4. Can you make it so that the default message in the review section (No reviews were found for this listing) in the listing details modal displays in italics like the default messages in **Refund Policy** and **Terms of Service** display?
screen shot 2017-07-06 at 3 38 34 pm

5. Can the reviews be right aligned in the user card? Currently there is a big gap on the right side. 
screen shot 2017-07-06 at 3 44 46 pm

That's all I've found so far

morebrownies:
The format is also off in list view mode on a store

screen shot 2017-07-08 at 10 38 53 am

jjeffryes:
@morebrownies this has been updated.

Notes:

Search Results - the seach data doesn't have the number of reviews yet. When Tyler adds that, they will show up.

Zero Ratings - this is fixed, it will show zeroes instead of blanks when there is no value.

Mismatch in ratings on the top bar: this is a server bug, I told @cpacia about it a while back, but it doesn't appear to be fixed. I have filed an issue at https://github.com/OpenBazaar/openbazaar-go/issues/566

Italicize and color 2 on no reviews text is fixed.

Right align rating is fixed.

Formatting in the grid view is fixed.

The spacing after the star was reduced, it had an extra space.

morebrownies:
Ok, great. Thanks for the update @jjeffryes 

Is this a bug I'm seeing here? I noticed on @cpacia store in the grid view it shows the listing has 4 reviews, but when I click into the listing detail modal I do not see any.

ob://QmZXGKrRRe5RTDhvzhZfSTe8vYBDGV7HBoNz5FS3U9FnUa/store/piss

screen shot 2017-07-10 at 11 05 34 am

screen shot 2017-07-10 at 11 05 40 am

jjeffryes:
@morebrownies that isn't a bug, Chris manually deleted his reviews while testing something, so that store has bad data.

morebrownies:
Ok curious why it still shows reviews at the grid view level? Shouldn't it show the same thing between grid view and listing modal?

jjeffryes:
@morebrownies he must have put them back, I get correct reviews on the listing now. 

image

morebrownies:
I see them now too, but the grid view and modal still aren't showing the same number of reviews and rating. Is this due to caching or something?

jjeffryes:
@morebrownies  I don't know, it's a server issue one way or another.

You can use the test store I set up, it's up to date and has clean data.

ob://QmYfqEPxcFqA1eTDx8vVeNkadLwWbSJpNzvhghDvsSSQG5/store/

morebrownies:
Ok cool. Other than some server weirdness, I think this is looking pretty good.

PR #529, opened 2017-06-27
Order details dispute

rmisio:
This PR adds the following functionality to the Order Details overlay:
- Ability to open a dispute
- Ability for a mod to decide on a dispute
- The UI for a user to accept a payout
- The Pay For Order section as in the Purchase flow

There are some server bugs blocking the completion of the dispute resolution, notably:
- A socket is missing that would go to the other party when a payout is accepted. So, for now, if you are viewing an order that a mod has made a decision on and the other party accepts the payout, you'll need to refresh.
https://github.com/OpenBazaar/openbazaar-go/issues/556
https://github.com/OpenBazaar/openbazaar-go/issues/548
- There are at least a couple of places where the server is not properly handling low priced items. For now, please test this purchasing an item of at least $10.
https://github.com/OpenBazaar/openbazaar-go/issues/558

closes #529


js/models/order/Case.js
```diff
@@ -40,6 +40,15 @@ export default class extends BaseModel {
       // convert price fields
       response.vendorContract.buyerOrder.payment.amount =
         integerToDecimal(response.vendorContract.buyerOrder.payment.amount, true);
+
+      // if (response.resolution) {
```
rmisio: This will be needed once the server is properly providing the payout amount. I'll uncomment it then.

js/models/order/Order.js
```diff
@@ -47,6 +47,15 @@ export default class extends BaseModel {
       // convert price fields
       response.contract.buyerOrder.payment.amount =
         integerToDecimal(response.contract.buyerOrder.payment.amount, true);
+
+      // if (response.contract.disputeResolution) {
```
rmisio: This will be needed once the server is properly providing the payout amount. I'll uncomment it then.

js/views/modals/listingDetail/Listing.js
```diff
@@ -312,7 +312,7 @@ export default class extends BaseModal {
       this.totalPrice = _totalPrice;
       const adjPrice = convertAndFormatCurrency(this.totalPrice,
         this.model.get('metadata').get('pricingCurrency'), app.settings.get('localCurrency'));
-      this.getCachedElement('.js-price').text(adjPrice);
```
rmisio: Shortened this guy's name.

js/views/modals/purchase/Payment.js
```diff
@@ -1,29 +1,44 @@
+/*
```
rmisio: Made a few tweaks to this view to make it more friendly to be re-used in multiple areas, notably:
- Made the inputs more explicit. It had been relying on two models to be passed in, but the Order Detail overlay does not have the same models (at least with the same structure), so I switched it were you just directly pass in the data it needs as options.
- Updated the view so it handles partial payments. In case someone (probably accidentally) makes a partial payment from an external wallet, the totals will be updated to reflect that and if they want to pay the rest from the internal wallet (Pay From Wallet action), it will only send the remaining balance.

js/views/modals/orderDetail/summaryTab/DisputeStarted.js
```diff
@@ -0,0 +1,72 @@
...
+
+  onClickResolveDispute() {
+    this.trigger('clickResolveDispute');
+  }
+
+  getState() {
+    return this._state;
+  }
+
+  setState(state, replace = false, renderOnChange = true) {
```
rmisio: Yes, it probably should. I didn't initially because this is a very specific way of working that only I was using and it was evolving, but even if only I use it, it won't hurt for it to be in the baseVw and devs who aren't interested in it, could leave it be.

If you don't mind though, I think I'll leave this as a separate task to circle back to, since there's now 35 occurrences of it in the app and it's probably not super hight priority to work through this now. Going forward, I'll put setState() and getState() in the baseVw so we don't accumulate more technical debt and at some point I'll re-factor the other views.

js/views/modals/orderDetail/ResolveDispute.js
```diff
@@ -0,0 +1,173 @@
...
+
+    if (!options.buyer) {
+      throw new Error('Please provide a buyer object.');
+    }
+
+    if (!isValidParticipantObject(options.buyer)) {
+      throw new Error(getInvalidParticpantError('buyer'));
+    }
+
+    options.buyer.getProfile().done(profile => {
```
rmisio: It's not necessary because all the profile are lazy loaded in and the view supports them not being there. For example, let's say we display the text "Billy made a purchase" with Billy being the buyer's name. Until Billy's profile is loaded, it will say "The buyer made a purchase" and when the profile arrives the name will pop in. When it's cached, it's so fast you barely notice the change.

In every case it should work similarly (sometimes, it's a GUID in place of handle, etc...).

rmisio:
@jjeffryes I just addressed the code review questions / requests, but please hold off before merging. I'll look into that local pickup bug. If it's a quicky, I'll sneak it in here.

js/views/modals/orderDetail/summaryTab/DisputeStarted.js
jjeffryes: That's fine.

js/views/modals/orderDetail/ResolveDispute.js
```diff
@@ -0,0 +1,173 @@
...
+          .attr('style', getAvatarBgImage(profile.get('avatarHashes').toJSON()));
+      }
+    });
+
+    this.listenTo(orderEvents, 'resolvingDispute', this.onResolvingDispute);
+    this.listenTo(orderEvents, 'resolveDisputeComplete resolveDisputeFail',
+      this.onResolveDisputeAlways);
+
+    this.boundOnDocClick = this.onDocumentClick.bind(this);
+    $(document).on('click', this.boundOnDocClick);
```
rmisio: done.

js/views/modals/purchase/Purchase.js
```diff
@@ -252,10 +252,18 @@ export default class extends BaseModal {
...
+            this.payment = this.createChild(Payment, {
+              balanceRemaining: this.purchase.get('amount'),
+              paymentAddress: this.purchase.get('paymentAddress'),
+              orderId: this.purchase.get('orderId'),
+              isModerated: !!this.order.get('moderator'),
+            });
+            this.listenTo(this.payment, 'walletPaymentComplete',
+              (pmtCompleteData => this.completePurchase(pmtCompleteData)));
+            this.$('.js-pending').append(this.payment.render().el);
+            this.updatePageState('pending');
```
jjeffryes: The page state needs to be updated before actionBtn is rendered, or it renders the wrong state. 

I think this is what caused #533.

js/views/modals/purchase/Purchase.js
```diff
@@ -252,10 +252,18 @@ export default class extends BaseModal {
           contentType: 'application/json',
         })
           .done((data) => {
-            this.updatePageState('pending');
             this.actionBtn.render();
             this.purchase.set(this.purchase.parse(data));
-            this.payment.render();
+            this.payment = this.createChild(Payment, {
```
jjeffryes: Shouldn't there be a if (this.payment) this.payment.remove(); before this?

rmisio: Since you're only going to post ob/purchase once, I don't think it's necessary. Do you see any possible scenario where that guy would be posted more than once?

jjeffryes: It shouldn't, but in #533 it looked like it was being appended twice. Though that may just be because the pay button could be clicked twice.

jjeffryes:
@rmisio there's a weird bug in fulfillment for physical orders. Somehow when the physicalDelivery data is passed from FulfillOrder.js to utils/order.js/fulfillOrder, the shipper and tracking number are both reset to blank.

It looks like it's a model instead of JSON when order.js gets it, I'm not sure how that's happening. 

But it throws and error and you can't fulfill any orders.

```
Uncaught Error: physicalDelivery.shipper: Please provide a shipping carrier.
    at C:/Users/J/ob/openbazaar-desktop/js/utils/order.js:174:17
```

rmisio:
@jjeffryes

> there's a weird bug in fulfillment for physical orders. Somehow when the physicalDelivery data is passed from FulfillOrder.js to utils/order.js/fulfillOrder, the shipper and tracking number are both reset to blank.

ok, should be fixed now.

jjeffryes:
When a dispute is decided but not accepted, should the payout section show zero prices for everything?

image

jjeffryes:
Re: zero prices, that's probably because it's a low price transaction, and the payout was eaten by fees, wasn't it?

@morebrownies I think that could be confusing, should we show a message somewhere explaining that?

jjeffryes:
For some reason the dispute with a settlement of all zeroes is stuck in the decided state for the buyer and seller, but not the moderator. That appears to be a server issue.

morebrownies:
> Re: zero prices, that's probably because it's a low price transaction, and the payout was eaten by fees, wasn't it?

> @morebrownies I think that could be confusing, should we show a message somewhere explaining that?

Well hmmm. That's an interesting scenario. I'm wondering if we should even allow a party to open a dispute when the payout will result in 0? At that point, you're disputing for the point of disputing, but there's no financial outcome for any party involved in the dispute, including the moderator. 

That may open a bigger can of worms, but the easiest thing would probably be to display a message when opening a dispute and viewing the payout details, "The bitcoin transaction fee will be $2.47 (estimated), which has been deducted from the payout totals."

jjeffryes:
> That may open a bigger can of worms, but the easiest thing would probably be to display a message when opening a dispute and viewing the payout details, "The bitcoin transaction fee will be $2.47 (estimated), which has been deducted from the payout totals."

I think that would be a decent approach. Though we'd also want to subtract the moderator fee (which is also estimated, since the moderator can change it). 

I'll file an issue for it, it's not something we need to do immediately.

PR #536, opened 2017-06-28
Remove extra return.

jjeffryes:


rmisio:
This is handled in a upcoming PR.

PR #544, opened 2017-06-30
Purchase patch

### LINKED ISSUE #533: The Pay button in the Purchase flow is not advancing to the next state
When you purchase an item, the PAY button does not change state:

image

Clicking it again will get you this fun scenario:

image

jjeffryes:
This fixes a variety of small issues.
- adds usecache to the moderators so they load faster at the cost of their data being potentially out of date.
- disables clicking on the pending button.
- fixes a bug where shipping options were sorted by price in the template, causing the first one selected in the view to not always match the first one in the template.
- merges in the changes for the purchasing code from #529 and fixes #533. If this PR is merged after 529 it will fix 533.


jjeffryes:
Closed, I will make a new version that syncs with the other PRs.

PR #552, opened 2017-07-06
Fix fee levels

jjeffryes:
This will fix the fee levels to always use PRIORITY, NORMAL, and ECONOMIC, instead of low, medium and high.

It will also correct the advanced tab to show errors with the fee levels on save, and homogenizes the language used so it always matches the server to avoid confusion.

### DIFF
--- a/js/languages/en-US.json
+++ b/js/languages/en-US.json
@@ -261,8 +261,7 @@
       "statusSaveFailed": "Unable to save general settings.",
       "saveErrorAlertTitle": "Unable to save general settings",
       "searchProvider": "Search Provider",
-      "restoreDefaultProvider": "Restore default",
-      "searchProviderError": "This is not a valid url for a search provider"
+      "restoreDefaultProvider": "Restore default"
     },
     "pageTab": {
       "helperAbout": "Displays on your Home page",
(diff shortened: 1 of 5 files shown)


js/languages/en-US.json
```diff
@@ -1363,14 +1362,18 @@
     "provideReview": "Please provide a review.",
     "provideRating": "Please select a rating."
   },
+  "localStorageModelErrors": {
```
rmisio: The model name is localSettings. Can the translation key be consistent with that?

jjeffryes: Done.

js/models/wallet/Spend.js
```diff
@@ -117,7 +117,8 @@ export default Spend;
 export function spend(fields) {
   const attrs = {
     currency: app && app.settings && app.settings.get('localCurrency') || 'BTC',
-    feeLevel: app && app.localSettings && app.localSettings.get('defaultTransactionFee') || 'high',
+    feeLevel: app &&
+    app.localSettings && app.localSettings.get('defaultTransactionFee') || 'PRIORITY',
```
rmisio: @jjeffryes Could you ensure this PR passes lint. If so, go ahead and merge it.

jjeffryes: Will do.

PR #560, opened 2017-07-11
Tor

### LINKED ISSUE #315: Add Tor Controls in the Client
## User Story
As a user, I want to be able to use TOR, and have the client give me the option to use TOR when it is available. I should be able to control which connections use TOR, and when.

### All Bundled Startups
- on startup, the client will use the status message process to start and stop the server and get an error code indicating if the server can access TOR.

### New Bundled Connection Startup
- if the server can use TOR, and the current connection has never had the "use TOR" flag set, the Manage Servers modal will open, with the current connection edit modal open. 
- a message will be shown asking the user if they want to set this connection to use TOR.
- there will be a "Use TOR" checkbox.
- there will be a "TOR SOCKS5 Proxy" text input field. It will be pre-filled with the detaul value.
- the proxy input may be hidden if the "Use TOR" checkbox is not checked.
- when the connection is saved, it will attempt to start the server with the TOR flag set.

### Existing Connection Startup
- if the server can use TOR, and the current connection has been set to use TOR, the client will start the server with the TOR flag set.
- if the server cannot use TOR, and the current connection has been set to use TOR, the client will show the Manage  Servers modal, with an error message stating that the server cannot be started in TOR mode. 
- the user can edit the connection to not use TOR, create a new connection to a different server, use a different existing connection, or fix the issue with the server and try again.

### Editing a Connection
- any time the Use TOR value is changed for a connection, a warning should be shown explaning that any account that has been used in non-TOR mode has associated its IP Address with its peerID. If the same server is used in TOR mode later, that association can be used to match their TOR activity with their IP address. 

### TOR Mode
- while using TOR, there should be a visual indicator that the client is in TOR mode.
- while using TOR, all connections from the client to other servers should go through the connected server node or through the TOR SOCKS 5 proxy (for example, externally hosted images in a field that allows HTML code).
- when a user in TOR mode clicks on a link that will be opened in an external browser, ideally a warning with cancel and confirm buttons (and maybe a "don't show this warning again" checkbox) would be shown, since the link may not open in their TOR browser.
    - it would be nice if the warning had an option to copy the href of the link so it can be examined and pasted in a TOR browser if safe.

rmisio:
This change-set wires in Tor controls to the client.

closes #315 

To test:
=====
1. Ensure Tor will run on your machine without authentication
a. Download the Tor browser.
b. Start up the Tor browser.
c. Attempt to start your server with the --tor flag. If it works, you'll see this in the terminal (`[Execute] [NOTICE] Using Tor exclusively`) and you're free to proceed to step 2. If it doesn't work, you'll likely see some type of auth error in the terminal. In that case, please ping @rmisio for further debugging.

The following steps will allow your client to masquerade as the bundled app and test Tor scenarios specific to the bundled (installer) version of the app.
 
2. Ensure no OB servers are running.
3. Start-up the client and run `localStorage.clear()` in the console.
4. Shut-down the client.
5. Download the [latest Server release](https://github.com/OpenBazaar/openbazaar-go/releases/) for your OS. Rename it to `openbazaard` on Linux or Mac, `openbazaard.exe` on Windows. Ensure the file is exectuable (`chmod +x on Mac and Linux`)
6. Edit main.js with the path to the executable in the previous step. Find the first line, comment it out and add the second line.

```
const serverPath = '/path/to/executable/'; // do not include filename, do include trailing `/`
```

4. Start up the client. It will create a server configuration and attempt to connect to the built-in server that you downloaded.

If you are able to set-up a remote server, it would be a really good test to toggle your connection between a remote server and the built-in server and ensure that the connection is properly in Tor mode when you configure it (check 'use Tor') to be.


cpacia:
Not seeing the tor icon in the url bar

rmisio:
@cpacia The tor indicator will show up when the following conditions are met:

1.) Your are connected to a server that was started in tor mode and that server is returning `tor: true` via the `ob/config` api.
2.) In the client, the server configuration for that connection is configured to use Tor:

image

Are both those conditions met for you?

cpacia:
Oh duh that's right I have to set it in the config

cpacia:
image

Ok this is what I wanted to show... the icon is a bit choppy on my system.

rmisio:
I think I'll leave choppy icons to @morebrownies . The client is sizing the icon down to 13px to try and match the design. Do you want to play around with maybe bumping it up or down a pixel or two and seeing if that helps?

(It may just be an overly detailed icon that's not designed to be shown that small)

morebrownies:
Ok, I'll try to get around to testing this later today.

morebrownies:
@rmisio @cpacia  I'm attempting to run my server with Tor and I'm running into this issue:

> | => go run openbazaard.go start -t --tor
OpenBazaar Server v0.6.4 starting...
12:43:40.067 [Execute] [ERROR] Authentication failed: invalid (empty) COOKIEFILE
Authentication failed: invalid (empty) COOKIEFILE
exit status 1

Any idea what is going on with my 🍪  file?

rmisio:
@morebrownies Yeah, it's the same thing Josh is going through. You're basically confirming that the Tor Browser needs specific configuration before being able to work with OB. @jjeffryes is working through putting together an instruction doc for it as I type (for Win and Mac). You may want to wait until he's done and work off that.

Basically, you need to configure your Tor Browser to use a password and give that password to the OB config.

If you want the cliff's notes version (for Mac):
- in terminal navigate to `/Applications/TorBrowser.app/Contents/Resources/TorBrowser/Tor`
- execute `./tor.real --hash-password mypassword`. This will output a hashed version of the password.
- edit `/Users/rob/Library/Application Support/TorBrowser-Data/torrc` by adding this line:
 `HashedControlPassword 16:0E016FA8F074C73460B7B55C1258A01CE5EA152282A20B8F7840948C3A`
(use the hashed password from the previous step)
- also add in this line `CookieAuthentication 0`. If there already was a line for `CookieAuthentication`, then replace it so it is set to 0.
- now edit your ob config, but give it the plain text password:
```
  "Tor-config": {
    "Password": "mypassword",
    "TorControl": ""
  },
``` 
- now you should be able to start the OB server in Tor mode. You'll need to bounce the Tor browser.

morebrownies:
Ok great thank you @rmisio I'll give this a shot

morebrownies:
I didn't get very far on this one @rmisio 

> | => cd /Applications/TorBrowser.app/Contents/Resources/TorBrowser/Tor
total 5576
drwxr-xr-x@ 3 Wolf  admin   102B Jul  2 10:28 PluggableTransports/
-rw-r--r--@ 1 Wolf  admin   3.5M Jul  1 07:53 geoip
-rw-r--r--@ 1 Wolf  admin   2.0M Jul  1 07:53 geoip6
-rwxr-xr-x@ 1 Wolf  admin   735B Jul  1 07:53 tor
-rw-r--r--@ 1 Wolf  admin   798B Jul  1 07:53 torrc-defaults
________________________________________________________________________________
> | /Applications/TorBrowser.app/Contents/Resources/TorBrowser/Tor @ Mikes-MacBook-Pro (Wolf) 
| => ./tor.real --hash-password mypassword
**-bash: ./tor.real: No such file or directory**

rmisio:
Eternal damnation! I gave you the wrong path!

Sorry @morebrownies, it's `/Applications/TorBrowser.app/Contents/MacOS/Tor`

js/views/modals/connectionManagement/Configurations.js
```diff
@@ -103,16 +127,61 @@ export default class extends baseVw {
...
+        links,
+      });
+
+      if (!this.$el.is(':visible')) {
+        // If the connection modal is not open, we'll open up the configuration form. The modal
+        // will be opened shortly upon the connection failure. If the user already had the modal
+        // open, we won't auto send them to the config form, since it may interrupt something else
+        // they may be doing.
+        this.trigger('editConfig', {
+          model: this.collection.defaultConfig,
```
jjeffryes: Can this also happen if you're using a configuration that isn't default, but is using Tor?

rmisio: No, we would only get a 'tor-not-available' error on the default (i.e. built-in) server, because that's the only one that we control and can start-up via the status command to determine if Tor is available. For stand-alon servers, it is the responsibility of the user to start the server in Tor mode. If they forgot to start the Tor browser and they start the server with --tor, then it will fail to start.

rmisio:
@jjeffryes this should be ready to code with code review updates. Also added in a 'Tor active' tooltip at Wolf's request.

PR #563, opened 2017-07-12
Fixed a number of styles and type issues

morebrownies:
**Changes**

- Changed body background color to #f0f0f0 to provide a bit more contrast between background layer and foreground layer  
- Made text in chat input slightly smaller 
- Remove transparency from chat window background (I felt this looked pretty bad one some pages, rather keep it simple for now) 
- Add back arrow to "Return to listing"
- Removed 950px button (since this hasn't been wired up yet, i'm thinking we should just make the grid responsive in the future instead of having to click a button)
- Cleaned up typography in the user card

### DIFF
--- a/styles/_theme.scss
+++ b/styles/_theme.scss
@@ -48,7 +48,7 @@ button,
 }
 
 .clrS {
-  background-color: $secondary;
+  background-color: #f0f0f0;
 
   input[type="radio"] + label:before,
   .fauxRadioBtn {
@@ -331,7 +331,7 @@ input[type=range][class~="clrS"] {
 }
 
 .chatOpen #appFrame #chatContainer {
-  background-color: $primaryGhost;
+  background-color: $primary;
   border-color: $border4;
 }
 
--- a/styles/_theme.scss
+++ b/styles/_theme.scss
@@ -31,24 +31,8 @@ button,
   }
 }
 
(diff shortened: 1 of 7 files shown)


morebrownies:
This is ready for review whenever you get a chance

styles/_theme.scss
```diff
@@ -48,7 +48,7 @@ button,
 }
 
 .clrS {
-  background-color: $secondary;
+  background-color: #f0f0f0;
```
jjeffryes: @morebrownies what's going on here? The .clrS class is literally the "make this the color of $secondary" class. 

If the default secondary color should be different, the place to change that is in the _variables.scss file. 

To make it the same as #f0f0f0, change it to:

$secondary: adjust-color($primary, $lightness: -6%);

styles/_theme.scss
```diff
@@ -331,7 +331,7 @@ input[type=range][class~="clrS"] {
 }
 
 .chatOpen #appFrame #chatContainer {
-  background-color: $primaryGhost;
+  background-color: $primary;
```
jjeffryes: If this is removed, you can also remove the $primaryGhost variable from the _variables.scss file, and the .clrPG class from the theme, since the chat container is the only place primaryGhost is currently used.

PR #566, opened 2017-07-14
Changes for Milestone 3

jjeffryes:
- added some more data to doorbell
- changed doorbell to be run from start.js
- hid the NSFW, Visual Effects, and Theme elements in settings
- bumped the milestone number in the milestone tag to 3
- bumped the version number in package.json to 2.0.5

This should not be merged until all other changes for Milestone 3 have been merged.


js/start.js
```diff
@@ -404,6 +404,45 @@ function onboardIfNeeded() {
   return onboardIfNeededDeferred.promise();
 }
 
+function addDoorbell() {
```
rmisio: Can we add this to the startup folder (either in the index file or it's own file)? That folder was created so that start.js doesn't get ridiculously huge and confusing with all these little loose ends.

jjeffryes: Yes, I'm planning to break it out today. That's part of why it's still WIP.

jjeffryes:
This is ready @rmisio

rmisio:
@jjeffryes Do you want to add TODO to the menu item shortcuts which are currently 'ctrl + ?'?
 Seems like something we could circle back to.

jjeffryes:
> @jjeffryes Do you want to add TODO to the menu item shortcuts which are currently 'ctrl + ?'?
Seems like something we could circle back to.

That's in now. I created an issue for the shortcuts. I went through the v1 shortcut code, I don't think it would be a huge task to put in, but it's big enough it should be separate from this, and isn't required for this milestone.

styles/_base.scss
```diff
@@ -368,6 +374,43 @@ svg {
   }
 }
 
+#ov1 #doorbell #doorbell-container {
```
rmisio: Should this really be in base? I would think this should get it's own file in the module folder -or- if you think it's too small, I suppose we could start a modules/_misc file (similar to components/_misc) where module stylings that are too small for their own file could go.

PR #568, opened 2017-07-17
Add Handle and @ Sign in User Page

jjeffryes:
This will add the @ in front of the handle when a listing is clicked on, and will add the handle with an @ when the tab is changed.


rmisio:
Closing because there will be a new PR to handle this given the updated reqs.

PR #570, opened 2017-07-17
Onboarding

### LINKED ISSUE #510: Add Onboarding
## User Story
As a user, when I launch the bundled application for the first time, I want to be guided through adding vital data and settings to the application so I can use it immediately.

## Functional Requirements
- [ ] The onboarding screens should only show if the user is launching the application for the first time.
- [ ] We should use the presence of Settings and Profile data to determine if onboarding is needed, so it doesn't get shown when a node that has been set up is accessed from another client.

### Splash Screen
- [ ] Show a splash screen with a Get Started button.
- [ ] The Get Started button will activate the next screen.
- [ ] The Connection Management button will launch the Connection Management modal.
    - [ ] The current connection should be shown, otherwise the user will need to open the modal to see what their current connection is.

### Information Screen
- [ ] This shows a subset of the forms in Settings General and Page.
- [ ] The functionality of the form elements should be identical to those in Settings.
- [ ] Ideally, the avatar would have the crop and zoom controls. If this turns out to be a lot of work, it can be deferred. 
- [ ] If next clicked and a field has an error, show an error message above it like in Settings (note that it will need to be validated without saving).
- [ ] The name should have a placeholder instead of showing a random name.
- [ ] The language should be set from the OS, without the user interacting with it.
- [ ] NOTE: we will add a back button to this screen. 

### Terms of Service
- [ ] This screen is just for showing the Terms of Service.
- [ ] The data from the information screen should not be saved until the I Agree button is clicked.

### Tooltips
- [ ] When onboarding closes, the app will go to the user's page.
- [ ] A tooltip for the Discover/Search button will show. It should only show if the user has just exited onboarding, and will stop showing when they click the search button.
- [ ] A tooltip for the user page controls will show if the user has just exited onboarding. 
NOTE: the app will track if the tooltips should be shown in the session. If the user restart the app they will not be shown again.
- [ ] The second state of the user page tooltip has various buttons that launch specific tabs on the settings modal. 
NOTE: Mike will update the design for the tooltip.

## Data

- [ ] the server will be updated to make name a required field.

rmisio:
closes #510


rmisio:
@jjeffryes this is ready to review when you're ready.

styles/components/_onboarding.scss
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    .infoScreen {
+      form {
+        width: 375px;
+        margin: 0 auto;
+
+        .select2 {
+          // without this the select2 dropdowns exceed the width of the form on re-render
+          width: 375px !important;
```
jjeffryes: I tested it with this property turned off, and the select2 elements always set themselves to the right width?

js/templates/modals/onboarding/onboarding.html
```diff
@@ -0,0 +1,137 @@
...
+      <div class="clrT2 txB"><%= ob.polyT('onboarding.pageOfPage', { startPage: 2, totalPages: 2 }) %></div>
+    <% } %>
+  </div>
+  <hr class="clrBr" />
+  <div class="mainContent">
+    <% if (ob.screen === 'intro') { %>
+    <div class="flexCent">
+      <div>
+        <div class="txCtr rowSm">
+          <span class="txUp txB clrT2"><%= ob.polyT('onboarding.introScreen.introLine') %></span><i class="btcIcon margRTn"></i>
```
rmisio: I gave it some space via CSS since adding in an actual space gives it more space than what the design reflects (haven't pushed yet).

js/languages/en-US.json
```diff
@@ -833,6 +838,34 @@
...
   },
+  "onboarding": {
+    "pageOfPage": "%{startPage} of %{totalPages}",
+    "profileFailedSaveTitle": "Unable To Save Your Profile",
+    "settingsFailedSaveTitle": "Unable To Save Your Settings",
+    "settingsFailedSaveAvatar": "Unable To Save Your Avatar",
+    "introScreen": {
+      "connectionLbl": "Connection:",
+      "btnChange": "Change",
+      "introLine": "A free marketplace. No fees. No restrictions. Earn Bitcoin",
```
rmisio: I was just matching the design on that one. If the design change, I imagine @morebrownies could polish this up.

styles/components/_onboarding.scss
rmisio: Initially it was the right width, but then if I clicked next and back, it would be the wrong with. But... I can't seem to reproduce that anymore. I updated the CSS.

js/languages/en-US.json
morebrownies: Yeah we should add a period at the end of "Earn Bitcoin". I missed that one.

js/start.js
```diff
@@ -394,7 +334,7 @@ const onboardIfNeededDeferred = $.Deferred();
 
 function onboardIfNeeded() {
   isOnboardingNeeded().done((onboardingNeeded) => {
-    if (onboardingNeeded) {
+    if (onboardingNeeded || true) {
```
jjeffryes: Don't forget to remove the debugging code.

PR #575, opened 2017-07-20
Various Fixes to Search, Listings, Etc.

### LINKED ISSUE #507: Clicking a tag on a listing detail screen should take you to search results
screenshot 2017-06-02 14 45 30

When I click **#ski**, it should push me to global search results page and return all listings that match the **#ski** tag.

### LINKED ISSUE #567: Search filter not sticking
Search filter dropdown is not setting and sticking.

image

jjeffryes:
This fixes a variety of small issues. 

- It fixes the issue with the sort by drop down not being applied in the search results.
- It fixes issues with the formatting of ratings, and makes the formatting a template helper so it's not duplicate code.
- It prevents the listing screen from briefly showing 0 for the rating as it's being loaded.
- It adds error handling for when a review is loaded over sockets and has an error.
- The title text on the listing card is truncated earlier if there is an avatar so it doesn't overlap.
- It adds placeholder text to the search bar.
- It reduces the size of the drop downs in the search screen.
- It adds an @ in front of handles in several places where it was missing.
- It makes tags clickable, clicking them will launch a search for that tag.

Closes #507
Closes #567


rmisio:
@morebrownies How do we want to represent when a listing has 0 reviews. What's in place now looks kinda odd to me for whatever reason. But, if you're ok with it, I'll be fine as well.

image

image

rmisio:
Is there a mismatch in the UI here. The lowest ratings filter is '1 star and up', but you could see a listing with an average rating of 0?

image

rmisio:
Similar to above, I filter for '5 star' listings and get ones with ratings of 4.x:

image

Is this a matter of maybe the server using 0 based ratings and the UI wanting to reflect 1 based ratings?

If so, I would say the UI should just increment the ratings by 1 in the display... but then in the example above, the listing would have an average rating of 5.67. Isn't 5 the most it could be? (confused)

morebrownies:
> @morebrownies How do we want to represent when a listing has 0 reviews. What's in place now looks kinda odd to me for whatever reason. But, if you're ok with it, I'll be fine as well.

When a listing has 0 reviews, it should display like this: ⭐ 0 (0)

jjeffryes:
@rmisio re: the stars, it's entirely up to the server how that filter works, and what is displayed for each option. So we'll need @tyler-smith to update that.

PR #576, opened 2017-07-20
Passing parent level errors into variant view

### LINKED ISSUE #498: Edit Listing Doesn't Clear Variant Errors
When a variant choice name is too long, and triggers an error on save, the error is not cleared even if the variant choice is shortened or deleted.

Steps to reproduce:
1. enter a variant name that is too long
2. try to save, get an error
3. remove or edit the too long name

image

The error checking seems to break if you only add one choice to a variant. No error is shown on save, but the variant with one choice won't actually be saved when the listing is re-opened after editing.

If you add one choice, then save, adding choices that are over 40 characters will not show an error for the variant that was saved with one choice (presumably because it was never actually saved)

rmisio:
closes #498

### DIFF
--- a/js/templates/modals/editListing/variant.html
+++ b/js/templates/modals/editListing/variant.html
@@ -4,11 +4,20 @@
 </div>
 <div class="col6 simpleFlexCol">
   <%
+    let variantsErrs = [];
+    
     Object.keys(ob.errors).forEach(errKey => {
       if (errKey.startsWith('variants[') && errKey.endsWith('].name')) {
-        print(ob.formErrorTmpl({ errors: ob.errors[errKey] }));
+        console.log('boom boom');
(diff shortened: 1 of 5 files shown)


rmisio:
One thing of note is a weird select2 issue. If you are focused in the select2 tagging field and you click out of it, it scroll you back to the field.

So... if you are focused in the variant Choices field and you add a choice, then scroll up and click Save, it doesn't register the click on the Save button. Instead select2 uses that to scroll you back to the Choices field.

What you really need to do when you are focused in the Choices field is when you are done, click out of it, then scroll up and click Save.

Just another weird quirk of select2. I'll keep it in mind and look to circle back to it. We've already modified their source to fix one weird scroll issue, so it's possible we could fix this one. It may have some valid use when you're using select2 as a drop down instead of a tagging field (but I have no idea if that's the case).

PR #580, opened 2017-07-23
Added packages required to build on OSX

joaocc:
These packages were needed when trying to build on OSX 10.12.6.
```
$ node --version && npm --version
v6.11.0
5.3.0
```


jjeffryes:
@joaocc as far as I can tell, this isn't required to build on Mac. What are you seeing that leads you to think this change is necessary?

jjeffryes:
This appears to not be necessary.

PR #585, opened 2017-07-25
Transmits Click Through Select2

jjeffryes:
This adds a click through option to select2, which will transmit clicks through to the element clicked on. This fixes the bug in edit listing where clicking save is blocked if you were in a variant, tag, or category.


rmisio:
@jjeffryes As we discussed, I tweaked the triggered click to happen in a timeout so it doesn't result in a tag being blown away. Also, I got rid of the clickThrough options because it looks like this option just works with it always being on for all types of dropdowns. If we do notice any weird behavior, in the process of debugging I would comment out the timeout to isolate whether this is the culprit (if so, we could always restore it as an optional thing).

PR #589, opened 2017-07-26
Social accounts

jjeffryes:
Adds social accounts to the settings/page tab.

Each account has a type and a username. Neither one is validated, they can be any string.

If the username is a url, it will be clickable when shown in the userpage/about tab.


js/collections/profile/SocialAccounts.js
```diff
@@ -1,16 +1,8 @@
-import { guid } from '../../utils';
```
rmisio: Could we actually revive this? Sorry, I didn't document it at all, but this is a common pattern I'm using for the Profile and Settings nested collections for models that don't have an ID. The reason is that we are working off of clones and after a save, we merge the cloned data back into the main models (`app.profile`, `app.settings`). If those models don't have IDs, then the "change" events think they are new. For example, let's say you had save out 3 social accounts and now you go back to settings and add a fourth. If those guys don't have id's, then the people listening on the add event of `app.profile.socialAccounts` will get 4 add events fired, one for each one, because without an ID, it can't tell it already had the 3.

(just a quirk of how we're working with a clone and then merging it back into the main model)

js/views/modals/Settings/Page.js
```diff
@@ -21,6 +22,11 @@ export default class extends baseVw {
 
     this.profile = app.profile.clone();
     this.listenTo(this.profile, 'sync', () => app.profile.set(this.profile.toJSON()));
+
+    this.socialAccounts = this.createChild(SocialAccounts, {
+      collection: this.profile.get('contactInfo').get('social').clone(),
```
rmisio: You don't need to clone here. The profile model is already clones (including nested objects) a couple lines above.

rmisio:
image

Would we want to style the 'Add additional' link to match the other "Add" actions (Create Listing has quite a few)?

image

rmisio:
If someone has reached the max amount of social accounts, would it be better to just hide the "Add additional" link rather than making it clickable with an error message? (that's how all these Add links work in Create Listing)

image

FWIW - if I find myself in the scenario above and then click Save, clicking Add additional again, does nothing. (This would be a moot point if we just hid that link)

js/templates/userPage/home.html
```diff
@@ -56,8 +56,10 @@
         <% } %>
         <% ob.contactInfo.social.forEach(function(account) { %>
           <div class="listItem">
-            <div><%= ob.polyT('properNames[account.type]') %></div>
-            <div class="clrT2"><%= account.username %></div>
+            <div><%= account.type %></div>
+            <div class="clrT2">
+              <% print(ob.is.url(account.username) ? `<a href="${account.username}">${account.username}</a>` : account.username) %>
```
rmisio: Could we add a `data-open-external` to that anchor? If they include a non-ob protocol (e.g. http://), then it could detect external, but if they just put `www.espn.com`, it thinks that's an internal link.

js/views/modals/Settings/Page.js
```diff
@@ -143,6 +149,11 @@ export default class extends baseVw {
   save() {
     const formData = this.getFormData();
 
+    // set the model data for the social accounts
+    this.socialAccounts.setCollectionData();
+    // set the data from the cloned social account
+    this.profile.get('contactInfo').get('social').set(this.socialAccounts.collection.models);
```
rmisio: If you get rid of the unneeded clone above, then this line isn't needed.

js/views/modals/Settings/Page.js
jjeffryes: I'll take that out, it was to deal with the weird things happening because the ids didn't match, now that they do it's not needed.

jjeffryes:
@rmisio this was updated, all the comments have been addressed, thanks for catching those.

@morebrownies this is ready to look at.

rmisio:
@morebrownies I'm going to merge this one. If you find anything, please feel free to log some bugs or add a polish PR.

PR #592, opened 2017-07-28
Purge button

jjeffryes:
This adds a purge button to settings/advanced that will remove all shared files from your server. This addresses concerns of users in areas that prohibit distribution of some content, if those users inadvertently view prohibited content they can clear their cache so it's not shared with other users over IPFS.

### DIFF
--- a/js/languages/en-US.json
+++ b/js/languages/en-US.json
@@ -369,9 +369,9 @@
         "peersHelper": "Peers you're connected to",
         "peersBtn": "Show Connected Peers",
         "purge": "Purge Shared Files",
-        "purgeHelper": "Delete files you are sharing with others",
+        "purgeHelper": "Delete cached files you are sharing with others",
         "purgeBtn": "Purge Files",
-        "purgeToolTip": "OpenBazaar shares content you have viewed with other users. If you have viewed text or images you don't want to share this button will delete them. This will not affect any files that belong to you.",
+        "purgeToolTip": "OpenBazaar shares cached content you have viewed with other users. If you have viewed text or images you don't want to share this button will delete them. This will not affect any files that belong to you.",
         "purgeError": "There was an error, your cached files may not have been deleted.",
         "purgeComplete": "Your cached files have been deleted."
       },
--- a/js/languages/en-US.json
+++ b/js/languages/en-US.json
@@ -329,8 +329,6 @@
       "helperWindowControls": "Maximize, Minimize & Close controls",
       "helperSaveMetaData": "Store additional transaction information",
       "helperDefaultFee": "Priority is recommended for the fastest transactions",
-      "helperServerManagement": "Switch and manage server connections",
-      "helperServerPeers": "Peers you're connected to",
       "smtp": {
         "sectionName": "Email Integration (SMTP)",
         "testAction": "Test Integration",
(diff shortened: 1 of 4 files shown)


js/views/modals/Settings/Advanced.js
```diff
@@ -47,6 +48,35 @@ export default class extends baseVw {
...
+        openSimpleMessage(
+          app.polyglot.t('settings.advancedTab.server.purgeError'),
+          failReason);
+      })
+      .done(() => {
+        this.getCachedEl('.js-purgeComplete').removeClass('hide');
+      });
+  }
+
+
```
rmisio: Don't forget the super.render().

At some point, I think I'll try and write a test that would check all views and ensure that if they are using getCachedEl, that they are calling super.render(). It's really easy to forget.

js/languages/en-US.json
```diff
@@ -358,9 +356,17 @@
...
-        "btnShowManagementAction": "Connection Management",
-        "btnShowPeersAction": "Show Connected Peers"
+        "peersHelper": "Peers you're connected to",
+        "peersBtn": "Show Connected Peers",
+        "purge": "Purge Shared Files",
+        "purgeHelper": "Delete files you are sharing with others",
+        "purgeBtn": "Purge Files",
+        "purgeToolTip": "OpenBazaar shares content you have viewed with other users. If you have viewed text or images you don't want to share this button will delete them. This will not affect any files that belong to you.",
+        "purgeError": "There was an error, your cached files may not have been deleted.",
+        "purgeComplete": "Your cached files have been deleted."
```
rmisio: You may want to check with @cpacia to know if 'ob/purgecache' returns after the purge is complete or if it just kicks off a purge, returns immediately, and the purge finished at some time. That's how the publish api works. If the purge API works the same way, I think we'd want to tweak the messaging since `Your cached files have been deleted.` wouldn't be completely accurate.

js/views/modals/Settings/Advanced.js
rmisio: It might not be a bad idea to pass `isPurging: this.purge && this.purge.state() === 'pending'` into the template and then use that in the template to decide whether the `processing` class should be applied to the button.

Otherwise, if you re-render during a purge (e.g. user clicks Save), then the spinner state is lost and the user might just kick off another purge.

jjeffryes:
@rmisio this is updated. I also updated the save button with a saving state, so it will maintain it's state the same way the purging button will, and updated the save button to use getCachedEl.

Chris confirmed the purge api returns after it has completed. It usually only takes 40-50ms, but in theory could be longer.

js/views/modals/Settings/Advanced.js
```diff
@@ -95,40 +135,36 @@ export default class extends baseVw {
...
     if ($firstErr.length) $firstErr[0].scrollIntoViewIfNeeded();
   }
 
-  get $btnSave() {
-    return this._$btnSave ||
-      (this._$btnSave = this.$('.js-save'));
-  }
-
   render() {
     loadTemplate('modals/settings/advanced.html', (t) => {
```
rmisio: Don't forget the super.render() if using getCachedEl.

PR #609, opened 2017-08-04
Update OpenBazaar 2 icons for OSX and Windows

hoffmabc:
This updates the icons to the new white ones to differentiate from OpenBazaar 1.

I'm going to test the build first before we merge @jjeffryes.


imgs/openbazaar2.icns
```diff
@@ -0,0 +1,22 @@
+How to use this icon:
```
jjeffryes: @hoffmabc do we need the how to text here?

hoffmabc: Fixing this

hoffmabc:
Ok @jjeffryes this one is fine regardless of what Travis says. I killed him.

PR #611, opened 2017-08-04
Add button to show block data.

jjeffryes:
This adds a button to settings/advanced that will show the current best hash and block height for the user's transactions.

The dialog created has a copy button that will copy the data to their clipboard.


js/templates/modals/settings/advanced.html
```diff
@@ -144,6 +144,24 @@ <h2 class="h4 clrT"><%= ob.polyT('settings.advancedTab.transaction.sectionName')
...
+                <%= ob.polyT('settings.advancedTab.server.blockData') %>
+              </label>
+              <div class="clrT2 txSm">
+                <%= ob.polyT('settings.advancedTab.server.blockDataHelper') %>
+              </div>
+            </div>
+            <div class="col9">
+              <div class="flexVCent gutterH">
+                <%= ob.processingButton({
+                className: `btn clrP clrBr clrSh2 js-blockData ${ob.isPurging ? 'processing' : ''}`,
```
rmisio: shouldn't this be `isGettingBlockData`?

jjeffryes: Yep.

PR #612, opened 2017-08-04
Resync button

jjeffryes:
This adds a resync button that will resync the wallet's transactions.


js/views/modals/Settings/Advanced.js
```diff
@@ -48,6 +49,29 @@ export default class extends baseVw {
...
+  resynchronize() {
+    this.getCachedEl('.js-resync').addClass('processing');
+    this.getCachedEl('.js-resyncComplete').addClass('hide');
+
+    this.resync = $.post(app.getServerUrl('wallet/resyncblockchain'))
+      .always(() => {
+        this.getCachedEl('.js-resync').removeClass('processing');
+      })
+      .fail((xhr) => {
+        const failReason = xhr.responseJSON && xhr.responseJSON.reason || '';
```
rmisio: If you're aborting this guy in remove, then make sure to return out of this fail handler if its aborted, otherwise you'll have the dialog come up.

jjeffryes: Good catch. I didn't have it in because originally it looked like the call could take several minutes, and the user should know if it failed after they left the page. 

Now that it's working right, the call takes seconds or less, but I forgot to go back and add the abort check.

PR #617, opened 2017-08-06
Various HTML and CSS Fixes

### LINKED ISSUE #615: Add tooltips to all icons in the title bar
I attempted to get this working but ran into a lot of issues when using the tooltip class. It seems tooltips aren't playing friendly with `<a>` objects. 

Below is the text the tooltips should display on hover:

screenshot 2017-08-06 11 25 56

1. Close
2. Minimize
3. Maximize
4. Back
5. Forward
6. Refresh

screenshot 2017-08-06 11 25 59

1. Discover
2. Wallet
3. Notifications
4. Navigation

### LINKED ISSUE #616: Need Bitcoin? link should point to correct URL
Right now it points to https://openbazaar.org and should point to https://www.openbazaar.org/bitcoin/

jjeffryes:
This will fix a variety of HTML and CSS issues.
- the milestone label is removed
- the feedback button will slide over when the chat sidebar is open
- tooltips were added to the nav bar buttons, with adjustments to make the align properly
- in mac mode, the tooltips on the left and right are adjusted so they don't extend outside the app
- the link to get bitcoins in the wallet is fixed to use the right url

Closes #615 
Closes #616


morebrownies:
The close button on mac gets cut off, but otherwise this looks good! 

screenshot 2017-08-06 21 18 00

jjeffryes:
> The close button on mac gets cut off, but otherwise this looks good!

That's the best compromise I came up with, if it moves farther to the right the "arrow" triangle at the top looks weird and/or sticks out past the left border of the box.

I can try nudging it over a few more pixels.

morebrownies:
If needed, we could also consider ditching the tooltips on the window controls. Most apps don't have them since those buttons are pretty well understood.

jjeffryes:
@morebrownies take a look at the update, I managed to get the tooltip on screen without the arrow going past the left side.

morebrownies:
Yeah looking better. This may get weird with translations though. If you feel long translations may break it, I'd be ok with removing the tooltips on window controls.

rmisio:
> Yeah looking better. This may get weird with translations though. If you feel long translations may break it, I'd be ok with removing the tooltips on window controls.

They seem unnecessary to me. These are standard OS controls and I don't see other apps showing tooltips for them. We detect your OS and decide wether to use Win or Mac style controls, so they should look familiar to you. (the one exception may be linux, which is inconsistent across different flavors of linux. We default to win style, but it is possible your flavor uses Mac style)

rmisio:
Yeah, translations could be tricky here, because if they get long the tips are cut off (not sure how long reasonably they could get):

image

image

rmisio:
When you hover, this icons get an underline.

image

@morebrownies Is this what we want?

js/templates/pageNav.html
```diff
@@ -3,22 +3,28 @@
...
       <div class="windowControlsHolder">
         <div class="windowControls">
-          <a class="navClose js-navClose">
-            <span class="ion-ios-close-empty"></span>
+          <a class="navClose js-navClose toolTip toolTipNoWrap" data-tip="<%= ob.polyT('pageNav.toolTip.close') %>">
+            <i class="ion-ios-close-empty"></i>
           </a>
-          <a class="navMin js-navMin">
-            <span class="ion-ios-minus-empty"></span>
+          <a class="navMin js-navMin toolTip toolTipNoWrap" data-tip="<%= ob.polyT('pageNav.toolTip.min') %>">
```
rmisio: toolTipNoWrap extends tooltip, so you don't need both.

jjeffryes: D'oh. Good point.

cryptojunkie:
I didn't even notice the underline. We shouldn't need them.

cryptojunkie:
Blah responding from phone

jjeffryes:
> When you hover, this icons get an underline.

That shouldn't be there.

PR #627, opened 2017-08-07
Notifications

### LINKED ISSUE #489: Add Notifications Functionality
As a user, I want to receive notifications of important events, and be able to see all of those notifications in a list, from newest to oldest. The notifications I haven't seen before should be marked, so I know they are new.

## Functional Requirements
- [ ] When a new notification is received, and the app is not in focus, the user should see a native alert with the same data. and a sound should play.
- [ ] When a new notification is received, the unread notification count in the page nav should go up by 1.
- [ ] The unread count should always show how many notifications are unread.
- [ ] Ideally, the app icon in the OS will show the unread count.
- [ ] When the notification button is clicked, a list of 20 notification should be shown.
- [ ] When the user scrolls to the end, load the next page of notifications. 
- [ ] When a new notification comes in, it is added to the top.
- [ ] Notifications that have not been viewed before are marked as unread.
- [ ] When the notifications are closed, call the API to mark all notifications as read, and update the client data to consider those notifications read. 
- [ ] There are navigation links/tabs
    - [ ] The all tab shows all notifications
    - [ ] The Orders tab shows only order related notifications
    - [ ] The Followers tab shows only Followers related notifications
- [ ] REMOVE: the mark all as read button.
- [ ] REMOVE: The Delete All button

## Data:
- [ ] The notifications API will need to take a filter for the type of notification
- [ ] The notifications API will need to take a last ID and a page size. 
- [ ] The mark as read API will need to take an "all" option
- the notifications API will return peer IDs, the avatars, hashes and other data for each user will need to be loaded asynchronously (with use cache as true), and updated as they come in.

rmisio:
closes #489 

If your server has not been wiped since this server commit `11e54111b2041e7716e7a212a771dd6c253d62f5`, you will want to wipe your data, otherwise your old notifications will not have the proper data the client is expecting.


js/utils/serverConnect.js
```diff
@@ -44,7 +44,8 @@ function log(msg) {
 
   const logMsg = `[SERVER-CONNECT] ${msg}${EOL}`;
   debugLog += logMsg;
-  ipcRenderer.send('server-connect-log', logMsg);
+  // "if (ipcRenderer)" needed to prevent the `npm test` from bombing.
```
rmisio: I really, really, really dislike having to modify a source file to accommodate a test, but ipcRenderer is undefined during testing and I can't seem to be able to stub it (using sinon or otherwise). I'll probably circle back to this, but it works for now.

morebrownies:
I'm not hearing a sound when a notification comes in. 

@rmisio has this been wired in yet?

rmisio:
> I'm not hearing a sound when a notification comes in.

I forgot about that. I'll wire that in.

@morebrownies @jjeffryes What was the reasoning in v1 we used a custom sound instead of letting the OS use it's designated notification sound?

morebrownies:
Typically a custom sound is used to make your app stand out from the bunch. 

Slack for example doesn't use the native sound and when I'm across the room I know what type of notification I'm getting just by the type of sound.

morebrownies:
We should make sure to wire up the sound to trigger for inbound chat messages too. 

I believe in V1 we only triggered the sound when the window is inactive, right?

rmisio:
> We should make sure to wire up the sound to trigger for inbound chat messages too.

There's a separate [issue](https://github.com/OpenBazaar/openbazaar-desktop/issues/628) for that. I'll circle back to that after this PR.

> I believe in V1 we only triggered the sound when the window is inactive, right?

Actually, looking at the code, we fired the sound irrespective of whether the window is active. Also, we fired the native notification no matter what as well. So, if we're going to match v1, then I'll tweak it so the sound and the native notifications come all the time.

morebrownies:
> There's a separate issue for that. I'll circle back to that after this PR.
👍 

> Actually, looking at the code, we fired the sound irrespective of whether the window is active. Also, we fired the native notification no matter what as well. So, if we're going to match v1, then I'll tweak it so the sound and the native notifications come all the time.

👍

js/views/notifications/Notificiations.js
```diff
@@ -0,0 +1,152 @@
...
+    return 'section';
+  }
+
+  className() {
+    return 'notifications clrBr border clrP clrSh1';
+  }
+
+  events() {
+    return {
+      'click .js-edit': 'onClickEdit',
```
jjeffryes: This line looks like it's not supposed to be here.

styles/modules/_notifications.scss
```diff
@@ -0,0 +1,113 @@
...
+      &.active, &:hover {
+        border-bottom-width: 1px;
+        border-bottom-color: inherit;
+      }
+    }
+  }
+
+  .tabContainer {
+    max-height: 400px;
+    max-height: 250px;
```
jjeffryes: Why is max-height here twice?

styles/modules/_notifications.scss
```diff
@@ -0,0 +1,113 @@
...
+        height: 200px;
+
+        .listFetcher {
+          width: 100%;
+          @include center();
+        }
+      }
+    }
+
+    .listFetcher {
```
jjeffryes: Why are there two places with rules with the listFetcher class? Is it sometimes not inside .fetcherContainer?

morebrownies:
A few more things I found:

1. The notification tray seems kind of tight (vertically) Could we increase `js-tabContainer tabContainer scrollBox` to min-height 320px?

2. The font size within notifications feels a bit large and could be bumped down one notch `.listBox .listItem` to font size 1.3rem?`

3. Links within notifications would probably be better set to green, they'll stand out a bit more.
screenshot 2017-08-08 13 58 58

4. The vertical line underneath the tab bar goes away when you scroll. It should stay fixed. See screenshot below:
screenshot 2017-08-08 13 59 18

morebrownies:
`No notifications found` displays in Orders tab when notifications exist 

screenshot 2017-08-08 14 26 13

morebrownies:
The link to `OpenBazaar Shirt` within the purchase notifications goes to

ob://QmcTMuGwpwzkc8hVqHe72AA2CDcuGYzvzZJWu98pf7tuot/store/openbazaar-shirt

instead of

ob://QmcFgSGtiNB1h4ycLWvPuYnvRrTJZcNLeuJdnynNC8vBW8/store/openbazaar-shirt

rmisio:
@jjeffryes @morebrownies This should be ready for re-review. Thanks for finding those things.

js/utils/index.js
```diff
@@ -235,3 +235,20 @@ export function getBlockChainTxUrl(txid, isTestnet) {
...
+
+export function playNotifSound() {
+  if (!document) {
+    throw new Error('The document object needs to be available to this function.');
+  }
+
+  if (!notifAudioEl) {
+    notifAudioEl = document.createElement('audio');
+    notifAudioEl.setAttribute('src', '../audio/notification.mp3');
+    console.log('silly');
```
jjeffryes: Remove your console log here.

js/utils/index.js
```diff
@@ -235,3 +235,20 @@ export function getBlockChainTxUrl(txid, isTestnet) {
...
+    window.silly = notifAudioEl;
```
jjeffryes: What's window.silly do?

morebrownies:
Ok great thank you. I'll circle back to this when I can.

morebrownies:
Looks great to me @rmisio. Thanks for getting those fixes in 👍

PR #629, opened 2017-08-07
Move Company in Purchase Address

jjeffryes:
This PR moves the company to before the address in 2 places:
1. the drop down in the purchase view
2. the order model

It also add a line break between each piece of the address in the order details template, to make the address format the expected way.

### DIFF
--- a/js/models/purchase/Order.js
+++ b/js/models/purchase/Order.js
@@ -34,12 +34,9 @@ export default class extends BaseModel {
 
   addAddress(sAddr) {
     // this will convert and set an address from the settings
-    const shipTo = sAddr.get('name');
-    const addressArray = [];
-    if (sAddr.get('company')) addressArray.push(sAddr.get('company'));
-    if (sAddr.get('addressLineOne')) addressArray.push(sAddr.get('addressLineOne'));
-    if (sAddr.get('addressLineTwo')) addressArray.push(sAddr.get('addressLineTwo'));
-    const address = addressArray.join(', ');
+    const company = sAddr.get('company');
+    const shipTo = `${sAddr.get('name')}${company ? `, ${company}` : ''}`;
+    const address = `${sAddr.get('addressLineOne')} ${sAddr.get('addressLineTwo')}`;
     const city = sAddr.get('city');
     const state = sAddr.get('state');
     const postalCode = sAddr.get('postalCode');
--- a/js/models/purchase/Order.js
+++ b/js/models/purchase/Order.js
@@ -35,8 +35,11 @@ export default class extends BaseModel {
   addAddress(sAddr) {
     // this will convert and set an address from the settings
     const shipTo = sAddr.get('name');
-    const address =
-      `${sAddr.get('addressLineOne')} ${sAddr.get('addressLineTwo')} ${sAddr.get('company')}`;
+    const addressArray = [];
+    if (sAddr.get('company')) addressArray.push(sAddr.get('company'));
+    if (sAddr.get('addressLineOne')) addressArray.push(sAddr.get('addressLineOne'));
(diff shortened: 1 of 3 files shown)


js/templates/modals/purchase/shipping.html
```diff
@@ -10,9 +10,9 @@ <h2 class="h4 required flexExpand"><%= ob.polyT('purchase.shippingTitle') %></h2
     <% ob.userAddresses.forEach((a, i) => { %>
     <option value="<%= i %>">
       <% let part1 = `${a.name}`;
+         part1 += a.company ? `, ${a.company},` : '';
```
rmisio: Should there be a comma before the city there?

image

Some commas seem to be missing here as well.

image

rmisio: I'm wondering if since nearly all the address fields are optional, if it would be better to push any that exist into an array and then join it with a comma?

js/templates/modals/orderDetail/summaryTab/orderDetails.html
```diff
@@ -50,7 +50,7 @@ <h2 class="tx4 margRTn"><%= ob.polyT('orderDetail.summaryTab.orderDetails.headin
           <% if (ob.order.shipping && ob.order.shipping.country !== 'NA') { %>
             <div><%= ob.order.shipping.shipTo %></div>
             <% if (ob.order.shipping.address) { %>
-            <div><%= ob.order.shipping.address %></div>
+            <div><%= ob.order.shipping.address.replace(/, /g, ',<br>') %></div>
```
rmisio: Would we want to remove the `,` from before the line break? Otherwise it's leaving a trailing comma at the end of the lines:

image

js/templates/modals/purchase/shipping.html
jjeffryes: Good idea.

js/templates/modals/orderDetail/summaryTab/orderDetails.html
jjeffryes: Good point. Technically I think the comma is supposed to be there, but it looks weird.

js/models/purchase/Order.js
```diff
@@ -34,9 +34,9 @@ export default class extends BaseModel {
 
   addAddress(sAddr) {
     // this will convert and set an address from the settings
-    const shipTo = sAddr.get('name');
-    const address =
-      `${sAddr.get('addressLineOne')} ${sAddr.get('addressLineTwo')} ${sAddr.get('company')}`;
+    const company = sAddr.get('company');
+    const shipTo = `${sAddr.get('name')}${company ? `, ${company}` : ''}`;
+    const address = `${sAddr.get('addressLineOne')} ${sAddr.get('addressLineTwo')}`;
```
jjeffryes: For some reason, the previous change to this broke the purchase call on the server, so I moved the company to be appended to the send to value, which is probably more correct.

js/templates/modals/purchase/shipping.html
```diff
@@ -9,16 +9,18 @@ <h2 class="h4 required flexExpand"><%= ob.polyT('purchase.shippingTitle') %></h2
...
-         part3 += a.postalCode ? ` ${a.postalCode}` : '';
+      <% const addr = [];
+         addr.push(a.name);
+         if (a.company) addr.push(a.company);
+         if (a.addressLineOne) addr.push(a.addressLineOne);
+         if (a.addressLineTwo) addr.push(a.addressLineTwo);
+         if (a.city) addr.push(a.city);
+         const state = a.state || '';
+         const code = a.postalCode || '';
+         const stateCode = `${state ? `${state} ` : ''}${code}`;
```
jjeffryes: There shouldn't be a comma between state and postal code, so those two are treated differently.

PR #634, opened 2017-08-07
All shipping region

rmisio:
This changeset handles the `ALL` shipping region by converting it to a list of individual countries in Listing.parse() and back to `ALL` in sync.


js/models/profile/Profile.js
```diff
@@ -1,10 +1,10 @@
-import BaseModel from '../BaseModel';
```
rmisio: This can be mostly ignored. I just re-organized these imports.

PR #635, opened 2017-08-07
Enhance feedback

jjeffryes:
Adds some extra data to the Doorbell feedback, like client and server version, whether the profile or settings loaded, and system RAM and CPU.

### DIFF
--- a/js/utils/feedback.js
+++ b/js/utils/feedback.js
@@ -1,34 +1,43 @@
 import app from '../app';
-
-const setFeedback = function () {
-  const profile = app.profile ? app.profile.toJSON() : {};
-  const contactInfo = profile.contactInfo || {};
-
-  window.doorbellOptions = {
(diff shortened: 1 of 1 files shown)


js/utils/feedback.js
```diff
@@ -1,34 +1,44 @@
...
   window.doorbellOptions = {
     appKey: 'lscnduocsmcCDtvh4DCZ4iQhGuCXZy4iexy7bIRa6wa5MFocLkSSutFU3zOii5k8',
-    name: profile.name,
-    email: contactInfo.email,
+    name: profile.name || 'name data missing',
+    email: contactInfo.email || 'email data missing',
     properties: {
-      peerID: profile.peerID,
-      vendor: profile.vendor,
+      settingsReadable: !!app.settigs,
```
rmisio: there is a typo here.

PR #639, opened 2017-08-09
Made buttons green in some places

morebrownies:
**CHANGES**

- Change "Save" buttons in settings to green color
- Changed "Save" button in create/edit listing to green color
- Changed "Return" button in transaction section
- Cleaned up some copy

screenshot 2017-08-09 10 37 18

screenshot 2017-08-09 10 35 33

screenshot 2017-08-09 10 35 10

### DIFF
--- a/js/templates/modals/orderDetail/orderDetail.html
+++ b/js/templates/modals/orderDetail/orderDetail.html
@@ -4,11 +4,10 @@
     // and componentize the css.
   %>
   <% if (ob.returnText) { %>
-  <div class="contentBox clrP clrSh3 clrBr clrT">
-    <a class="padSm gutterHSm overflowAuto margRSm flexVCent returnBox js-returnBox">
-      <span class="clrBr2 clrSh1 disc avatar flexNoShrink" style="<%= ob.getAvatarBgImage(ob.ownProfile.avatarHashes) %>"></span>
-      <span class="txUnl tx3 clamp clrT name"><%= ob.ownProfile.name %></span>
-      <span class="txU flexNoShrink tx6"><%= ob.returnText %></span>
+  <div class="btnStrip clrSh3">
+    <a class="btn clrP clrBr clrT js-returnBox">
+      <span class="ion-chevron-left margRSm"></span>
+      <%= ob.returnText %>
     </a>
   </div>
   <% } %>
--- a/js/templates/modals/settings/moderation.html
+++ b/js/templates/modals/settings/moderation.html
@@ -3,7 +3,7 @@
     <div class="flexHCent">
       <h2 class="h3 clrT">Moderation</h2>
       <%= ob.processingButton({
-        className: 'btn clrP clrBr clrSh2 modalContentCornerBtn js-save',
+        className: 'btn clrP clrBAttGrad clrBrDec1 clrTOnEmph modalContentCornerBtn js-save',
         btnText: ob.polyT('settings.btnSave')
       }) %>
     </div>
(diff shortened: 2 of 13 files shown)


morebrownies:
@jjeffryes I made a few small changes. this is ready whenever you get a chance.

js/templates/modals/settings/moderation.html
```diff
@@ -164,4 +164,10 @@ <h2 class="h3 clrT">Moderation</h2>
       </div>
     </div>
   </div>
+
+  <div class="contentBox padMd clrP clrBr clrSh3">
```
jjeffryes: If the other save button uses the processing template, this one should too to stay consistent. 

You should be able to just copy and paste the other one in the template.

rmisio:
> As part of this PR, could you update the buttons to use the processing template?

If the button wasn't using the processing template before, it means the js code to update the processing state isn't in place. Wouldn't it be better to make a separate PR to wire in processing buttons and their functionality where needed, rather than just put in impotent processing markup?

jjeffryes:
I thought I saw in some places save buttons were getting processing classes that weren't doing anything, so the wiring was there, but the template wasn't. 

We could make that separate though.

js/templates/modals/orderDetail/orderDetail.html
```diff
@@ -4,11 +4,10 @@
     // and componentize the css.
   %>
   <% if (ob.returnText) { %>
-  <div class="contentBox clrP clrSh3 clrBr clrT">
-    <a class="padSm gutterHSm overflowAuto margRSm flexVCent returnBox js-returnBox">
-      <span class="clrBr2 clrSh1 disc avatar flexNoShrink" style="<%= ob.getAvatarBgImage(ob.ownProfile.avatarHashes) %>"></span>
-      <span class="txUnl tx3 clamp clrT name"><%= ob.ownProfile.name %></span>
-      <span class="txU flexNoShrink tx6"><%= ob.returnText %></span>
+  <div class="btnStrip clrSh3">
```
rmisio: Since the `returnBox` styling is no longer needed, can you remove it from `_orderDetail.scss`?

Also, with what was taken out above, this template no longer needs `ownProfile`, can you remove the following last line from `OrderDetail.js`?

```
  render() {
    loadTemplate('modals/orderDetail/orderDetail.html', t => {
      const state = this.getState();

      this.$el.html(t({
        ...state,
        ...this.model.toJSON(),
        ownProfile: app.profile.toJSON(), // remove this line since ownProfile is no longer needed in the template
```

js/templates/modals/settings/moderation.html
rmisio: In this place I concur with @jjeffryes, if you use the `processingButton` template for this button, then it will get a spinner while a save is in progress. The JS for that is already wired in, so if you just replace:

`<a class="btn clrP clrBAttGrad clrBrDec1 clrTOnEmph js-save"><%= ob.polyT('settings.btnSave') %></a>`

with:

```
      <%= ob.processingButton({
        className: 'btn clrP clrBAttGrad clrBrDec1 clrTOnEmph modalContentCornerBtn js-save',
        btnText: ob.polyT('settings.btnSave')
      }) %>
```

it should work. If it's necessary, you are free to adjust the classes you send into that button via the `className` variable as long as you leave the `js-save` class.

jjeffryes:
Since it was just two places, I pushed the change to update to the processing buttons.

morebrownies:
Perfect, thank you

rmisio:
@morebrownies I just found a couple things in my review. Should be quick fixes.

morebrownies:
Ok great thank you. I'll circle back to this when I can.

js/templates/modals/settings/moderation.html
morebrownies: It looks like @jjeffryes beat me to this

js/templates/modals/orderDetail/orderDetail.html
morebrownies: Alright this has been removed

morebrownies:
This should be ready for review again @rmisio @jjeffryes

rmisio:
@jjeffryes this lgtm.

PR #645, opened 2017-08-10
Native notifications for chat

### LINKED ISSUE #628: Fire off native notification when chat message comes in while the app is not active
If the OB app is not focused and a chat message comes in, fire off a native notification (as is being done for other notifications). Consider updating the unread notification count in the OS app bar badge.

rmisio:
closes #628 

This PR will fire off a native notification when you receive a chat message as long as you are not actively chatting with the person the message is from. It will also include the unread message count with the unread notification count in the count that is set on the app bar icon.

This PR builds on #627 and its diff-report will look much leaner once the latter is merged.


rmisio:
something weird is going on, will re-open.

PR #650, opened 2017-08-10
Added many new tooltips, cleaned up copy and text styling

morebrownies:
**CHANGES**

- Added tooltips to `delete icons` in Add / Edit listing screen (within shipping option, variants, coupons)
- Added tooltip to `close icons` within the modal overlays
- Added tooltip to the `Grid/List button` on the store page
- Added tooltip to the `Bitcoin price ticker` in the wallet and about OpenBazaar section
- Made icon buttons a bit smaller globally
- Made text darker within the order modal
- Changed notification text from `Your payment has been received` to `Your payment has been delivered`
- Hovering an avatar within listing search results now displays tooltip for store name and location
- Improved the colors within the chat panel
- Added a href to the user avatar / name in top of transaction section to link back to your page

**PROBLEMS**

- Tooltips are not working correctly in some areas. The delete/close icon no longer display when i add the `toolTipNoWrap` class. (see below)
- The grid/list view toggle button on the store page acts strange when the `toolTipNoWrap` class is added too. (see below)

@jjeffryes I'm going to need help on those issues ^ 

screenshot 2017-08-10 11 41 27

screenshot 2017-08-10 11 41 06

screenshot 2017-08-10 11 40 33


js/templates/modals/editListing/variant.html
```diff
@@ -31,6 +31,6 @@
         <div class="tagsDropdown js-dropDownContainer"></div>
       </div>
     </div>
-    <a class="iconBtn ion-trash-b clrBr clrP margLSm js-btnRemoveVariant btnRemoveVariant"></a>
+    <a class="iconBtn ion-trash-b clrBr clrP clrSh2 margLSm toolTipNoWrap js-btnRemoveVariant btnRemoveVariant" data-tip="<%= ob.polyT('editListing.variants.toolTip.delete') %>"></a>
```
jjeffryes: Same thing here with the missing icon.

js/templates/userPage/store.html
```diff
@@ -7,8 +7,8 @@
     </div>
   </div>
   <div>
-    <div class="btnStrip clrT2 clrSh2">
-      <a class="iconBtn clrP clrBr js-toggleListGridView"><i class="ion-android-apps gridViewIcon"></i><i class="ion-navicon listViewIcon"></i></i></a>
+    <div class="clrT2 clrSh2">
```
jjeffryes: I fixed the missing tooltip here by taking off the btnStrip class. It wasn't needed now that there's only one button, and it had an overflow: hidden on it.

If we want to put tooltips on any other elements that need to keep the btnStrip styling, I can go back and see if it really needs the overflow: hidden, that may be there  to compensate for something we're not doing any more.

js/views/ListingCard.js
```diff
@@ -166,7 +166,9 @@ export default class extends baseVw {
             model: this.fullListing,
             profile: this.options.profile,
             vendor: this.options.vendor,
-            closeButtonClass: 'cornerTR ion-ios-close-empty iconBtn clrP clrBr clrSh3',
+            closeButtonClass: 'cornerTR iconBtn clrP clrBr clrSh3 toolTipNoWrap',
+            innerButtonClass: 'ion-ios-close-empty',
+            closeButtonTip: app.polyglot.t('pageNav.toolTip.close'),
```
jjeffryes: I fixed the issues with the listing edit close button by moving the icon to an inner element, and separating the class for it. The tooltip text is also passed in here, in case a future modal needs different text.

morebrownies:
thank you @jjeffryes very much appreciated!

js/templates/listingCard.html
```diff
@@ -42,7 +42,10 @@
   </div>
   <% if (ob.vendor) { %>
   <a class="userIconWrapper js-userName">
-    <div class="userIcon disc clrBr2 clrSh1"
+    <% const vendorTip = `${ob.vendor.name}${ob.vendor.location ? `  📍  ${ob.vendor.location}` : ''}`;
```
rmisio: @jjeffryes this emoji needs to be emojified.

js/templates/miniProfile.html
```diff
@@ -1,9 +1,12 @@
 <div class="flexVCent">
   <div>
-    <div class="userIcon disc clrBr2 clrSh1" style="<%= ob.getAvatarBgImage(ob.avatarHashes || {}) %>"></div>
+    <% const userLink = ob.peerIDx ? `href="#${ob.peerID}"` : ''; %>
```
rmisio: @jjeffryes Does the template really give you a variable named `peerIDx`? Is that coming from the server? I'm not seeing where that is coming from.

js/templates/listingCard.html
jjeffryes: Good catch. I was just thinking about fixing up the HTML and forgot that.

js/templates/miniProfile.html
jjeffryes: No, that was debugging code to see what happened when it failed, I forgot to take it out.

js/views/ListingCard.js
rmisio: This is not needed since the baseModal defaults it to same thing.

js/templates/listingCard.html
jjeffryes: Since this is in a data attribute used for the content of a tooltip, we can't emojify it. As far as I can tell, it looks identical to the emojified version on Windows, we could probably leave this as-is, unless there's a big difference on other OSes. 

rmisio: > As far as I can tell, it looks identical to the emojified version on Windows, we could probably leave this as-is, unless there's a big difference on other OSes.

@jjeffryes Keep in mind this also dependent on the version of an OS. If you go back far enough, those user will see jibberish (probably an empty rectangle).

So, @morebrownies, since we can't emojify (convert to an image) emojis in a tooltip, there are a couple options.
1.) Don't use emojis at all in tooltips.
2.) Use emoji's, but don't emojify them. The risk with this is that the emojis will look different across different OSs or look completely broken on non-supporting OSs.

jjeffryes: The patch for this PR is #665, I've removed the emoji there, @morebrownies chime in if you think it's important enough to put back in despite the potential issues.

morebrownies: I think we should remove the location from displaying in the tooltip if that is the case.

PR #651, opened 2017-08-10
Patch for Auto Update

jjeffryes:
Some adjustments to the auto update.
- win64 will be detected and added to the path for the updater, previously it was always win32.
- a conditional was removed that only set the autoUpater path for Mac.
- the check for updates option was removed from the help right click menu, if it is added back in later it should be conditional on the app being installed.
- some missing package.json data was added.


jjeffryes:

js/start.js
```diff
@@ -622,6 +623,20 @@ $(window).on('beforeunload', () => {
...
 ipcRenderer.on('show-server-log', () => launchDebugLogModal());
 
+// Handle update events from main.js
+ipcRenderer.on('updateChecking', () =>
+  showUpdateStatus(app.polyglot.t('update.checking'), '', 'pending'));
+ipcRenderer.on('updateAvailable', () =>
+  showUpdateStatus(app.polyglot.t('update.available'), '', 'pending'));
+ipcRenderer.on('updateNotAvailable', () =>
+  showUpdateStatus(app.polyglot.t('update.notAvailable')));
+ipcRenderer.on('error', (e, msg) =>
```
rmisio: Could we make this `updateError`? `error` seems a bit too generic here.

js/utils/autoUpdate.js
```diff
@@ -0,0 +1,60 @@
+import $ from 'jquery';
+import { ipcRenderer } from 'electron';
+import app from '../app';
+import Dialog from '../views/modals/Dialog';
+
+let statusMsg;
+let removeStatusMsgTimout;
+
+export function showUpdateStatus(status = '', msg = '', type = 'message') {
```
rmisio: Is it really necessary to have both the `status` and `msg` variables? It seems `msg` is never being passed in and status is always being passed in as a sentence (with a period), so it would never make sense to concatenate the two with a colon.

js/views/modals/about/About.js
```diff
@@ -73,7 +74,7 @@ export default class extends BaseModal {
 
   checkForUpdateClick() {
     console.log('checking for an update');
```
rmisio: Do we still need this console log?

main.js
```diff
@@ -476,60 +479,86 @@ function createWindow() {
...
+  ipcMain.on('checkForUpdate', () => {
+    autoUpdater.checkForUpdates();
+  });
 
-    autoUpdater.setFeedURL(feedURL);
+  autoUpdater.setFeedURL(feedURL);
 
+  mainWindow.webContents.on('dom-ready', () => {
     // Check for updates every hour
     autoUpdater.checkForUpdates();
```
rmisio: Could we make a separate checkForUpdates function which would reset the interval? Otherwise, this is checking every hour irrespective of any manual checks from the client. So, if the user manually checks 1 minute before the interval is due up, then there would needlessly be a second check a minute after a previous one.

Something like this should work:

```
let checkForUpdatesInterval;
const checkForUpdates = () => {
  clearInterval(checkForUpdatesInterval);
  autoUpdater.checkForUpdates();
  checkForUpdatesInterval = setInterval(() => {
    autoUpdater.checkForUpdates();
  }, 60 * 60 * 1000);
}
```

then anywhere you are calling `autoUpdater.checkForUpdates();`, call `checkForUpdates()` instead.

js/utils/autoUpdate.js
jjeffryes: Earlier code passed in the msg variables from various updater events, I took that out and forgot to clean up here. Good catch.

js/views/modals/about/About.js
jjeffryes: No, these will all come out, they were in for testing yesterday.

main.js
jjeffryes: Good call.

PR #665, opened 2017-08-14
Fix style fixes

jjeffryes:
- fixes a bad template variable
- removes a symbol in the tooltip that can't be emojified, and may not display right on some OSes
- fixed a spelling mistake
- took out some useless variables

### DIFF
--- a/js/templates/listingCard.html
+++ b/js/templates/listingCard.html
@@ -42,7 +42,7 @@
   </div>
   <% if (ob.vendor) { %>
   <a class="userIconWrapper js-userName">
-    <% const vendorTip = `${ob.vendor.name}${ob.vendor.location ? `  📍  ${ob.vendor.location}` : ''}`;
+    <% const vendorTip = `${ob.vendor.name}${ob.vendor.location ? ` ${ob.vendor.location}` : ''}`;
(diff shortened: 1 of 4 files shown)


js/templates/listingCard.html
```diff
@@ -42,7 +42,7 @@
   </div>
   <% if (ob.vendor) { %>
   <a class="userIconWrapper js-userName">
-    <% const vendorTip = `${ob.vendor.name}${ob.vendor.location ? `  📍  ${ob.vendor.location}` : ''}`;
+    <% const vendorTip = `${ob.vendor.name}${ob.vendor.location ? ` ${ob.vendor.location}` : ''}`;
```
rmisio: > I think we should remove the location from displaying in the tooltip if that is the case.

Unless I'm misinterpreting, @morebrownies said to remove the location entirely if we can't show the emoji.

jjeffryes: I think the location should stay, I think we were only talking about removing the icon. @morebrownies which one were you saying?

morebrownies: I suggested removing location altogether because it won't be clear what that value represents without some type of label.

jjeffryes: Huh. Ok.

PR #675, opened 2017-08-17
Clean up emojis

jjeffryes:
Cleans up a few templates, based off of #665, when that is merged this will be very small.

- emojifies some icons in the moderator detail and usercard templates
- hides the non-functional reputation in the moderator detail template
- makes the user icon link work in the moderator detail template

### DIFF
--- a/js/templates/modals/moderatorDetails.html
+++ b/js/templates/modals/moderatorDetails.html
@@ -4,7 +4,9 @@
     <div class="contentBox mDetailBox padMd clrP clrBr clrSh3 tx5">
       <div class="flex gutterH row">
         <div>
-          <a class="userIcon disc clrBr2 clrSh1" style="<%= ob.getAvatarBgImage(ob.avatarHashes) %>"></a>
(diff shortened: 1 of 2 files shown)


js/templates/modals/moderatorDetails.html
```diff
@@ -31,14 +33,14 @@
...
           <div class="tx5b clrT2"><%= ob.polyT('moderatorDetails.location') %></div>
         </div>
       </div>
-      <div class="rowDivV clrBr"></div>
-      <div class="flexExpand">
+      <div class="rowDivV clrBr TODO"></div>
+      <div class="flexExpand TODO">
         <div class="flexCol flexCent">
-          👍 XX<% // placeholder for reputation %>
+          <%= ob.parseEmojis('👍') %> XX<% // placeholder for reputation %>
```
jjeffryes: This is hidden with the todo class, but I added the parse emoji anyways so it's not forgotten in the future.

PR #676, opened 2017-08-17
Follow refactor

### LINKED ISSUE #200: Home: Follower count incorrect

rmisio:
This PR refactors the following / follower functionality as follows (see what I did there?):
- puts in faux pagination, so profiles aren't fetch for an entire list of followers / followings
- fetches the profile using async and usecache flags so the request aren't jamming up the xhr request queue
- since other nodes following / followers list are cached, the follow data of our own node will be used to adjust the list as appropriate (at least as it pertains to our own node showing up or not on a list)

closes #200


js/views/UserCard.js
```diff
@@ -73,19 +72,16 @@ export default class extends BaseVw {
...
     if (guid === app.profile.id) {
-      // don't fetch our this user's own profile, since we have it already
-      this.profileFetch = $.Deferred().resolve();
-      profile = app.profile;
+      // don't fetch this user's own profile, since we have it already
+      this.profileFetch = $.Deferred().resolve(app.profile);
     } else {
-      profile = new Profile({ peerID: guid });
-      this.profileFetch = profile.fetch();
+      this.profileFetch = getCachedProfiles([guid])[0];
```
rmisio: I decided to use the cached profiles after all. The latency improvement is substantial. And... even having the freshest moderator info doesn't really solve the issue. After a mod is added, whether it's 1 minute, 1 hour or 6 months later, that mod can change their terms and the users using that mod would have no idea. Perhaps it's as simple as sending a notification to users of a mod if the mod changes their terms? Seems like a bigger UX discussion we could circle back to.

jjeffryes:
The connected peers page looks a little weird now.

image

jjeffryes:
I made a quick patch to apply the same changes to the connected peers page. #688

jjeffryes:
Edit: deleted my comment about the following, I got tripped up by Mike following me back within seconds of me following him.

js/views/userPage/Follow.js
```diff
@@ -1,120 +1,355 @@
...
+    if (view) {
+      view.remove();
+      this.userCardViews.splice(this.userCardViews.indexOf(view), 1);
+      delete this.indexedUserCardViews[peerId];
     }
   }
 
-  className() {
-    return 'userPageFollow flexRow';
+  renderUsers(models = [], insertionType = 'append') {
```
rmisio: It doesn't hurt. If they passed in null, it would default it.

js/views/userPage/UserPage.js
```diff
@@ -47,6 +61,18 @@ export default class extends baseVw {
...
+    this.listenTo(app.ownFollowing, 'add', this.onOwnFollowingAdd);
+    this.listenTo(app.ownFollowing, 'remove', this.onOwnFollowingRemove);
+
+    this.followsYou = false;
+    followsYou(this.model.id).done(data => {
+      if (this.miniProfile) {
+        this.miniProfile.setState({ followsYou: data.followsMe });
+      }
+
+      if (this.followerCount === 0) this.followerCount += 1;
```
rmisio: I'll fix that. It should be if your on someone else's user page and they follow you, but their cached following count is 0, it should be 1 (because we know they're following at least one person).

PR #679, opened 2017-08-18
Fix naming GUID -> peer ID

JustinDrake:


js/templates/userPage/home.html
```diff
@@ -34,7 +34,7 @@
       </div>
       <div class="listItem">
         <div class="flex">
-          <span class="flexExpand">GUID</span>
+          <span class="flexExpand">Peer ID</span>
```
jjeffryes: Can you make this a translatable phrase?

jjeffryes: The phrase should go in the root level of the language file, since we'd use it anywhere the phrase "peer ID" appears.

Issue #581, opened 2017-07-24
Tweak Connected Peers


jjeffryes:
The Connected Peers screen should have the following changes applied:

Move the link to Settings/Advanced from the page navigation menu.
Show an error with advice on what to do if no peers are returned.

PR #688, opened 2017-08-18
Apply the same changes to the connected peers page.

jjeffryes:
Just a tweak so the connected peers page gets the same changes as the follow tabs. Otherwise the layout breaks.

### DIFF
--- a/js/templates/connectedPeersPage.html
+++ b/js/templates/connectedPeersPage.html
@@ -5,7 +5,9 @@
       <div class="tx4 border clrBr clrP pad"><%= ob.polyT('connectedPeersPage.totalPeers', { totalPeers: ob.peers.length }) %></div>
     </div>
   </div>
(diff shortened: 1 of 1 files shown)


rmisio:
Hmmm.... I didn't realize the connected peers page uses the `userPage` styling, which is a bit confusing. I think we probably did it this way because we thought this was a temporary page. If this does end up being a more permanent page, which should probably break out the common css into some common module (`userCards` or something) - but we could circle back to that.

jjeffryes:
> Hmmm.... I didn't realize the connected peers page uses the userPage styling, which is a bit confusing. I think we probably did it this way because we thought this was a temporary page. If this does end up being a more permanent page, we should probably break out the common css into some common module (userCards or something) - but we could circle back to that.

It was definitely considered a temporary thing for testing, there's an issue to make it nicer, #581 eventually. When we do that it should get it's own styles or the styles should be made common.

PR #691, opened 2017-08-18
Status Bar Under Chat

### LINKED ISSUE #649: The status bar is covering the chat window
This prevents me from being able to chat when the status bar is visible 

screenshot 2017-08-10 11 22 19

jjeffryes:
Moves the status bar in index.html so it is under the chat (they have the same z-index).

Closes #649

### DIFF
--- a/index.html
+++ b/index.html
@@ -15,11 +15,11 @@
       <section id="contentFrame" class="clrBr">
         <div id="pageContainer"></div>
       </section>
-
+      <section id="statusBar"></section>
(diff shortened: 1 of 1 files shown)


rmisio:
I was actually going to make it so the width of the status bar is shorter so it only bumps up against chat. That would account for the case of a status message being really long…. but that would be quite the edge case. Although, it would make things more future proof in case we ever allow a narrower app.

This is probably fine for now. I’ll merge.

PR #692, opened 2017-08-18
Misc fixes

jjeffryes:
Adds space between coupon applied text and remove link.
Closes #669 

Make everything in the pageNav bar disabled when not navigable
Closes #663

Take unneeded flex gutterH classes off of divs that only have text in them in the order details summary, so they don't overflow their containers.
Closes #661

If usecache or async in the options is undefined, make it false
Closes #647

Add left align to variants on listing details. 
Closes #640

Change modal gutters to 12px. 
Closes #518

PR #695, opened 2017-08-19
Coupon after Variants in Sidebar

jjeffryes:
Calculates the total in the purchase sidebar by applying the coupons after the variants.

Also moves the variants in the display to before the coupon.

### DIFF
--- a/js/templates/modals/purchase/coupons.html
+++ b/js/templates/modals/purchase/coupons.html
@@ -6,7 +6,7 @@
   </div>
 <% } %>
 <% ob.couponCodes.forEach(code => { %>
-  <div class="txSm rowTn flex">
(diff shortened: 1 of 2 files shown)


js/templates/modals/purchase/coupons.html
```diff
@@ -6,7 +6,7 @@
   </div>
 <% } %>
 <% ob.couponCodes.forEach(code => { %>
-  <div class="txSm rowTn flex">
+  <div class="txSm rowTn flexVCent gutterH">
```
jjeffryes: Ignore this, it got added to my local master from #692 by accident.

jjeffryes:
This is now part of #706

PR #705, opened 2017-08-22
Update address validation for bech32

cpacia:
Check bech32 address encoding in addition to base58 when validating bitcoin
addresses.

### DIFF
--- a/js/utils/index.js
+++ b/js/utils/index.js
@@ -7,6 +7,7 @@ import app from '../app';
 import multihashes from 'multihashes';
 import bitcoreLib from 'bitcore-lib';
 import twemoji from 'twemoji';
+import bech32 from 'bech32';
 
 export function getGuid(handle, resolver) {
   const deferred = $.Deferred();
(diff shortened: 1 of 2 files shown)


js/utils/index.js
```diff
@@ -142,7 +143,12 @@ export function isValidBitcoinAddress(address) {
     bitcoreLib.encoding.Base58Check.decode(address);
     return true;
   } catch (exc) {
-    return false;
```
rmisio: Can we simplify this and consolidate both checks into a single try / catch?

```
export function isValidBitcoinAddress(address) {
  if (typeof address !== 'string') {
    throw new Error('Please provide a string.');
  }

  try {
    bitcoreLib.encoding.Base58Check.decode(address);
    bech32.decode(address);
    return true;
  } catch (exc) {
    return false;
  }
}
```

cpacia: What you have there will always throw and exception since if something is base58 encoded it can't also be bech32 encoded and vice versa.

rmisio: ah, ok.

@jjeffryes This LGTM. If it's ok with you, feel free to merge.

jjeffryes:
Per @cpacia we'll need to adjust the QR code settings, he had to change the type to 8 to get it to work, which caused some layout issues.

image

We should patch that before merging this PR.

rmisio:
> Per @cpacia we'll need to adjust the QR code settings, he had to change the type to 8 to get it to work, which caused some layout issues.

@jjeffryes ok, i'll look into that.

rmisio:
closed in liu of #709

PR #706, opened 2017-08-22
Prevent Rounding Error in Receipt

### LINKED ISSUE #660: Wrong sum of rounded items on Order Summary
Total order price should sum all rounded items (in this case should be 107.99).

Also fix the typo "Plese".

round

jjeffryes:
When fiat prices are converted to BTC, then converted to a different fiat price, it can result in a rounding error of 0.01 in the displayed total vs. the individual prices of the listing, options, coupons and shipping. This PR fixes that with the following changes:
- show the listing price as the listing price plus the surcharge from the options chosen
- don't show the option prices
- subtract coupons before adding the shipping price
- convert and round all prices to 2 decimal places before doing any calculations
- if the quantity is more than 1, show a subtotal with the price for 1 listing after calculations
- show the total with quantity for the multiple of the subtotal by the quantity

This PR also refactors the purchases view to move the creation of child views out of the render function, so refreshes are handled correctly.

Closes #660

### DIFF
--- a/js/languages/en-US.json
+++ b/js/languages/en-US.json
@@ -1059,11 +1059,12 @@
     },
     "receipt": {
       "summary": "Order Summary",
-      "listing": "Listing (%{quantity})",
+      "listing": "Listing",
       "coupon": "Coupon",
       "shipping": "Shipping",
       "options": "Options",
-      "total": "Total"
+      "subtotal": "Subtotal",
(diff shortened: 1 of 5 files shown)


jjeffryes:
Closed because I forgot to change the option to be combined with the listing price.

PR #707, opened 2017-08-22
The useCache parameter was missing in the purchase view.

jjeffryes:
Just adds the paramenter. Good catch, @cpacia

### DIFF
--- a/js/views/modals/purchase/Moderators.js
+++ b/js/views/modals/purchase/Moderators.js
@@ -22,6 +22,7 @@ export default class extends baseVw {
       // set defaults
       apiPath: 'fetchprofiles',
       async: true,
+      useCache: true,
       moderatorIDs: [],
       method: 'POST',
       include: '',
@@ -60,7 +61,7 @@ export default class extends baseVw {
     const op = this.options;
     const includeString = op.include ? `&include=${op.include}` : '';
     const urlString =
-      `ob/${op.apiPath}?async=${!!op.async}${includeString}&usecache=${op.useCache || true}`;
+      `ob/${op.apiPath}?async=${op.async}${includeString}&usecache=${op.useCache}`;
     const url = app.getServerUrl(urlString);
 
     this.notFetchedYet = IDs;
--- a/js/views/modals/purchase/Moderators.js
+++ b/js/views/modals/purchase/Moderators.js
@@ -60,7 +60,7 @@ export default class extends baseVw {
     const op = this.options;
     const includeString = op.include ? `&include=${op.include}` : '';
     const urlString =
-      `ob/${op.apiPath}?async=${!!op.async}${includeString}&usecache=${!!op.useCache}`;
+      `ob/${op.apiPath}?async=${!!op.async}${includeString}&usecache=${op.useCache || true}`;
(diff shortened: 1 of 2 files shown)


js/views/modals/purchase/Purchase.js
```diff
@@ -452,6 +452,7 @@ export default class extends BaseModal {
         singleSelect: true,
         selectFirst: true,
         radioStyle: true,
+        useCache: false,
```
rmisio: Wouldn't we want this to be true? Isn't the idea to have the mods load faster?

jjeffryes: We keep going back and forth.

It's pretty important to have the latest info available when making a decision on which moderator to choose.

On the other hand, that's how I used to have it, and I forget who, but someone was adamant faster was more important that accurate.

@hoffmabc you want to weigh in here?

rmisio: If we don't want to usecache, then that's what's in place now, because if you don't pass it in, it defaults to false.

hoffmabc: Sorry coming in late here. We are talking about waiting for mods to load in when purchasing? I really thinking waiting kind of sucks but picking bad moderators was a huge suck on 1.x. Getting latest info might actually be best here.

rmisio: well then:

1.) Do we need a larger discussion on this. @cpacia was under the impression that we would opt for cached mods to decrease latency.
2.) @jjeffryes If we are going to go with the uncached option, then how is this PR necessary as that is what's in place now? If you don't pass in a value for usecache to the moderators View, it is undefined and then cast into a boolean... so it would be sent as false.

jjeffryes: @cpacia is the one that suggested this PR, and using the non-cached mods. 

If we don't want to use them, then we just close this PR, and everything is good.

rmisio: not how I read it:

image

cpacia: I'd vote for using the `usecache` flag. Like 99% of the time it will show the most up to date information. Maybe 1% of the time it will be stale.

jjeffryes:
Ok, flipped around so moderators defaults to using cache, unless told otherwise, and does not get told otherwise in purchase.

js/views/modals/purchase/Moderators.js
```diff
@@ -60,7 +60,7 @@ export default class extends baseVw {
     const op = this.options;
     const includeString = op.include ? `&include=${op.include}` : '';
     const urlString =
-      `ob/${op.apiPath}?async=${!!op.async}${includeString}&usecache=${!!op.useCache}`;
```
rmisio: In this view, you declare defaults in the constructor. For consistency, should we follow suit with useCache?

```
const opts = {
      // set defaults
      apiPath: 'fetchprofiles',
      async: true,
      ...
}
```

jjeffryes: I forgot I did that. You're right, that's a better place to put it than checking in the url string.

PR #709, opened 2017-08-22
Bech32

rmisio:
This PR updates the Bitcoin address validation to handle segwit addresses. It also tweaks the UI of the Payment view to handle segwit addresses which are much longer than regular addresses.

This PR should be run in conjunction with this [server PR](https://github.com/OpenBazaar/openbazaar-go/pull/636).


TheButterZone:
Is this a step towards having OpenBazaar wallet give out bech32 addresses for payment, and the seed be imported as an Electrum SegWit wallet?

PR #711, opened 2017-08-23
fixing issue where renderjson was throwing an exception in the install…

### LINKED ISSUE #672: renderjson doesn't work in installed app
In the installed version of the client, the renderjson.js library fails with this error:

Uncaught ReferenceError: renderjson is not defined at renderjson.js:58

That line is:
(_module || {}).exports = renderjson = function () {

When the client is run from the command line, it works normally.

rmisio:
…er app

closes #672

### DIFF
--- a/js/lib/renderjson.js
+++ b/js/lib/renderjson.js
@@ -1,3 +1,10 @@
+// WARNING WARNING WARNING
+// The way this module exports itself (the "module.exports = ") line has been modified
+// from the original source, since the original way was not allowing this module to
+// be imported (at least using ES6 syntax). If you update this file to a new version,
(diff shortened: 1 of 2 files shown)


js/lib/renderjson.js
```diff
@@ -51,7 +51,7 @@
 //     .array.syntax  ("[", "]")
 
 var module;
-(module||{}).exports = renderjson = (function() {
+(module||{}).exports = window.renderjson = (function() {
```
jjeffryes: Should we add a comment somewhere that this has been modified so we don't overwrite it by updating the library?

jjeffryes:
I'm merging this, but we'll need to test it with an installed version when we do our next release.

PR #712, opened 2017-08-23
various minor tweaks to the auto update status messages

rmisio:
This PR:

- fixes issue where the type on update status events was not being updated unless you explicitly passed in a type (e.g. the "notAvailable" status message would show a pending status)
- eliminates the `pending` status from the status bar message since the icon looked really odd. Adjusted the auto update 'checking' status message to have a trailing ellipse, which is how we handle similar pending messages across the site (e.g. 'Saving settings...', 'Publishing to the network...', etc...)


jjeffryes:
This looks fine to me. It will need to be tested with the next release using an installed app.

PR #714, opened 2017-08-23
making pop-up messages not dismissible, by default

### LINKED ISSUE #674: Closing the "1 new transaction" alert within the Wallet makes it very difficult to see new transactions later on
I noticed in one of @SamPatt user testing videos that the user closed the "1 new transaction" alert, and then later didn't know how to see his new transactions.

screenshot 2017-08-16 11 33 59

This leads to a UX issue: it's very difficult to see new transactions after the alert is closed. I believe the only way to see them is by waiting for another transaction to come through, and then clicking `refresh` on the alert.

Here's 2 solutions on how we could fix this issue:

1) Remove the `close` action on the new transaction alert. The alert will purposely annoy users until they click `refresh`. 

2) Keep the new transaction alert exactly as it is (with the close button), but each time the Wallet is opened, automatically pull in new transactions. 

I'm thinking 1 would be easier. 

Thoughts?

rmisio:
closes #674

### DIFF
--- a/js/views/PopInMessage.js
+++ b/js/views/PopInMessage.js
@@ -8,7 +8,7 @@ export default class extends baseVw {
     super(options);
 
     const opts = {
-      dismissable: true,
+      dismissable: false,
       ...options,
     };
 


morebrownies:
👏

rmisio:
Hey, I'm sure if you look at the change-set of this PR, you could see what a big effort this was. An applause emoji is the least you can do ;)

hoffmabc:
Mind blown

morebrownies:
👏 👏

PR #718, opened 2017-08-24
adjusting the sequence of where the processing state in the payment m…

### LINKED ISSUE #619: Multiples charges for the same purchase
v. 2.0.6
This occurs on Purchases page (when user first close the purchase window and returns later).
It is possible for the user to click many times on the "Yes, Pay" button, and he gets charged on the wallet many times for the same purchase.

multiple chargtes

rmisio:
…odule is updated

closes #619 

@jjeffryes The only way I could think of this bug happening is that we were removing the spinner once the spend Xhr completed, but the module doesn't consider the payment to be complete (and fire off the 'walletPaymentComplete' event) until a payment socket is received with the funds. My best guess is that socket was delayed for whatever reason (maybe it's dependent on the blockchain or some blockchain related api and those are known to have latency quirks).

So... I updated it so the 'processing' state on the 'Yes, Pay' button won't clear until we have confirmed payment or the Spend call fails.

### DIFF
--- a/js/views/modals/purchase/Payment.js
+++ b/js/views/modals/purchase/Payment.js
@@ -51,6 +51,8 @@ export default class extends BaseVw {
           if (e.jsonData.notification.orderId === this.orderId) {
             const amount = integerToDecimal(e.jsonData.notification.fundingTotal, true);
             if (amount >= this.balanceRemaining) {
+              this.$confirmWalletConfirm.removeClass('processing');
+              this.$confirmWallet.addClass('hide');
               this.trigger('walletPaymentComplete', e.jsonData.notification);
             } else {
               // Ensure the resulting balance has a maximum of 8 decimal places with not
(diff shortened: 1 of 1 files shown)


js/views/modals/purchase/Payment.js
```diff
@@ -115,8 +117,6 @@ export default class extends BaseVw {
       .fail(jqXhr => {
         openSimpleMessage(app.polyglot.t('purchase.errors.paymentFailed'),
           jqXhr.responseJSON && jqXhr.responseJSON.reason || '');
-      })
-      .always(() => {
         if (this.isRemoved()) return;
```
jjeffryes: Shouldn't the if removed happen before the simple message is opened?

rmisio: I figured people would want to know if their payment attempt failed, even if they've closed the modal.

jjeffryes: Gotcha.

PR #719, opened 2017-08-25
Scrollable update dialog

jjeffryes:
This fixes the update dialog being too tall and badly formatted.
- changes the p to a div 
- adds a class with a max height and overflow for the inner message when the autoUpdate creates it.
- uses a pre element so the update text retains it's formatting

image


jjeffryes:
@rmisio this is ready with updates.

PR #720, opened 2017-08-26
allowing dialog buttons to have custom classes and customizing the bu…

rmisio:
…ttons on the auto update dialog

### DIFF
--- a/js/views/modals/Dialog.js
+++ b/js/views/modals/Dialog.js
@@ -45,6 +45,7 @@ export default class extends BaseModal {
       messageClass: '',
       buttons: [],
       removeOnClose: true,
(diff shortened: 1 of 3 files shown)


rmisio:
@jjeffryes IMHO, this could absolutely wait until next week to be reviewed and merged. Enjoy the weekend!

js/views/modals/Dialog.js
```diff
@@ -45,6 +45,7 @@ export default class extends BaseModal {
       messageClass: '',
       buttons: [],
       removeOnClose: true,
+      defaultBtnClass: 'btn clrP clrBr',
```
jjeffryes: Should this also have the button shadow class on it?

Issue #569, opened 2017-07-17
Search Enhancements


jjeffryes:
## User Story
As a user, when I am viewing the search page I want to be able to easily switch to other default search providers. I should be able to switch back to my default provider, and manually entered search providers should not be lost by switching. I should also be able to manually enter new providers here, and see the results.

### Provider Selector
- [x] There is a OB logo, with the text "OpenBazaar" and "Discover". It is just for display, and does not have interactivity (or it switches to the browse experience).
- [x] There is a series of square buttons for search providers.
- [x] The list should be composed of the default providers, and the current default search provider if it is not on that list. Each provider should have:
    - title
    - url to their logo
    - url for the search query
    - url for Tor search (optional)
- [x] If the client is in Tor mode and the provider has a Tor url, use the Tor url.
- [x] We will manually limit the number of providers to fit horizontally.
    Each one:
    - [x] Has the logo for the search provider if available.
	- [x] If the logo is not available or doesn't load, a default should be shown.
	- [x] The currently active provider should be marked with an underline.
    - [x] Clicking on a search provider will start a blank search with that provider.
- [x] There should be an add provider button. 
NOTE: the plus should stay a plus, 
    - [x] When clicked, a tool tip is shown with a text input for the new search provider.
    - [x] There is a cancel button that will close the tooltip.
    - [x] There is an add button that will apply this provider to the current search parameters. If it returns valid data, it will be added to the list of providers.
    - [x] If the logo url is returned, save it for that provider.
    - we will save the list of custom providers locally, and add it on the fly to the default providers.
    - We will limit the maximum number of search providers that can be added so they don't exceed the horizontal space.
    - [x] If the provider entered is not a valid URL, an error should be shown.	
- [x] There is a button for Transactions.
- [x] There is a button for My Page.
- [x] The currently selected provider only applies to the on-page search input, not the address bar.
- [x] There is a remove provider link for manually added providers only.

### Search Suggestions
- [x] There should be a row of suggestions under the search bar.
- [x] This may still need to be implemented yet on the search server.
    Each suggestion:
	- [x] When clicked, will search for that term (or is it a full url with parameters?)
- [ ] If more suggestions are provided than fit in the horizontal space, there should be a More link that will show the rest in a tool tip. (if the more link is problematic, it can be truncated instead)

### Report Buttons
- [x] The server will provide an optional report URL
- [x] If the URL is present, all results will have a report button.
[...]

PR #731, opened 2017-08-28
Add search selector

jjeffryes:
This adds the ability to manage search providers on the search page.
- Two providers are set as defaults, OB1 and Duo Search. More can be added by adding them to the default providers in the data folder.
- The defaults are loaded when the client is started. If they are somehow deleted or changed, this will restore them.
- If a user comes to the page from the address bar, their default provider will be used. If they come to the page from the discover button, their active provider will be used. The currently used provider will be used for searches from the search bar on the page.
- If the user comes to the page from the address bar, but the address bar has a url for a search provider in it, that provider will be selected if it exists. If it doesn't, it will be used, and the user will have an option to save that provider as a new provider.
- Only providers with non-tor addresses are shown when in clear mode. Only providers with Tor addresses are shown in Tor mode.
- Each provider has 4 urls, one for general search, one for listing search, and Tor versions of each both. We don't currently use general search, it searches for users and listings (the code exists to activate it once we decide how the user would choose to make a general search).
- If a search provider is selected while on the page, it is saved as the active provider, and will be used the next time the discover button is clicked.
- If the current provider is not the default provider, it can be set to the default by clicking a link next to it's name.
- The currently used provider can be deleted using a link next to its name, unless it is one of the default providers.
- A new provider can be added by clicking the add button and filling in the address.
- If the user is in tor mode, the new provider is treated as a tor provider, otherwise as a normal provider.
- When data is returned from providers, their information is updated and saved if it has changed.
- The inputs for managing a default search provider in settings have been removed.

This PR completes the first section of #569

Note: the Duo Search default provider is using demo urls, they will need to be updated before release one.


morebrownies:
Hey @jjeffryes is it possible to safely increase the # of results per page?I find myself having to hit `next` frequently with a limit of 12 per page. 

Could we safely increase the limit to 24 per page? Or are load times going to suffer?

morebrownies:
If easy to do, it would be helpful to add tooltips to the page control buttons: `Previous` and `Next`

screenshot 2017-08-28 16 05 47

morebrownies:
If this is easy, it would be nice to have at launch. If not, let me know and I'll file a GH issue. 

Inactive tabs should be faded out to more clearly display which tab is selected.

**Current**
screenshot 2017-08-28 16 13 44

**Ideal** (I set these to `opacity: .35;` in this screenshot)
screenshot 2017-08-28 16 13 34

jjeffryes:
Changing the number of results is fine, it's just a parameter. I don't think 24 would make any difference with speed.

Fading the inactive provider buttons should be simple to do.

morebrownies:
> Changing the number of results is fine, it's just a parameter. I don't think 24 would make any difference with speed.

> Fading the inactive provider buttons should be simple to do.

❤️ 

> Do we need tooltips on the next/prev buttons? The arrow symbols are pretty universal.

Since we have them in the address bar, I figured it would be helpful to have them here too.

jjeffryes:
> Since we have them in the address bar, I figured it would be helpful to have them here too.

I think those may need tooltips a little more, since not all apps have back and forward, but arrows next to a page number is pretty well known, don't you think?

I was noticing today the tooltips can get kind of annoying if you've seen them all before and they keep popping up everywhere, so I'm hesitant to go overboard with them.

morebrownies:
I personally never find the tooltips to be annoying, even the obvious ones. At the very minimum, they may be useful for people overseas that may not be used to our design patterns. Maybe I'm overthinking it though.

Anyone else have an opinion on this?

js/start.js
```diff
@@ -263,15 +265,19 @@ let ownFollowingFailed;
...
+let searchProvidersFetch;
+let searchProvidersFetchFailed;
 
 function fetchStartupData() {
   ownFollowingFetch = !ownFollowingFetch || ownFollowingFetch ?
     app.ownFollowing.fetch() : ownFollowingFetch;
   exchangeRatesFetch = exchangeRatesFetch || fetchExchangeRates();
   walletBalanceFetch = !walletBalanceFetch || walletBalanceFetch ?
     app.walletBalance.fetch() : walletBalanceFetch;
+  searchProvidersFetch = !searchProvidersFetch || searchProvidersFetch ?
```
rmisio: You're working off of an old master here. These lines have been updated. For example:

```
walletBalanceFetch = !walletBalanceFetch || walletBalanceFetch ?
```

has become:

```
walletBalanceFetch = !walletBalanceFetch || walletBalanceFetchFailed ?
```

and similarly:

```
searchProvidersFetch = !searchProvidersFetch || searchProvidersFetch ?
```

needs to be:

```
searchProvidersFetch = !searchProvidersFetch || searchProvidersFetchFailed ?
```

js/start.js
```diff
@@ -387,11 +397,30 @@ function start() {
...
       fetchStartupData().done(() => {
         app.pageNav.navigable = true;
         app.pageNav.setAppProfile();
         app.loadingModal.close();
+
+        // set the default search providers.
+        app.searchProviders.add(getDefaultSearchProviders());
+        // if active and default providers aren't set, set them now.
+        if (!app.searchProviders.activeProvider) {
+          app.searchProviders.setProvider(app.searchProviders.at(0), 'active');
```
rmisio: Instead of calling `setProvider`, would we want to use the setters here?

`app.searchProviders.activeProvider = app.searchProviders.at(0);`

rmisio:
> Anyone else have an opinion on this?

I do find tooltips to be annoying. In places where most people would reasonably know what to do, I'd vote to omit them.

jjeffryes:
@morebrownies @rmisio this is ready for review.

js/start.js
```diff
@@ -264,15 +266,19 @@ let ownFollowingFailed;
...
+let searchProvidersFetch;
+let searchProvidersFetchFailed;
 
 function fetchStartupData() {
   ownFollowingFetch = !ownFollowingFetch || ownFollowingFailed ?
     app.ownFollowing.fetch() : ownFollowingFetch;
   exchangeRatesFetch = exchangeRatesFetch || fetchExchangeRates();
   walletBalanceFetch = !walletBalanceFetch || walletBalanceFetchFailed ?
     app.walletBalance.fetch() : walletBalanceFetch;
+  searchProvidersFetch = !searchProvidersFetch || searchProvidersFetch ?
```
rmisio: This should be:

`searchProvidersFetch = !searchProvidersFetch || searchProvidersFetchFailed ?`

rmisio:
@morebrownies @jjeffryes 

image

I would think we would need some type of badge or indicator which is the default provider, since the underline is the active provider, but it's not easily understandable which is the default one (since they are two different things).

morebrownies:
The plus icon has the wrong cursor when hovering. It currently displays the pointer instead of the hand.

screenshot 2017-08-29 10 36 56

rmisio:
After clicking the `set as default` link, it's tooltip still remains:

image

morebrownies:
`.confirmBox` needs `clrSh1` added to it

**Without shadow**
screenshot 2017-08-29 10 39 16

**With shadow**
screenshot 2017-08-29 10 39 23

morebrownies:
We're also missing `clrSh2` in a number of spots as well:

The **main search input** and **submit button** should have `clrSh2`
screenshot 2017-08-29 10 46 24

The **filter boxes** should all have `clrSh2`
screenshot 2017-08-29 10 46 16

The **URL for search provider input** and **Add button** should have `clrSh2`
screenshot 2017-08-29 10 48 07

morebrownies:
> I would think we would need some type of badge or indicator which is the default provider, since the underline is the active provider, but it's not easily understandable which is the default one (since they are two different things).

How about a ⭐️ ?

screenshot 2017-08-29 10 51 38

morebrownies:
@jjeffryes I noticed the change from **12** to **24** listings only works when viewing the OB1 search results. Is this a bug? Or does DUO need to change something on their end in order for 24 listings to display?

js/data/defaultSearchProviders.js
```diff
@@ -0,0 +1,28 @@
+export function getDefaultSearchProviders() {
```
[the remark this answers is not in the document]
jjeffryes: Good call. When I started I had some other stuff happening in there, I took it out but didn't change this part.

js/data/defaultSearchProviders.js
```diff
@@ -0,0 +1,28 @@
...
+  const defaultSearchProviders = [
+    {
+      id: 'ob1',
+      name: 'OB1',
+      logo: '../imgs/ob1searchLogo.png',
+      search: 'https://search.ob1.io/search',
+      listings: 'https://search.ob1.io/search/listings',
+      torsearch: 'http://obsearchwhcqnhvj.onion/search',
```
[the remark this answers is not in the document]
jjeffryes: That's the real link.

jjeffryes:
> I noticed the change from 12 to 24 listings only works when viewing the OB1 search results. Is this a bug? Or does DUO need to change something on their end in order for 24 listings to display?

Duo turned off the pagesize parameter and hard coded it to 12. They were having an issue with the page size, I can't replicate it so I asked them to turn it back on.

jjeffryes:
This has been updated with the initial review comments.

The star on the default seems out of place to me, I haven't done anything with that yet. I'll do some experimenting with it, if any of us comes up with a great idea let's discuss.

js/models/search/SearchProvider.js
```diff
@@ -0,0 +1,60 @@
...
+    const urls = ['listings', 'torlistings'];
+    const validUrls = [];
+
+    urls.forEach((url) => {
+      if (attrs[url] && is.string(attrs[url]) && is.url(attrs[url])) {
+        validUrls.push(url);
+      }
+    });
+
+    if (!validUrls.length) {
```
rmisio: Why would you push a `listings` error if the torListing url is invalid? The error key should map directly to the field name.

```
    urls.forEach((url) => {
      if (attrs[url] && is.string(attrs[url]) && is.url(attrs[url])) {
         addError(url, app.polyglot.t(`searchProviderModelErrors.${url}`));
      }
    });
```

js/languages/en-US.json
```diff
@@ -1574,6 +1590,11 @@
     "totalPercentageTooLow": "The sum of the buyer and vendor amounts must add up to 100.",
     "provideResolution": "Please explain your decision."
   },
+  "searchProviderModelErrors": {
+    "name": "The name must be a string.",
```
rmisio: I don't think you want to use field names as the error keys since a field could have any number of potential errors. Those keys would be better as `provideName`, `invalidLogo`, `invalidListingUrl`. This would play nice if those fields had additional error, e.g. `nameTooLong`.

js/templates/search/AddProvider.html
```diff
@@ -0,0 +1,12 @@
+<div class="tx5 confirmBox arrowBoxCenteredTop padMd clrBr clrP clrSh1 clrT addBox js-addProvider <% if (ob.hideBox) print('hide') %>">
+  <div class="flexCol gutterV txLft">
+    <h3><%= ob.polyT('search.addTitle') %></h3>
+    <% const plhdr = ob.polyT(`${ob.usingTor ? 'search.addTorPlaceholder' : 'search.addPlaceholder'}`) %>
+    <% if (ob.errors['listings']) print(ob.formErrorTmpl({ errors: ob.errors['listings'] })) %>
```
rmisio: I think you would want to pass in `usingTor` into here and if it's true show the `torListings` error if it exists instead of the `listings` error.

rmisio:
> The star on the default seems out of place to me, I haven't done anything with that yet. I'll do some experimenting with it, if any of us comes up with a great idea let's discuss.

FWIW, I'm not sold on the confusing UX of "default" vs. "active" search providers, but the star looked ok to me. It's better than no default provider indicator.

js/start.js
```diff
@@ -39,7 +39,7 @@ import { fixLinuxZoomIssue, handleLinks } from './startup';
 import ConnectionManagement from './views/modals/connectionManagement/ConnectionManagement';
 import Onboarding from './views/modals/onboarding/Onboarding';
 import SearchProvidersCol from './collections/search/SearchProviders';
-import { getDefaultSearchProviders } from './data/defaultSearchProviders';
+import getDefaultSearchProviders from './data/defaultSearchProviders';
```
rmisio: `getDefaultSearchProviders` doesn't make sense as the name for this anymore since you're not getting anything. It's already a list. I'd go with `defaultSearchProviders`.

js/start.js
```diff
@@ -38,6 +38,8 @@ import listingDeleteHandler from './startup/listingDelete';
 import { fixLinuxZoomIssue, handleLinks } from './startup';
 import ConnectionManagement from './views/modals/connectionManagement/ConnectionManagement';
 import Onboarding from './views/modals/onboarding/Onboarding';
+import SearchProvidersCol from './collections/search/SearchProviders';
+import getDefaultSearchProviders from './data/defaultSearchProviders';
```
rmisio: `getDefaultSearchProviders` doesn't make sense as the name for this anymore since you're not getting anything. It's already a list. I'd go with `defaultSearchProviders`.

js/data/defaultSearchProviders.js
```diff
@@ -0,0 +1,26 @@
...
+  {
+    id: 'ob1',
+    name: 'OB1',
+    logo: '../imgs/ob1searchLogo.png',
+    search: 'https://search.ob1.io/search',
+    listings: 'https://search.ob1.io/search/listings',
+    torsearch: 'http://obsearchwhcqnhvj.onion/search',
+    torlistings: 'http://obsearchwhcqnhvj.onion/search/listings',
+    locked: true,
+    order: 1,
```
rmisio: Is this order attribute really necessary? If the only reason for it is to ensure the default ones are first, can we just splice them in at the beginning?

`app.searchProviders.add(getDefaultSearchProviders(), { at: 0 });`

jjeffryes: The idea here is to make sure no other process can accidentally break things by putting other models in front of them.

I'll go change the code that assumes the original default provider is at position 0, that's a fragile way of doing it.

js/models/search/SearchProvider.js
jjeffryes: The reason is it's an old copy and paste I never changed. Good catch!

js/templates/search/AddProvider.html
jjeffryes: Old copy and past again.

js/models/search/SearchProvider.js
jjeffryes: Ah, not a copy and paste. Listings is the name I was using for the listings search URL, which is confusing.

Showing a generic error if any of the urls are invalid isn't a great way to do this. I've rewritten it to show specific errors for each url, and then leave it up to the template to display the one that is relevant.

js/templates/search/AddProvider.html
jjeffryes: Not actually copy and paste, see earlier comments. This has been refactored to use separate url errors based on a passed in url type (since there are 4 possible urls currently).

js/models/search/SearchProvider.js
```diff
@@ -0,0 +1,55 @@
...
+    return LocalStorageSync.sync.apply(this, args);
+  }
+
+  validate(attrs, options) {
+    const errObj = {};
+    const addError = (fieldName, error) => {
+      errObj[fieldName] = errObj[fieldName] || [];
+      errObj[fieldName].push(error);
+    };
+    const urlTypes = options.urlTypes || ['search', 'listings', 'torsearch', 'torlistins'];
```
rmisio: There's a typo in the last item.

jjeffryes:
@rmisio typo fixed.

PR #732, opened 2017-08-28
503

### LINKED ISSUE #503: Sending payment from wallet needs to make network fee more clear
When I sent **$10** to an address it showed up as **$12.25** in my transaction list. I initially thought it was a bug, but then I realized the fee had been included. We probably need to be more clear about what is going on here instead of tacking on the fee behind the scenes.
 
I suggest we remove the "Double check that the address and amount are correct" text and replace it with: "**A $2.20 (0.0002 BTC) network fee will be included**" Cancel | Yes, Send

screenshot 2017-06-01 18 49 12

rmisio:
- Creates a centralized fees module which includes a function to get the estimated fee from the server. The function caches its results.
- Adds the estimated network fee to the confirm box when clicking Send to send money from the wallet (closes #503).
- Refactors the confirmWallet box from the Purchase Payment view:
    1. Gets the estimated fee from the newly created utility module.
    2. When the fee is not available (because it is in the process of being fetched or the fetch failed), the messaging will use the payment amount without the fee and include generic messaging instructing the user that a fee will be / needs to be added in.


jjeffryes:
Something is off in the math here. 

If I have 0.00274301 funds, and I try to send 0.0018, the estimated fee is 0.00078, for a total of 0.00258.

But I get a server message that the funds are insufficient. I had to lower the amount to 0.0015 (total of 0.00228) before it would send (0.0016 was too high), which would seem to make the fee around 0.00108.

image

js/views/modals/wallet/SendMoney.js
```diff
@@ -189,6 +201,12 @@ export default class extends baseVw {
       this._$sendConfirm = null;
 
       this.$('#walletSendCurrency').select2();
+
+      const fee = this.sendConfirmBox && this.sendConfirmBox.getState().fee;
```
jjeffryes: If sometimes this.sendConfirmBox exists, shouldn't the view be removed before creating a new one below?

js/languages/en-US.json
```diff
@@ -701,7 +701,8 @@
       "sendPaymentFailDialogTitle": "Unable to send payment",
       "sendConfirm": {
         "heading": "Are you sure?",
-        "body": "Double check that the address and amount are correct.",
+        "body": "A %{fiatBtcPairing} estimated network fee will be included",
```
jjeffryes: Maybe change this to "A network fee will be included, currently estimated at %{fiatBtcPairing}"

That way it's more clear it can change by the time you click send, and it makes it sound more like a guess (which it is).

js/languages/en-US.json
```diff
@@ -1062,15 +1063,17 @@
       "note4": "Once you have paid, it may take a bit for the interface to update.",
       "confirmWallet": {
         "title": "Are you sure?",
-        "payBTC": "Pay %{amountBTC} to the vendor",
-        "payFiat": "Pay %{amountFiat} (%{amountBTC}) to the vendor",
+        "payMsg": "Pay %{fiatBtcPairing}* to the vendor.",
+        "payMsgNotePlusFee": "*The network fee will be added on top of the listed amount",
+        "payMsgNoteFeeIncluded": "*Includes the estimated network fee of %{amount}",
```
jjeffryes: Same idea here, maybe "Includes a network fee, currently estimated at %{amount}"

js/languages/en-US.json
```diff
@@ -1062,15 +1063,17 @@
...
         "cancel": "Cancel",
         "confirm": "Yes, Pay",
         "fundsLow": "Your wallet doesn't have enough Bitcoin. Add funds to complete this purchase.",
         "insufficientFundsTitle": "Insufficient Funds",
         "insufficientFundsMsg1": "Your wallet currently holds %{funds}.",
         "insufficientFundsMsg2": "Add %{fundGap}* to your wallet, or pay with an external wallet to continue.",
-        "insufficientFundsNote": "*%{amount} plus an estimated fee of %{fee}"
+        "insufficientFundsNote": "*%{amount} plus an estimated fee of %{fee}",
```
jjeffryes: "*%{amount} plus fee, currently estimated at %{fee}"

rmisio:
@jjeffryes On our end, it's just an estimated fee, so if you're right at the edge with your available balance, I think there will always be potential mismatches between the client and server. There are a couple of things at play here:

1.) The server caches the estimated fee API for one minute. I don't know if it uses this cache when spending from the wallet, or if it always gets a fresh value. Either way, even if it does use the cache. The cache may have expired from what the UI fetched.

Also, the UI is caching the fee for 5 minutes. ! minute seemed a little too low, especially given that if the UI and server cache time line up, the still won't be totally in sync.

2.) BTC fee are per byte and the UI estimated fee does not take the actual transaction size into account. It just uses an average transaction size of 200 bytes. I would think the server is using the actual transaction size.

I'll go ahead and make your suggested text changes. Perhaps, they'll help at least a little?

I do want to stress that this only is an issue when the user is right at the edge of barely having enough money to cover the transaction and fee. But, I do agree, it's weird having the UI estimate one value, that you meet and then the server failing with `insufficient funds`.

PR #735, opened 2017-08-29
handling the payment finalized state in the order summary tab

rmisio:
To test, add the following line before the return in parse() of the Order model:
`response.state = 'PAYMENT_FINALIZED';`

If an order is in the finalized state, the following message will be added below the progress bar:

image

The message is geared towards the vendor because the server will only ever update the vendor's order with that state.

When in this state, the progress bar will default to the first position, even though it's possible the order progressed beyond that position. It would take a bit of work to determine what state they had been in before the finalized state and since we're minimally addressing the finalized state at this point, we'll defer on the progress bar being completely accurate.

### DIFF
--- a/js/templates/modals/orderDetail/summaryTab/summary.html
+++ b/js/templates/modals/orderDetail/summaryTab/summary.html
@@ -8,6 +8,7 @@
 </div>
 <hr class="clrBr rowLg" />
 <div class="js-statusProgressBarContainer statusProgressBarContainer"></div>
+<p class="js-paymentFinalizedMsg clrTErr hide" style="margin-top: 10px"><span class="ion-alert-circled padSm"></span><%= ob.polyT('orderDetail.summaryTab.vendorFinalizedPayment') %></p>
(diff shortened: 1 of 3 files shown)


js/templates/modals/orderDetail/summaryTab/summary.html
```diff
@@ -8,6 +8,7 @@ <h1 class="inline tx4"><%= ob.polyT('orderDetail.summaryTab.orderNumber', { orde
 </div>
 <hr class="clrBr rowLg" />
 <div class="js-statusProgressBarContainer statusProgressBarContainer"></div>
+<p class="js-paymentFinalizedMsg clrTErr hide" style="margin-top: 10px"><span class="ion-alert-circled padSm"></span><%= ob.polyT('orderDetail.summaryTab.vendorFinalizedPayment') %></p>
```
rmisio: Cut a corner and embedded some css, since this is a temporary message that we'll very likely be refactoring.

PR #742, opened 2017-08-31
Fix search pagination

jjeffryes:
This fixes 2 issues:
- the first number in the page range was incorrect when the end was smaller than the page size (ex: when showing 20 to 33, the first number was 14 instead of 20).
- when doing a new search, the old server page value was not being cleared, so if someone was on page 4 and made a new search, the new search would start on page 4.


js/models/wallet/Spend.js
```diff
@@ -118,7 +118,7 @@ export function spend(fields) {
   const attrs = {
     currency: app && app.settings && app.settings.get('localCurrency') || 'BTC',
     feeLevel: app &&
-    app.localSettings && app.localSettings.get('defaultTransactionFee') || 'PRIORITY',
+      app.localSettings && app.localSettings.get('defaultTransactionFee') || 'PRIORITY',
```
jjeffryes: I'm not sure why this updated. It was over the line character limit, but I don't think I have my IDE set to automatically fix that.

jjeffryes: I think this is an old fix it picked up for this line being over the 80 character limit.

PR #745, opened 2017-08-31
Update search

jjeffryes:
- remove the active provider and all related code
- if no default has been set, show a tooltip instead of search results, when the user selects a search provider it will be set as the default and the search query will fire

To test this, you can either clear your local storage, or do the following:

1. add search.ob1.io as a new provider
2. set it as your default
3. remove it
4. navigate away
5. navigate back to search

image


morebrownies:
Looks good to me 👍

js/templates/search/Providers.html
```diff
@@ -8,7 +8,14 @@
   </div>
 </div>
 <div class="providersBar flexExpand">
-  <div class="providerWrapper gutterH margR js-providerWrapper"></div>
+  <div class="providerWrapper gutterH <% if(!ob.selecting) print('margR') %> js-providerWrapper">
+    <% if (ob.selecting) { %>
+      <div class="selectingBox confirmBox arrowBoxTop clrP clrBr clrSh1">
+        <h2>Select a default search provider</h2>
```
rmisio: I think we need some translations here.

js/views/search/SearchProviders.js
```diff
@@ -83,7 +83,10 @@ export default class extends BaseView {
         const view = this.createProviderView(provider);
         if (view) {
           this.providerViews.push(view);
-          view.render().$el.appendTo(providerFrag);
+          if (provider.get(this.options.urlType)) {
```
rmisio: Could we do something with urlType in the constructor? Either default it or throw an exception if it's not provided? It's use multiple times in this view and, unless a dev digs through the code, they'd have no idea they should provide it.

rmisio:
@morebrownies are you fine with the text size of 10px in that tooltip? It seems a little small to me?

morebrownies:
> @morebrownies are you fine with the text size of 10px in that tooltip? It seems a little small to me?

Good call, would be worth bumping it up a few pixels.

jjeffryes:
@morebrownies @rmisio Good catches. This is updated.

morebrownies:
👍

PR #749, opened 2017-09-01
fixing issue where a view is not being removed before being recreated

rmisio:
@jjeffryes good catch on this one in #503. You merged before I had a chance to fix.


jjeffryes:
Oops. Sorry. When you merged the language stuff I thought it was ready.

PR #751, opened 2017-09-01
Adds a beta label

### LINKED ISSUE #744: Add Beta Label
While in beta, add a beta label:
- if in beta on mainnet, show a beta label where the testnet label is
- if in beta on testnet, show a beta testnet label where the testnet label is (ie: add the word beta to i)
- if out of beta on testnet show the normal testnet label

jjeffryes:
Adds s beta label where the test label is. The label is always visible, if the user is using testnet it will say "Beta Testnet", if the user is on mainnet it will just say "Beta".

Closes #744

### DIFF
--- a/js/templates/pageNav.html
+++ b/js/templates/pageNav.html
@@ -37,9 +37,9 @@
           </div>
 
           <div id="testnetFlag" class="btn barBtn normalBtn clrP clrBr" >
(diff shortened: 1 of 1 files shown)


rmisio:
The code looks fine to me. I'll leave the visual review to @morebrownies.

morebrownies:
Looks good 👍

PR #753, opened 2017-09-01
Added missing shadow to filters on transaction page. Also unbolded no…

morebrownies:
Added missing shadow to filters on transaction page. Also unbolded no results found text.


morebrownies:
@jjeffryes i resubmitted the PR

PR #756, opened 2017-09-02
Order patch

### LINKED ISSUE #658: Written memo is not being displayed on order details
v.2.0.6
w7 64 bits

It always displays as "n/a" regardless of what is written.
memo

### LINKED ISSUE #659: Email address should be displayed on order details
Email address given by buyer should be displayed on order details.

emailorder

### LINKED ISSUE #728: Email input on additional information has no validation
There is no validation on the format, it accepts text without @

jjeffryes:
This fixes some issues with order data.
- the memo is correctly saved when purchasing an item.
- the alternate contact data is shown in the order details.
- the Email field in the purchase is updated to "Email Address or Phone Number", since this is not an email field, the field is for any alternate contact method.

Closes #728 
Closes #659 
Closes #658 

image

image

### DIFF
--- a/js/templates/modals/orderDetail/summaryTab/orderDetails.html
+++ b/js/templates/modals/orderDetail/summaryTab/orderDetails.html
@@ -74,7 +74,7 @@
                   <b><%= ob.polyT('orderDetail.summaryTab.orderDetails.addressNotes') %></b>
                 </div>
                 <div>
-                  <i><%= ob.order.shipping.addressNotes %></i>
+                  <%= ob.order.shipping.addressNotes %>
                 </div>
               </div>
             <% } %>
--- a/js/templates/modals/orderDetail/summaryTab/orderDetails.html
+++ b/js/templates/modals/orderDetail/summaryTab/orderDetails.html
@@ -113,7 +113,7 @@
       <div class="flexRow gutterH">
         <div class="col6">
           <div class="gutterVTn">
-            <div class="txB">Memo</div>
+            <div class="txB"><%= ob.polyT('orderDetail.summaryTab.orderDetails.memo') %></div>
             <div><%= item.memo ? ob.parseEmojis(item.memo) : ob.polyT('orderDetail.summaryTab.notApplicable') %></div>
           </div>
         </div>
@@ -121,7 +121,7 @@
           <div>
             <div class="gutterVTn">
               <div class="txB"><%= ob.polyT('orderDetail.summaryTab.orderDetails.alternateContact') %></div>
-              <div><%= ob.order.alternateContactInfo %></div>
+              <div><%= ob.order.alternateContactInfo ? ob.order.alternateContactInfo : ob.polyT('orderDetail.summaryTab.notApplicable') %></div>
             </div>
           </div>
         </div>
--- a/js/templates/modals/orderDetail/summaryTab/orderDetails.html
+++ b/js/templates/modals/orderDetail/summaryTab/orderDetails.html
@@ -50,14 +50,17 @@
           <% if (ob.order.shipping && ob.order.shipping.country !== 'NA') { %>
             <div><%= ob.order.shipping.shipTo %></div>
             <% if (ob.order.shipping.address) { %>
-            <div><%= ob.order.shipping.address %></div>
+              <div><%= ob.order.shipping.address %></div>
             <% } %>
             <% if (addressLine3) { %>
-            <div><%= addressLine3 %></div>
+              <div><%= addressLine3 %></div>
             <% } %>
-            <div><%= addressLine4 %></div>
+              <div><%= addressLine4 %></div>
             <% if (ob.order.shipping.addressNotes) { %>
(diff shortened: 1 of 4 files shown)


rhcastilhos:
"Ship to" field shouldn't display additional information (e-mail and/or phone) together with shipping address. The first one is private information (only seller should be informed) and the other is public information (shipping carrier can know).

jjeffryes:
@rhcastilhos I don't understand your comment. The additional contact information is optional, and is added to the contract for the seller to use to contact the buyer.

It's only displayed to the buyer and seller.

rhcastilhos:
I agree that additional information should be displayed on order details, but not on the same field of the shipping address.

As a customer, when I see my telephone or email written together with shipping address, it feels that those informations will be transmited to shipping carrier. But I want only my shipping address to be transmitted to third parties (besides vendor).

jjeffryes:
All of the information is only shown to the vendor. If they wanted to give it to a shipping company they can, regardless of where we show it.

rhcastilhos:
Contact information shouldn't be displayed together with shipping address because:
1. Buyer may not want his e-mail/phone printed outside of his package (on a shipping address tag).
2. Buyer may want to send his order to some friend, but want to inform his own phone and not his phone

rhcastilhos:
If additional contact field can be used both for email and phone, I believe it should be split in 2 different fields (one for email and one for phone). This would facilitate exporting customer data to some database sheet.

jjeffryes:
This is updated to move the additional contact information to another area. It also no longer copies it with the address.

The field can be anything currently, we can't separate it because we don't know what data is in it. 

We can consider changing it to specific, validated fields for email, phone, or other contact methods in the future.

rhcastilhos:
Good. So I would suggest changing field title from "Email address or Phone Number" to "Additional Contact Information" and changing default value from "name@email.com" to "name@email.com or your phone number".

js/templates/modals/orderDetail/summaryTab/orderDetails.html
```diff
@@ -110,10 +110,23 @@ <h2 class="tx4 margRTn"><%= ob.polyT('orderDetail.summaryTab.orderDetails.headin
         </div>
       </div>
       <hr class="clrBr" />
-      <div class="gutterVTn">
-        <div class="txB">Memo</div>
-        <div><%= item.memo ? ob.parseEmojis(item.memo) : ob.polyT('orderDetail.summaryTab.notApplicable') %></div>
+      <div class="flexRow gutterH">
+        <div class="col6">
+          <div class="gutterVTn">
+            <div class="txB">Memo</div>
```
rmisio: Can we translate this? (I understand that was not an oversight on your part - I missed it)

js/templates/modals/orderDetail/summaryTab/orderDetails.html
```diff
@@ -110,10 +110,23 @@ <h2 class="tx4 margRTn"><%= ob.polyT('orderDetail.summaryTab.orderDetails.headin
...
+            <div><%= item.memo ? ob.parseEmojis(item.memo) : ob.polyT('orderDetail.summaryTab.notApplicable') %></div>
+          </div>
+        </div>
+        <div class="col6">
+          <div>
+            <div class="gutterVTn">
+              <div class="txB"><%= ob.polyT('orderDetail.summaryTab.orderDetails.alternateContact') %></div>
+              <div><%= ob.order.alternateContactInfo %></div>
```
rmisio: To follow suit with the other fields, can we show `ob.polyT('orderDetail.summaryTab.notApplicable')` if there is not alternate contact info set?

image

^^^ all empty fields show an n/a.

rmisio:
@morebrownies Does that address note look ok to you?

image

It all looks a bit too jammed together to me, but in case I'm overthinking, I defer to you.

jjeffryes:
Updated design for the address notes coming up.

jjeffryes:
image

rmisio:
image

A few thoughts:
1.) I'm not sure we want the colon after the notes, since none of the other labels have it.
2.) Are the italic really necessary here? Since the address notes are in their own section, they already are aesthetically separated from the address. Seems a bit overkill to me.
3.) Is it clear that these are address notes? If I didn't know they were, I would just think this was a section for generic notes. If it makes sense to others, I could totally get over this concern.

Aside for these points, this LGTM. I'll hold off on merging until @morebrownies can chime in.

@jjeffryes If it's urgent that this go in tonight, I suppose we could always merge and then put in a patch if Mike wants anything changed.

jjeffryes:
Good points. Updated.

image

morebrownies:
I like what you guys landed on. One small suggestion...

Let's make sure to be consistent with the `Delivery notes` label. In the new address form, we label it as `Shipping notes`. It should be changed to `Delivery notes`

screenshot 2017-09-06 09 26 44

jjeffryes:
Updated to Delivery Notes in the settings too.

PR #764, opened 2017-09-04
Don't show moderators that have the wrong currency

jjeffryes:
This will check the currency accepted by a mod, and not show moderators that use the wrong currency (BTC vs. tBTC) for the network (mainnet vs. testnet).

### DIFF
--- a/js/collections/Moderators.js
+++ b/js/collections/Moderators.js
@@ -61,8 +61,11 @@ export default class extends Collection {
         // don't add excluded ids or your own id
         const notExcluded = excludeList.indexOf(mod.id) === -1 && mod.id !== app.profile.id;
         // don't add if not a mod or the mod data is missing
+        // don't add if the currency doesn't match the network
+        const cur = !!app.serverConfig.testnet ? 'tBTC' : 'BTC';
+        const validCur = mod.get('moderatorInfo').get('acceptedCurrencies').indexOf(cur) > -1;
         if (!mod.isModerator) this.trigger('invalidMod', { id: mod.id });
-        return mod.isModerator && notExcluded;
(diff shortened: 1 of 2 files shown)


js/collections/Moderators.js
```diff
@@ -61,8 +61,11 @@ export default class extends Collection {
         // don't add excluded ids or your own id
         const notExcluded = excludeList.indexOf(mod.id) === -1 && mod.id !== app.profile.id;
         // don't add if not a mod or the mod data is missing
+        // don't add if the currency doesn't match the network
+        const cur = !!app.serverConfig.testnet ? 'tBTC' : 'BTC';
```
rmisio: Could you use `app.serverConfig.cryptoCurrency` to determine the currency as opposed to the testnet flag? That would make it more robust once we start handling other crypto currency values.

js/views/modals/purchase/Moderators.js
```diff
@@ -98,7 +98,16 @@ export default class extends baseVw {
                     // don't add profiles that are not moderators. The ID list may have peerIDs
                     // that are out of date, and are no longer moderators.
                     if (eventData.profile.moderator && eventData.profile.moderatorInfo) {
-                      this.moderatorsCol.add(eventData.profile);
+                      // if the moderator has an invalid currency, remove them from the list
+                      const buyerCur = !!app.serverConfig.testnet ? 'tBTC' : 'BTC';
```
rmisio: Same as above. Can we use `app.serverConfig.cryptoCurrency`?

FWIW, in my config, I'm seeing the `TBTC` as all caps. If you're seeing it sometimes come back as `tBTC`, we may need to uppercase it.

PR #766, opened 2017-09-04
Update electron to 1.7.5 [WIP]

jjeffryes:
Note: do not merge until this is thoroughly tested on all OSes.

- [ ] Tested on Win32
- [ ] Tested on Win64
- [ ] Tested on Mac
- [ ] Tested on Linux

This update is not considered required for beta or launch,  but it would be nice to get it in when we can.


jjeffryes:
Superseded by #946

PR #768, opened 2017-09-05
No Mod Section if You Are the Only Mod

### LINKED ISSUE #708: In the Purchase modal, move moderator error message to the Moderator box
Move the `You must select a moderator to make a Moderated Payment.` error message to the Moderator box, since that is where the list of moderator they need to select from are (or will be if they're loading).

image

jjeffryes:
In the purchase screen, the check to not show the moderator section was not accounting for situations where you were the only moderator, and would show a blank list of moderators.

This changes the view and template to use the same list, after the view has removed any IDs that shouldn't be in the list by passing hasModerators in as a template parameter instead of calculating it separately in the template.

This also moves the moderator error to the moderator section, and closes #708.

### DIFF
--- a/js/templates/modals/purchase/purchase.html
+++ b/js/templates/modals/purchase/purchase.html
@@ -1,5 +1,4 @@
 <%
-const hasModerators = !!ob.listing.moderators && ob.listing.moderators.length
 // when multiple listings are supported, the prices array will have one price object for each
 const totalPrice = ob.prices[0].price + ob.prices[0].vPrice;
 %>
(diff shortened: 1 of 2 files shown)


js/templates/modals/purchase/purchase.html
```diff
@@ -97,9 +95,10 @@ <h2 class="h4 required"><%= ob.polyT('purchase.paymentTypeTitle') %></h2>
           <% } %>
         </div>
       </section>
-      <% if (hasModerators) { %>
+      <% if (ob.hasModerators) { %>
         <section class="contentBox padMd clrP clrBr clrSh3 showIfPay js-moderator">
           <h2 class="h4"><%= ob.polyT('purchase.moderatorTitle') %></h2>
+          <div class="js-moderated-errors"></div>
```
rmisio: Why was this added if nothing else was added to handle this? Is this the remnants of a refactor?

jjeffryes: I forgot, that was to fix #708 which is what I was originally working on when I noticed the mod section could show when there were no valid mods.

PR #769, opened 2017-09-05
Update Duo's search endpoint

JustinDrake:
I am not 100% what the plain `/search` endpoint does (as opposed to `/search/listings`). As far as I can tell it is not used. Right now Duo does not return anything for it.

### DIFF
--- a/js/data/defaultSearchProviders.js
+++ b/js/data/defaultSearchProviders.js
@@ -11,12 +11,12 @@ const defaultSearchProviders = [
   },
   {
     id: 'duo',
-    name: 'Duo Search',
+    name: 'Duo',
     logo: '../imgs/duoSearchLogo.png',
-    search: 'http://demo.duosear.ch/search',
-    listings: 'http://demo.duosear.ch/search/listings',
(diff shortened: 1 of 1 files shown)


js/data/defaultSearchProviders.js
```diff
@@ -11,12 +11,12 @@ const defaultSearchProviders = [
...
   {
     id: 'duo',
-    name: 'Duo Search',
+    name: 'Duo',
     logo: '../imgs/duoSearchLogo.png',
-    search: 'http://demo.duosear.ch/search',
-    listings: 'http://demo.duosear.ch/search/listings',
-    torsearch: '',
-    torlistings: '',
+    search: 'https://ob2.duosear.ch/search',
```
jjeffryes: The plain search endpoint will be used in the future for both users and listings (and anything else we add support for), if that endpoint doesn't work currently it's best to leave it blank, and not return a value for it in the results.

It doesn't do anything right now, but we'll be adding that functionality at some point in the future, most of the work for it is already done.

PR #774, opened 2017-09-05
showing wallet transaction memo

rmisio:


jjeffryes:
@rmisio can we make the note wrap if it's long, or do something so you can see the whole thing instead of cutting it off?

image

js/templates/modals/wallet/transaction.html
```diff
@@ -65,8 +65,11 @@
       <div class="" style="flex-shrink: 0;max-width: 80px">
         <div class="noOverflow"><a class="clrT2 js-txidLink"><%= ob.txid %></a></div>
       </div>
-      <div>
-        <div class="noOverflow"><%= ob.translatedMemo %></div>
+      <%
+        const memo = ob.translatedMemo || ob.memo.slice(0, 300);
```
rmisio: The "untruncated" version is shown in a tooltip. Actually, I do cut it off at 300 so it doesn't overwhelm the layout. I don't think the server limits the size, so I want to prevent some ridiculous size from being shown. I would think 300 is big enough for 99.9% of notes.

jjeffryes:
@rmisio can you resolve the merge conflict? Then I think this is good to go if @morebrownies has signed off.

rmisio:
@jjeffryes Can you look into the conflict? I think it deals with code from your patch and you would probably recognize better how to fix it.

jjeffryes:
@rmisio will do.

morebrownies:
I'll take a look once the conflict is resolved, thanks.

jjeffryes:
@morebrownies conflicts should be resolved now.

morebrownies:
The note doesn't seem to display immediately after I send the payment

screenshot 2017-09-06 13 02 22

Once I restart the client it displays

screenshot 2017-09-06 13 02 54

morebrownies:
Also, can you please add clrSh1 to 
`<div class="js-sendConfirm confirmBox sendConfirm arrowBoxTop clrBr clrP clrT">`

screenshot 2017-09-06 13 01 21

screenshot 2017-09-06 13 04 56

rmisio:
I'll look into those things.

rmisio:
@jjeffryes @morebrownies Ok, Mike's last two comments have been addressed.

morebrownies:
Yes, looks good to me now. Thank you @rmisio

PR #780, opened 2017-09-06
Patch wallet note

jjeffryes:
This is proposing an alternate way to handle the memo in the wallet transactions, it will wrap the memo if it is too long.

While that can make some memos larger than others, I think that is more visually appealing than having tooltips on all of them. 

I'm not sure about truncating, maybe we should still truncate the text above a certain character length? Or just limit the memo length on the client and server.

image

### DIFF
--- a/js/templates/modals/wallet/transaction.html
+++ b/js/templates/modals/wallet/transaction.html
@@ -67,7 +67,10 @@
           <div class="noOverflow"><a class="clrT2 js-txidLink"><%= ob.txid %></a></div>
         </div>
       </div>
-      <div class="clrT2 tx6" style="vertical-align: text-top;"><%= ob.translatedMemo || ob.memo %></div>
+      <%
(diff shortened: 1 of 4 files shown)


morebrownies:
Is there any limit in place now @jjeffryes? 

What you have above looks fine to me, but if some memos can grow to be 10x longer, that may introduce some display issues.

jjeffryes:
@morebrownies Rob's PR had a limit of 300 characters, which I think is reasonable.

rmisio:
> What you have above looks fine to me, but if some memos can grow to be 10x longer, that may introduce some display issues.

I don't believe the server imposes a limit, so we should cap it on the client. I would think after three or four lines we could truncate it with an ellipse. In the (probably very rare) case someone puts a note longer than that, they just wouldn't see the full note.

morebrownies:
I can open a ticket on the server to set a limit. 300 seems reasonable to me.

morebrownies:
Issue created: https://github.com/OpenBazaar/openbazaar-go/issues/663

PR #806, opened 2017-09-07
Small fixes

### LINKED ISSUE #789: 'Beta' Shows as a Button and Does Nothing
beta_button_does_nothing

jjeffryes:
This PR addresses a variety of small visual issues. 

- the beta label has a tooltip if the user is not in testnet mode.
- the superfluous info icon next to the coupon label in purchase was removed
- a placeholder was added for the coupon input in purchase
- the coupon input being blank is now handled separately from an invalid code
- the ... button on the user page was removed
- the windows controls were updated to look more like Mac controls and be centered properly
- other nav bar icons that were off center vertically were fixed

Closes #789, 790, 791, 788, and 786

### DIFF
--- a/js/languages/en-US.json
+++ b/js/languages/en-US.json
@@ -1053,7 +1053,8 @@
     "codeErrors": {
       "invalid": "\"%{code}\" is not a valid code",
       "duplicate": "You can only add \"%{code}\" once",
-      "excessive": "You cannot apply discounts that total 100% or more."
+      "excessive": "You cannot apply discounts that total 100% or more.",
+      "blank": "You must enter a code."
     },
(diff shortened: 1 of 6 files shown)


rmisio:
The middle icon gets an underline when you hover over it. I'm assuming it shouldn't since the others don't and it arguably looks cleaner without.

image

morebrownies:
Noticed the same issue that @rmisio reported above too. Otherwise, looks good to me.

jjeffryes:
Huh. Good catch. I'm not sure how that happened. Will investigate and update later tonight.

jjeffryes:
@rmisio underline removed.

PR #816, opened 2017-09-08
Reset serverPage to 0 on filter or sortby change. 

### LINKED ISSUE #814: Blank search page after returning from a listing on second page of search results
v.2.0.10 win7 64 bits

How to reproduce:

1. Open any listing on second page of search results (either OB1 or DUO)
2. While on listing, click on back button on navbar (this error doesn't happen if listing is closed through close modal button)
3. Change any filter (condition, rating, shipping, type or sorting)
4. You get a blank page

error in search pagination

error in search pagination2

jjeffryes:
Closes #814

### DIFF
--- a/js/views/search/Search.js
+++ b/js/views/search/Search.js
@@ -349,12 +349,14 @@ export default class extends baseVw {
 
   changeSortBy(e) {
     this.sortBySelected = $(e.target).val();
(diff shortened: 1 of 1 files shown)


jjeffryes:
I'm merging this now so we can get it into the build, since it's very tiny.

PR #817, opened 2017-09-08
Fetch settings after it is saved in Onboarding.

### LINKED ISSUE #726: About OpenBazaar page not opening on first load
When program is loaded for the first time, about OpenBazaar page doesn't open when clicking on its link in menu. If program is closed and reopened, it starts working.

V. 2.0.9

jjeffryes:
When onboarding completes, app.settings only has data created by the client. If the About modal is opened before some other process fetches settings, it will throw an error because settings is missing the version data.

This fetches the settings after it is saved in onboarding, so all of the settings data from the server is available.

@rmisio there may be a better way to do this, input welcome here.

Closes #726

### DIFF
--- a/js/views/modals/onboarding/Onboarding.js
+++ b/js/views/modals/onboarding/Onboarding.js
@@ -127,6 +127,7 @@ export default class extends BaseModal {
     }
 
     $.when(...saves).done(() => {
+      app.settings.fetch();
       this.trigger('onboarding-complete');
     }).fail((jqXhr) => {
       let title;


rmisio:
> @rmisio there may be a better way to do this, input welcome here.

Yes, the server should send back the settings data (including the server generated data) on a successful POST. I ran this by @cpacia and he is fine with that. He says he'll push shortly.

Closing this one for now.

jjeffryes:
> Yes, the server should send back the settings data (including the server generated data) on a successful POST. I ran this by @cpacia and he is fine with that. He says he'll push shortly.

Ok, great. I thought that was the right answer, but figured there was some reason it wasn't being done already.

PR #818, opened 2017-09-08
Add missing postal code to order details.

jjeffryes:
This adds the missing postal code to the order details modal template.

### DIFF
--- a/js/templates/modals/orderDetail/summaryTab/orderDetails.html
+++ b/js/templates/modals/orderDetail/summaryTab/orderDetails.html
@@ -55,12 +55,16 @@
             <% if (addressLine3) { %>
               <div><%= addressLine3 %></div>
             <% } %>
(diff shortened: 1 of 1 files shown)


rmisio:
Would it be better to put the postal code on the same line with the city and state? That is proper US address format and it would be consistent with how settings is showing the address:

image

If you agree, then you could remove the code this PR puts in and put the following line on line 4 in the template:

`    if (ob.order.shipping.postalCode) addressLine3 = addressLine3.length ? `${addressLine3} ${ob.order.shipping.postalCode}` : ob.order.shipping.postalCode;`

jjeffryes:
> Would it be better to put the postal code on the same line with the city and state?

I was thinking it'd be better separate for international addresses, but that was probably just the lack of coffee. 

I'll change it.

PR #827, opened 2017-09-09
Update shapeshift

jjeffryes:
This PR fixes the payment address for Shapeshift.


jjeffryes:
Update: Shapeshift is working now, so this PR only fixes the variable being incorrect.

PR #833, opened 2017-09-11
Sanitize strings in json response from search

### LINKED ISSUE #832: XSS from search provider
The client will execute JavaScript provided by the search API.

Example:
```
"options":{
  "test":{
    "label":"<script>alert(\"test\");</script>"
    }
}
```

This may or may not have security implications for the user. Could it be used to deanonymize Tor users?

Tested on v2.0.11.

jjeffryes:
Strips dangerous characters from the search JSON to prevent XSS attacks from malicious search providers.

Closes #832 

This can be tested by adding a dangerous option to the data before it is sanitized.

```
pData.options = [
            {
              "test": {
                "label": "<script>alert(\"test filter header\");</script>",
                "type": "dropdown",
                "options": [
                  {
                    "label": "foo",
                    "value": "bar"
                  },
                  {
                    "label": "<script>alert(\"test label\");</script>",
                    "value": "<script>alert(\"test value\");</script>"
                  }
                ]
              }
            },
            {
              "test": {
                "label": "test",
                "type": "<script>alert(\"test filter type\");</script>",
                "options": [
                  {
                    "label": "foo",
                    "value": "bar"
                  }
                ]
              }
            }
            ];
```

### DIFF
--- a/js/views/search/Search.js
+++ b/js/views/search/Search.js
@@ -214,7 +214,15 @@ export default class extends baseVw {
         url: searchUrl,
         dataType: 'json',
       })
-        .done((data, status, xhr) => {
+        .done((pData, status, xhr) => {
+          let data = JSON.stringify(pData, (key, val) => {
+            // sanitize the data from any dangerous characters
+            if (typeof val === 'string') {
+              return val.replace(/["&'\/<>]/g, '');
+            }
+            return val;
+          });
+          data = JSON.parse(data);
           // make sure minimal data is present
           if (data.name && data.links) {
             // if data about the provider is recieved, update the model
--- a/js/views/search/Search.js
+++ b/js/views/search/Search.js
@@ -1,6 +1,7 @@
 import _ from 'underscore';
 import $ from 'jquery';
 import is from 'is_js';
+import sanitizeHtml from 'sanitize-html';
 import baseVw from '../baseVw';
 import loadTemplate from '../../utils/loadTemplate';
 import app from '../../app';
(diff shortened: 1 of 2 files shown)


cpacia:
Wouldn't this break putting in valid tags like bold?

jjeffryes:
> Wouldn't this break putting in valid tags like bold?

Yes, but in this situation there shouldn't be any tags at all.

cpacia:
I'm not sure why. People have put tags in their titles in the past and the server will relay them as long as they pass the sanitizer.

jjeffryes:
The current design doesn't account for tags anywhere in the data we're showing. This would just be for the grid view of results and the filters.

We can add exceptions. @morebrownies would you want things like bold and italic showing up in the titles of search results? Underline?

Those are the only ones I can think of that would be ok. Obviously we wouldn't want someone to wreck the layout with an h2 or a <p> in their title.

cpacia:
If they're aren't displayed then that's ok

morebrownies:
> We can add exceptions. @morebrownies would you want things like bold and italic showing up in the titles of search results? Underline?

I would vote against having text styling in titles. Consistent text styling will provide the best UX imo. Bolding titles would likely be abused to get more attention on listings. I'd rather keep it simple.

jjeffryes:
Just to make things more future-proof, I've updates this to use html-sanitize so it's easy to add in allowed tags or attributes later. 

It's possible we'd want to allow a limited set of tags in the filters.

jjeffryes:
@rmisio this is ready to go.

PR #834, opened 2017-09-11
Add support to build RPM binaries

hoffmabc:
This adds the RPMs into the Travis build script and also fixes a little bug I noticed where we don't specify version for whatever reason in the install script...not sure how it still works!

image

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -60,6 +60,10 @@ case "$TRAVIS_OS_NAME" in
 
     echo 'Install npm packages for Linux'
     npm install --save-dev electron-installer-debian --silent
+    npm install -g --save-dev electron-installer-redhat --silent
+
+    # Install rpmbuild
+    sudo apt-get install rpm
 
     # Ensure fakeroot is installed
     sudo apt-get install fakeroot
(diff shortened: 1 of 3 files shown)


build.sh
```diff
@@ -60,6 +60,10 @@ case "$TRAVIS_OS_NAME" in
 
     echo 'Install npm packages for Linux'
     npm install --save-dev electron-installer-debian --silent
+    npm install -g --save-dev electron-installer-redhat --silent
```
jjeffryes: Why is redhat installed globally but debian isn't?

Is the global flag needed?

hoffmabc: I left old code working as is and so didn't fuck with it, but I can take a look.

hoffmabc: Actually that might be an issue so I'll test it on Travis to see how it goes. Perhaps it doesn't work on a clean run (i.e. continuous integration).

hoffmabc:
Ok added in the global to ensure it's working properly and updated the Travis script. This should be good to go.

jjeffryes:
@rmisio can you give this a quick look before we merge it?

rmisio:
> @rmisio can you give this a quick look before we merge it?

@jjeffryes Sorry, just seeing this not. It LGTM.

PR #836, opened 2017-09-11
Chat tweaks

### LINKED ISSUE #373: Chat messages should be sanitized.
For example, potentially malicious tags like 'script' need to be handled and not output by the app literally.

The chat head would need a similar treatment.

As of now, the server is not sanitizing, but probably should be.

### LINKED ISSUE #389: Links should be clickable in chat conversations
We should make web and OpenBazaar links clickable in chat. Currently they display as plain text.

screen shot 2017-03-09 at 2 00 06 pm

### LINKED ISSUE #793: You can run JS in Chat
run_js_in_chat

rmisio:
This PR processes chat messages so a.) they are sanitized (i.e. do not execute JS) b.) any peerIds, handles (prefaced with an @) and ob links are clickable.

closes #793 
closes #373 
closes #389

### DIFF
--- a/js/models/chat/ChatHead.js
+++ b/js/models/chat/ChatHead.js
@@ -1,3 +1,4 @@
+import { processMessage } from './ChatMessage';
 import app from '../../app';
 import BaseModel from '../BaseModel';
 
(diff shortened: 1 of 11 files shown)


jjeffryes:
@rmisio it looks like you didn't add sanitize-html to package.json.

PR #837, opened 2017-09-11
Remove Shapeshift

jjeffryes:
Comments out the Shapeshift button until it's working with all of our addresses.

Note: don't merge yet until we have a clear ETA on when Shapeshift may be compatible.


jjeffryes:
Note: update this to disable the button, but leave it visible, and show a note explaining this feature is coming soon.

rmisio:
Perhaps we want to mention in the note that they could shapeshift into the wallet if they like?

We could also link to the wallet > receive section. I'd need to make a minor tweak to the wallet launcher in the modal manager (I'll do that as part of the PR that integrates the button into there).

jjeffryes:
> We could also link to the wallet > receive section

I think that's a great idea.

jjeffryes:
Note: add language explaining the address formats, and why some external wallets can't be used with the addresses.

jjeffryes:
@rmisio @morebrownies this is ready for review.

morebrownies:
Will check this out in a bit, thanks

js/languages/en-US.json
```diff
@@ -1118,7 +1118,12 @@
       "note2": "Your payment will be held securely in escrow. If any issues arise, you’ll have the ability to open a dispute.",
       "note3": "*The moderator service fee only applies if a dispute is opened.",
       "note4": "Once you have paid, it may take a bit for the interface to update.",
-      "btnConfirmedPay": "Yes, Pay"
+      "btnConfirmedPay": "Yes, Pay",
+      "altcoinNote": "<b>Altcoins</b> will be available when Shapeshift updates to Segregated Witness.",
```
rmisio: Should we camel-case shapeshift here?

jjeffryes: Yes. Good catch.

morebrownies:
Love the copy on the payment screen, super helpful @jjeffryes 

Unfortunately I can't test if ShapeShift is working correctly since it's blocked in Washington state. 😢 

screenshot 2017-09-22 11 15 40

morebrownies:
Looks good otherwise 👍

rmisio:
I'm surprised the ShapeShift folks can't give a more usable error screen there... perhaps they have legal reasons not to. Anyhow, that's out of our control.

PR #842, opened 2017-09-11
not forcing ssl on remote servers, but defaulting and showing a warning

rmisio:
Since apparently there are people using remote servers on a private network which are probably under protection from net sniffing (e.g. running the server on a PI inside your home wireless network) and it would be a big PIA for those people to have to set-up a certificate, with this PR, we are no longer forcing remote servers to use SSL.

We are defaulting it on for remote servers and showing a warning if you turn it off and try and save the configuration. Beyond that, it's up to the user to be smart and recognize that if you are indeed running the server on a publicly accessible server, you are being foolish to do so without SSL.

### DIFF
--- a/js/languages/en-US.json
+++ b/js/languages/en-US.json
@@ -677,7 +677,13 @@
       "torPwLabel": "Tor Password",
       "torPwPlaceholder": "The unhashed Tor control password",
       "warning": "Warning:",
-      "torServerWarning": "%{warning} In order to be anonymous, it is critical that you also start the server in Tor mode."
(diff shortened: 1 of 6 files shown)


TheButterZone:
Is there an online scanner you recommend for probing your router, with a pi behind it, to see if it is truly protected from net sniffing? Or just a guide on how to lock down DD-WRT & the Pi but still allow OB2 non-admin traffic? Looks like there's a newer version available for WRT54GL... https://www.dd-wrt.com/site/support/router-database

PR #848, opened 2017-09-12
Add languages

jjeffryes:
This adds several languages, and removes the fake pl file that was only used for testing.


jjeffryes:
Closing this so I can resubmit after making sure the git history is clean.

PR #861, opened 2017-09-13
Update French from Transifex

Kirvx:


jjeffryes:
We are now updating the language files with each release. The current release has both the French language files as of yesterday.

jjeffryes:
I'll merge this PR, but you don't need to submit PRs for language updates in the future, @Kirvx, I update all the languages at once when we do a release. 

Thank you for submitting it.

PR #878, opened 2017-09-14
Set default fee level to normal.

jjeffryes:


rmisio:
@jjeffryes We're fine with the fact that this will only apply to fresh installs of the app? Any existing user's level will remain unchanged?

rmisio:
> @jjeffryes We're fine with the fact that this will only apply to fresh installs of the app? Any existing user's level will remain unchanged?

Actually, nevermind, we're adding in that piece where it will be clear what your fee level is and they'll be a link to change it.

jjeffryes:
> Actually, nevermind, we're adding in that piece where it will be clear what your fee level is and they'll be a link to change it.

I was literally typing that when your message appeared. :)

PR #889, opened 2017-09-15
External ob links

### LINKED ISSUE #862: the client needs to handle external ob:// links
Looks like main.js is targetting `ob2` as external links, but we decided to use `ob` even though it overlaps with v1.

It will be interesting to see which app opens if the user has both versions installed (this behavior may vary by OS).

rmisio:
This PR updates the code to handle external ob links.

Some notes:
- To test this properly will require a new build. @hoffmabc 
- Since we are using the `ob://` protocol to open links in both the v1 and v2 apps, it is not entirely known which one will take precedence (I imagine it's dependent on the OS). For me on OSX, with both installed, the link still opened in v1. I've come to know it's quite a pain to break the association on OSX, so I fixed things by just deleting the v1 app.
- The electron method we are using to associate the protocol (`setAsDefaultProtocolClient`) doesn't list Linux as being supported. So, I would guess this would not work on Linux. Not sure if v1 supported external ob links on Linux?

closes #862

### DIFF
--- a/js/router.js
+++ b/js/router.js
@@ -1,4 +1,5 @@
 import $ from 'jquery';
+import { ipcRenderer } from 'electron';
 import { Router } from 'backbone';
 import { getGuid, isMultihash } from './utils';
 import { getPageContainer } from './utils/selectors';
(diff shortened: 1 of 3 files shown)


roberfoster:
These methods don't work on Linux as you've supposed. Electron suggest to use this one

[makesingleinstance](https://electronjs.org/docs/api/app#appmakesingleinstancecallback) for Linux based OS.

PR #890, opened 2017-09-15
Handle NSFW in Search

jjeffryes:
- fixes the checkbox filter in search.
- sets the NSFW filter to the user's default on the initial search.
- uncomments out the showNSFW setting in settings/general.

Currently the NSFW filter in the OB1 search uses radio buttons. Once this PR is merged we'll switch it to be a checkbox.


jjeffryes:
@rmisio this is ready to go.

The OB1 search API isn't updated yet to have working checkboxes, you can wait to review this until then if you want. I would like to get this in for release so we can add the part that filters search by the user's nsfw setting automatically.

jjeffryes:
@rmisio the search API has been updated, and works the way this PR expects now.

PR #902, opened 2017-09-19
Restrict Moderation if the Purchase Price is Low

### LINKED ISSUE #795: Shipping options don't update price after address is added or removed
v.2.0.10 win7 64bits

1. Open a listing with multiple shipping options (ob://QmcNqUgkrWpy4XfbjPYX7ZicU23LLLVP65Y5vpmr9epgyd/store/test-physical-listing).
2. Add or remove an address in settings and close settings modal.
3. Go back to listing already opened.
4. When you click on each shipping option, shipping price is not updated anymore on Order Summary

ship options

### LINKED ISSUE #852: Restrict Moderated Purchases
When making a moderated purchase, the network fees are taken out of the total payment, reducing the funds the seller receives (unlike direct payments, where they are paid separately by the buyer). 

Because of this, if a listing is priced lower than the fee, it will probably fail, and if it is priced close to the fee the seller will receive a minimal payment. The transaction can also become impossible to moderate, as the moderator's fee is taken out of the total, which can reduce it to zero.

To fix this, the following changes should be made:

- [x] When purchasing, the option to make the purchase moderated should be disabled or hidden unless the purchase total is above the estimated fee times 10. 
- [x] Call the fee per byte API and multiply the Normal by 350 to get the fee.
- [x] If the user changes the shipping or quantity, and it changes the purchase total, the option to make the purchase moderated should update appropriately.
- [x] When the option to make the purchase moderated is inactive, explanatory text should be shown. 
    - ex: "The total price of this purchase is too low for moderation."

image

### LINKED ISSUE #908: Duplicate mods present during checkout
Moderators are duplicated during checkout and both duplicates are selectable:

ob://QmScF5n99EHZvgiNTc3ygXQc3PvwN6CXi1rkBgE3nEQFwK/store/apple-wireless-keyboard

image

jjeffryes:
This will check the total price of the current order on the purchase page, and if it's below the required total, disable the option to select a moderator.

This also fixes a bug where changing the shipping option did not update the receipt.

Closes #852 
Closes #795 
Closes #908

### DIFF
--- a/js/languages/en-US.json
+++ b/js/languages/en-US.json
@@ -1088,7 +1088,7 @@
     "quantity": "Quantity",
     "useShapeShift": "Or use ShapeShift in the next step",
     "moderatedPayment": "Moderated Payment",
-    "tooLowForMods": "Based on the current estimated fee, moderated payment is not available for orders under %{minModPrice}. A direct payment will be used instead.",
+    "tooLowForMods": "Based on the current estimated fee, the moderated payment option is not available for orders under %{minModPrice}. A direct payment will be used instead.",
     "noValidModerators": "No valid moderators are available for this order.",
     "moderationDisclaimer": "Your payment will be held securly in escrow. If any issues arise, you’ll have the ability to open a dispute with a moderator. *The moderator service fee only applies if a dispute is opened.",
     "chooseModerator": "Choose This Moderator",
--- a/js/languages/en-US.json
+++ b/js/languages/en-US.json
@@ -1143,7 +1143,8 @@
       "ownIDTitle": "You Can't Purchase Your Own Listings",
       "ownIDMsg": "You can preview the purchase process for your own listings, but you cannot purchase them.",
       "paymentFailed": "The Payment Could Not Be Completed",
-      "zeroPrice": "You cannot purchase a listing with a total price of zero."
+      "zeroPrice": "You cannot purchase a listing with a total price of zero.",
+      "feesFailed": "Transaction Fees Cannot Be Loaded"
     }
   },
   "moderators": {
--- a/js/languages/en-US.json
+++ b/js/languages/en-US.json
@@ -1088,7 +1088,7 @@
     "quantity": "Quantity",
     "useShapeShift": "Or use ShapeShift in the next step",
     "moderatedPayment": "Moderated Payment",
-    "tooLowForMods": "The total price is too low for this order to be moderated.",
+    "tooLowForMods": "Based on the current estimated fee, moderated payment is not available for orders under %{minModPrice}. A direct payment will be used instead.",
     "noValidModerators": "No valid moderators are available for this order.",
     "moderationDisclaimer": "Your payment will be held securly in escrow. If any issues arise, you’ll have the ability to open a dispute with a moderator. *The moderator service fee only applies if a dispute is opened.",
     "chooseModerator": "Choose This Moderator",
--- a/js/languages/en-US.json
+++ b/js/languages/en-US.json
@@ -1079,6 +1079,7 @@
     "quantity": "Quantity",
     "useShapeShift": "Or use ShapeShift in the next step",
     "moderatedPayment": "Moderated Payment",
+    "tooLowForMods": "The total price is too low for this order to be moderated.",
     "noValidModerators": "No valid moderators are available for this order.",
     "moderationDisclaimer": "Your payment will be held securly in escrow. If any issues arise, you’ll have the ability to open a dispute with a moderator. *The moderator service fee only applies if a dispute is opened.",
     "chooseModerator": "Choose This Moderator",
--- a/js/templates/modals/purchase/receipt.html
+++ b/js/templates/modals/purchase/receipt.html
@@ -4,12 +4,12 @@
   const listingCurrency = ob.listing.metadata.pricingCurrency;
   const viewingCurrency = ob.displayCurrency;
   const isBTC = viewingCurrency === 'BTC';
-  ob.items.forEach((item, i) => {
+  ob.prices.forEach((priceObj, i) => {
     // convert and truncate the prices here, to prevent rounding errors in the display
-    const basePrice = Number(ob.formatPrice(ob.convertCurrency(ob.prices[i].price, listingCurrency, viewingCurrency), isBTC));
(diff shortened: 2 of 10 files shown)


js/templates/modals/purchase/receipt.html
```diff
@@ -4,12 +4,12 @@
   const listingCurrency = ob.listing.metadata.pricingCurrency;
   const viewingCurrency = ob.displayCurrency;
   const isBTC = viewingCurrency === 'BTC';
-  ob.items.forEach((item, i) => {
+  ob.prices.forEach((priceObj, i) => {
```
jjeffryes: This was shifted to the priceObj, since it had the same data in a more concise form, and is now always up to date.

js/views/modals/listingDetail/Listing.js
```diff
@@ -426,6 +426,11 @@ export default class extends BaseModal {
       vendor: this.vendor,
       removeOnClose: true,
       showCloseButton: false,
+      initialState: {
+        isFetching: true,
+        fetchError: '',
+        fetchFailed: false,
```
jjeffryes: In the future it would be good to move more of the template to using the state, I spent some time retrofitting it, but backed those changes out because it would require a lot of refactoring. For now the state only handles the initial load and any errors.

js/views/modals/purchase/Purchase.js
```diff
@@ -237,7 +294,11 @@ export default class extends BaseModal {
   changeCoupons(hashes, codes) {
     // combine the codes and hashes so the receipt can check both.
     // if this is the user's own listing they will have codes instead of hashes
-    this.receipt.coupons = hashes.concat(codes);
+    const hashesAndCodes = hashes.concat(codes);
+    const filteredCoupons = this.listing.get('coupons').filter((coupon) =>
+      hashesAndCodes.indexOf(coupon.get('hash') || coupon.get('discountCode')) !== -1);
+    this.couponObj = filteredCoupons.map(coupon => coupon.toJSON());
+    this.receipt.coupons = this.couponObj;
```
jjeffryes: This was moved from the receipt view to here, so the coupon data is available for the calculation of the current total.

morebrownies:
Hey @jjeffryes 

I'm a bit worried the copy may trip some people up:

`The total price is too low for this order to be moderated.`

Initial thoughts that came to mind... Does that mean I can't check out? Do I need to increase the total of my order before checking out? I'm not sure? 

To eliminate possible confusion, maybe copy like this would help: 

`Moderated payments are not available for purchases at this price point. A direct payment will be used instead.`

Thoughts?

Oh and could we bump the text down from `1.5rem` to `1.3rem`. Feels a bit large as is.

jjeffryes:
@morebrownies good point, it might not be clear what that message is referring to. 

The fact it can turn on and off depending on quantity and shipping options is probably confusing too.

It would be nice if we didn't play coy about the numbers, even though they change.

How about:

"Based on the current estimated fee, moderated payment is not available for orders under [amount]."

You think maybe it should be green instead of red, so it doesn't suggest there's an error?

morebrownies:
I like the idea of using the actual [amount] value. Maybe we can combine both of our approaches? I like having `A direct payment will be used instead` at the end of the message because it reinforces the user can still proceed with checking out. 

`"Based on the current estimated fee, moderated payment is not available for orders under [amount]. A direct payment will be used instead."`

jjeffryes:
Sounds good. What about changing the color from red?

morebrownies:
The red text does feel like you've hit an error, especially since all of the other errors in the checkout experience display red. 

Maybe we should make it black text to differentiate it from an error state? It's not really an error, it's more so information that is helpful to know.

jjeffryes:
I made it bold and normal color.

morebrownies:
👍

js/views/modals/purchase/Purchase.js
```diff
@@ -131,6 +142,7 @@ export default class extends BaseModal {
     return {
       'click .js-goToListing': 'close',
```
rmisio: This is minor, but IMO `goToListing` is the wrong semantics here. We happen to open the Purchase modal above the listing, so closing it will take it to the listing, but it doesn't have to be that way. In theory, the Purchase modal can be opened above anything, so semantically `js-close` (or something to indicate it's just a close) would be more appropriate.

rmisio:
I'm running into this bug on this branch (not sure if it also happens on master):
https://github.com/OpenBazaar/openbazaar-desktop/issues/908

rmisio:
Also finding that when I first open the Purchase modal, Moderated Payment is checked by default, but there is not a list of mods to select from. When I click Pay, nothing happens. I presume that's because there's an error instructing me to select a mod, but that error is hidden along with the hidden mod list.

image

js/languages/en-US.json
```diff
@@ -1088,6 +1088,7 @@
     "quantity": "Quantity",
     "useShapeShift": "Or use ShapeShift in the next step",
     "moderatedPayment": "Moderated Payment",
+    "tooLowForMods": "Based on the current estimated fee, moderated payment is not available for orders under %{minModPrice}. A direct payment will be used instead.",
```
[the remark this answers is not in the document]
jjeffryes: It's singular because it's referring to the control, which is labeled "Moderated Payment", not the general idea of moderated payments. It could be made more clear though.

js/views/modals/purchase/Purchase.js
```diff
@@ -115,8 +120,14 @@ export default class extends BaseModal {
       vendor: this.vendor,
     });
 
+    // on the initial load, fetch the fees
+    this.getFees();
+
     this.listenTo(app.settings, 'change:localCurrency', () => this.showDataChangedMessage());
     this.listenTo(app.localSettings, 'change:bitcoinUnit', () => this.showDataChangedMessage());
+    this.listenTo(this.order.get('items').at(0), 'change', () => this.refreshPrices());
```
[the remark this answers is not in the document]
jjeffryes: In this case really only the quantity or the shipping changing are important, and the shipping is listened to on the next line. I'll update to someChange so one listener can catch both.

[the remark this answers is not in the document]
jjeffryes: Looks like someChange only catches nested models, not collections.

PR #904, opened 2017-09-19
Fee estimate tweaks

rmisio:
This PR updates the client to account for the changes to the fee estimation apis on the backend.


js/utils/fees.js
```diff
@@ -9,41 +11,126 @@ export const feeLevels = [
...
+let getFeesCache = {};
+
+/**
+ * Will call the fees api ('wallet/fees') on the server.
+ *
+ * @return {object} An jQuery defered which on success will resolve with the fees
+ *   per byte in Satoshi for each fee level. If the call fails, the deferred will
+ *   fail and pass on the args the xhr fail handler receives.
+ */
+export function getFees() {
```
rmisio: @jjeffryes Turns out I also needed to use `wallet/fees`, so I integrated it into the fees util module. It gives us caching which makes it easier to use it in certain use-cases. For example, let's say you have a list of transaction views and they all need to get the estimated fee. Instead of having to centralize it in a parent and pass it down, each child could just call the api knowing that the util module will batch them into one call.

It's probably not super urgent, but if you wanted to, feel free to plug this into the Purchase modal. Should be mostly as simple as replacing your `$.get()` call with a `getFees()` call.

Also, don't worry about aborting the cal. The returned promise doesn't expose an abort method. You wouldn't want to blindly abort, since another process may also be waiting on the call.

js/utils/fees.js
```diff
@@ -9,41 +11,126 @@ export const feeLevels = [
...
+let getFeesCache = {};
+
+/**
+ * Will call the fees api ('wallet/fees') on the server.
+ *
+ * @return {object} An jQuery promise which on success will resolve with the fees
+ *   per byte in Satoshi for each fee level. If the call fails, the deferred will
+ *   fail and pass on the args the xhr fail handler receives.
+ */
+export function getFees() {
```
rmisio: @jjeffryes Turns out I also needed to use wallet/fees, so I integrated it into the fees util module. It gives us caching which makes it easier to use it in certain use-cases. For example, let's say you have a list of transaction views and they all need to get the estimated fee. Instead of having to centralize it in a parent and pass it down, each child could just call the api knowing that the util module will batch them into one call.

It's probably not super urgent, but if you wanted to, feel free to plug this into the Purchase modal. Should be mostly as simple as replacing your $.get() call with a getFees() call.

Also, don't worry about aborting the call. The returned promise doesn't expose an abort method. We wouldn't want to blindly allow an abort, since another process may also be waiting on the call.

Issue #824, opened 2017-09-09
Feature Request: Full Screen on Maximize


ekerstein:
This is a minor user interface annoyance. The standard behavior on mac os is that the green "maximize / full screen" button takes you directly into full screen view.

![maximize](https://user-images.githubusercontent.com/6055508/30238148-58085864-950f-11e7-8c6b-a640644a6057.png)

However, on OB clicking the green button simply expands the window without taking you to full screen. You have to go to View -> Toggle Full Screen to do that.

This deviates from standard mac behavior and may be a minor annoyance to new users, as it was for me.

jjeffryes:
This was closed by #905

PR #905, opened 2017-09-20
Enable fullscreen for Mac

drwasho:
The green maximise button in macOS makes native apps go fullscreen. Previously a generic maximise function was being used, which works for Linux, Mac, and Windows, which worked but it was weird for Mac users.

This commit uses the fullscreen function if the OS is a Mac.

### DIFF
--- a/js/views/PageNav.js
+++ b/js/views/PageNav.js
@@ -225,19 +225,7 @@ export default class extends BaseVw {
   }
 
   navMaxClick() {
-    if (remote.process.platform === 'darwin') {
-      if (remote.getCurrentWindow().isFullScreen()) {
-        remote.getCurrentWindow().setFullScreen(false);
-      } else {
-        remote.getCurrentWindow().setFullScreen(true);
-      }
-    } else {
-      if (remote.getCurrentWindow().isMaximized()) {
-        remote.getCurrentWindow().unmaximize();
-      } else {
-        remote.getCurrentWindow().maximize();
-      }
-    }
+    remote.getCurrentWindow().setFullScreen(!remote.getCurrentWindow().isFullScreen());
   }
 
   onRouteSearch() {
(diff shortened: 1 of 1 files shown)


drwasho:
This need testing on Linux and Windows, and is working in my hands on Mac.

rmisio:
This LGTM. I'll leave it to @jjeffryes to test on Windows.

rmisio:
> This LGTM. I'll leave it to @jjeffryes to test on Windows.

@drwasho oops, I spoke too soon. You have a linting error. Would you mind fixing it? (`npm run lint`)

js/views/PageNav.js
```diff
@@ -225,10 +225,18 @@ export default class extends BaseVw {
   }
 
   navMaxClick() {
-    if (remote.getCurrentWindow().isMaximized()) {
-      remote.getCurrentWindow().unmaximize();
+    if (remote.process.platform === 'darwin') {
```
jjeffryes: There's no need to do the check for the platform, Windows also supports fullScreen, and that's the expected behavior when clicking this button. 

It's also what the right click, view, toggle fullscreen option does.

Using maximize originally was probably just an oversite.

Take out the if and just do the part with setFullscreen.

If you want to be very concise, you can do it as:
```
navMaxClick() {
    remote.getCurrentWindow().setFullScreen(!remote.getCurrentWindow().isFullScreen());
  }
```

rmisio: So we understand... this type of behavior of maximize is for new OSs. I remember when I used Windows, maximize didn't do full screen. I'm on OS mavericks and it doesn't do fullscreen for maximize either.

I have no problem with this change, but we'll probably get a bug or two from users on old OSs, who may not like and I think we'd have to live with that. I certainly wouldn't want to have to detect the OS version and selectively decide based off of that.

rmisio: > There's no need to do the check for the platform, Windows also supports fullScreen, and that's the expected behavior when clicking this button.

What about Linux? Linux may do a more traditional maximize (i.e. not fullscreen).

morebrownies:
I just noticed there is an issue open for this feature: https://github.com/OpenBazaar/openbazaar-desktop/issues/824

Tag it so we can close it out 👍

PR #920, opened 2017-09-21
added Fund Via ShapShift button to the receive section of the wallet

rmisio:


rmisio:
@jjeffryes This is ready to go. I actually didn't need to make any changes to support linking to the Receive section. Just put the following in your click handler:

```
import { launchWallet } from '../../../utils/modalManager';
launchWallet().sendModeOn = false;
```

PR #921, opened 2017-09-22
showing a spinner on listing cards until the listing image loads

### LINKED ISSUE #875: Display a spinner when listing photos are loading
When a listing photo is loading, it displays as an empty white card, and it's not clear if it's loading or doesn't have a photo

screenshot 2017-09-14 12 50 11

rmisio:
closes #875


js/views/ListingCard.js
```diff
@@ -265,6 +295,25 @@ export default class extends baseVw {
...
+
+        // there will not be an imgUrl if the listing image failed to load
+        if (imgUrl) {
+          url = `url("${imgUrl}"), ${url}`;
+        }
+
+        this.getCachedEl(`.js-${this.viewType}ViewListingImage`)[0]
+          .style
+          .backgroundImage = url;
+        this.getCachedEl('.js-listingImageLoadSpinner')
```
rmisio: Based on some experiments I did along with some thoughts in [this thread](https://stackoverflow.com/questions/46333377/will-hidden-obscured-svg-animations-still-cause-browser-repaints-and-or-perfor), I'm removing the spinner svg once the image loads. Even if it is obscured or hidden, it's taking up resources.

styles/components/_misc.scss
```diff
@@ -99,3 +99,8 @@
     opacity: .35;
   }
 }
+
+// This could probably go into an animations components file once we have a couple
+.fadeIn {
```
rmisio: This isn't being used. I was experimenting with having the image fade in on load, but there are some complications on that since it will re-fade in on re-renders. Also, a re-render mid animation doesn't look quite right.

But... leaving the fadeIn animation class in since it could be useful.

PR #922, opened 2017-09-22
added in the ability to clone your own listing from the listing card

### LINKED ISSUE #859: Add ability to easily clone a listing
We had it in 1.0 and it's highly requested in 2.0 already. I'll add designs for it when I get a chance and we can work it into our queue.

rmisio:
@morebrownies I jumped the gun on this one. It was requested enough times and the scope was relatively low, so I just knocked it out. Let me know if you think any tweaks are needed.

I just took a guess on the icon:

image

closes #859

### DIFF
--- a/js/views/ListingCard.js
+++ b/js/views/ListingCard.js
@@ -137,6 +137,7 @@ export default class extends baseVw {
         if (xhr.statusText === 'abort' || this.isRemoved()) return;
         const clonedModel = this.fullListing.clone();
         clonedModel.unset('slug').guid = this.ownerGuid;
+        clonedModel.lastSyncedAttrs = {};
 
         this.editModal = launchEditListingModal({
           model: clonedModel,
--- a/js/views/ListingCard.js
+++ b/js/views/ListingCard.js
@@ -52,8 +52,6 @@ export default class extends baseVw {
       this.$el.addClass('ownListing');
     }
 
-    this.fullListingFetches = [];
-
     if (this.ownListing) {
       this.listenTo(listingEvents, 'destroying', (md, destroyingOpts) => {
         if (this.isRemoved()) return;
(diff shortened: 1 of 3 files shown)


morebrownies:
This feature is awesome. Icon looks good. 👍

js/views/ListingCard.js
```diff
@@ -108,55 +107,63 @@ export default class extends baseVw {
...
   }
 
-  onClickConfirmedDelete() {
-    if (this.destroyRequest && this.destroyRequest.state === 'pending') return false;
+  onClickClone(e) {
+    app.loadingModal.open();
+
+    this.fetchFullListing()
+      .done(xhr => {
+        if (xhr.statusText === 'abort' || this.isRemoved()) return;
```
jjeffryes: Same comment here, if the .always callback was added first, couldn't it handle all the checking for abort and isRemoved?

js/views/ListingCard.js
```diff
@@ -108,55 +107,63 @@ export default class extends baseVw {
   onClickEdit(e) {
     app.loadingModal.open();
 
-    const fullListingFetch = this.fullListing.fetch()
-      .done(() => {
-        if (fullListingFetch.statusText === 'abort' || this.isRemoved()) return;
+    this.fetchFullListing()
+      .done(xhr => {
+        if (xhr.statusText === 'abort' || this.isRemoved()) return;
```
jjeffryes: If you add the .always callback first, couldn't it handle the checking for abort and isRemoved?

rmisio: Even if you opt out early out of the always() callback, that will have no bearing on the done() or fail() callbacks, which will still fire.

Am I misunderstanding you?

jjeffryes: No, you're right, I was misthinking thing.

PR #946, opened 2017-09-26
Client Only Binaries for Build Process

hoffmabc:
Still testing.

* Creates client-only builds for Linux, OSX and Windows
* Bumps Electron to 1.7.8
* Ensures npm is 5.2 for a build error
* Cleans up some redundant babel packages in package.json


jjeffryes:
This will need some thorough testing to make sure the update to Electron 1.7.8 doesn't cause any surprises.

hoffmabc:
I need to test the build on the Linux side, but OSX and Windows look ok for now.

PR #951, opened 2017-09-26
Social buttons

### LINKED ISSUE #901: "I Just Bought Something on OpenBazaar" Social Media Sharing Tool
As a user, when I purchase something, I want an easy way to immediately send a message to social media about my purchase.

It should explain that I just made a purchase on OpenBazaar and let me select between 2 types of messages, a "general" message, such as "I just made a purchase on OpenBazaar!" with a link to download OpenBazaar OR a "specifc" message that shares the item I bought, that I bought it on OpenBazaar, and have a link to the item.

### LINKED ISSUE #948: Escrow payment text showing in payment screen on direct order
pasted image at 2017_09_26 02_08 pm

This is a direct order, but notice the text below the QR code stating it is an escrow order and that user can open a dispute.

The vendor did have a moderator, but I was unable to use moderated payments since the order size was too small.

jjeffryes:
Adds social sharing buttons to the payment complete screen.

Also fixes a bug where a race condition caused the first moderator to be invisibly selected when the order total was too low for a moderator. 

Closes #901 
Closes #948


js/views/modals/purchase/Moderators.js
```diff
@@ -184,7 +184,7 @@ export default class extends baseVw {
       this.removeNotFetched(model.id);
       this.modCards.push(newModView);
       // if required, select the first  moderator
-      if (this.options.selectFirst && !this.firstSelected) {
+      if (this.options.selectFirst && !this.firstSelected && !this.noneSelected) {
```
jjeffryes: The noneSelected can be passed in to prevent the first moderator from being selected when the total is too low.

js/views/modals/purchase/Purchase.js
```diff
@@ -250,18 +251,20 @@ export default class extends BaseModal {
   disableModerators() {
     this.getCachedEl('#purchaseModerated').prop('checked', false);
     this.order.moderated = false;
+    this.moderators.deselectOthers();
     this.order.set('moderator', '');
   }
 
-  toggleModeration(bool) {
+  moderationOn(bool) {
```
jjeffryes: The name was changed here so it's more obvious what it's doing, and whether the value passed in should be true or false.

PR #952, opened 2017-09-26
Fix race condition in moderation

### LINKED ISSUE #948: Escrow payment text showing in payment screen on direct order

jjeffryes:
Fixes a bug where a race condition caused the first moderator to be invisibly selected when the order total was too low for a moderator.

Closes #948

### DIFF
--- a/js/views/modals/purchase/Complete.js
+++ b/js/views/modals/purchase/Complete.js
@@ -20,7 +20,7 @@ export default class extends BaseVw {
     this.processingTime = this.options.listing.get('item').processingTime ||
       app.polyglot.t('purchase.completeSection.noData');
     this.vendorPeerID = this.options.listing.get('vendorID').peerID;
-    this.orderID = ''; // supplied by pending.js when the order is paid for
+    this._orderID = '';
   }
 
   className() {
@@ -77,6 +77,14 @@ export default class extends BaseVw {
     e.preventDefault();
     this.sendMessageInput();
     e.preventDefault();
+  }
+
+  get orderID() {
+    return this._orderID;
+  }
+
+  set orderID(ID) {
+    if (ID !== this._orderID) this._orderID = ID;
   }
 
   get $messageInput() {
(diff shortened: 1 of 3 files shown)


js/views/modals/purchase/Complete.js
```diff
@@ -79,6 +79,14 @@ export default class extends BaseVw {
     e.preventDefault();
   }
 
+  get orderID() {
```
jjeffryes: This is just cleaning up the code to use a getter and setter.

js/views/modals/purchase/Moderators.js
```diff
@@ -184,7 +184,7 @@ export default class extends baseVw {
       this.removeNotFetched(model.id);
       this.modCards.push(newModView);
       // if required, select the first  moderator
-      if (this.options.selectFirst && !this.firstSelected) {
+      if (this.options.selectFirst && !this.firstSelected && !this.noneSelected) {
```
jjeffryes: noneSelected is set if no moderators should be selected.

This should be set to false if the total of the order is too low for moderation, since deselecting all the moderators won't affect moderators that aren't loaded yet.

js/views/modals/purchase/Purchase.js
```diff
@@ -250,18 +251,20 @@ export default class extends BaseModal {
   disableModerators() {
     this.getCachedEl('#purchaseModerated').prop('checked', false);
     this.order.moderated = false;
+    this.moderators.deselectOthers();
     this.order.set('moderator', '');
   }
 
-  toggleModeration(bool) {
+  moderationOn(bool) {
```
jjeffryes: This was renamed to make the it more clear what true or false for the passed in value will do.

PR #955, opened 2017-09-27
Modularize Follow/Message Buttons

### LINKED ISSUE #600: message button on product page
it would be great to have a message button on the store page, for easy interaction.

auswahl_001

### LINKED ISSUE #807: When viewing a listing that is not mine, display message/follow buttons next to the close button
Most users will land on the listing detail screen via search. Currently, it's not possible to message the vendor from that view. They'll need to know to close the modal and click the message button from the store page. 

screenshot 2017-09-07 14 34 24

jjeffryes:
This modularizes the follow/unfollow and message button, and uses the new module in 3 places:
- replace the old buttons in the user page
- replace the old buttons in the moderator details modal
- add them to the listing page when viewing a listing that isn't yours

Closes #807 
Closes #600


morebrownies:
Looking good @jjeffryes

js/views/SocialBtns.js
```diff
@@ -0,0 +1,63 @@
...
+export default class extends BaseVw {
+  constructor(options = {}) {
+    if (!options.targetID) throw new Error('You must provide a targetID');
+
+    super(options);
+    this.options = options;
+
+    this._state = {
+      following: followedByYou(options.targetID),
+      processing: false,
```
rmisio: Could we call this isFollowing instead of the generic processing? The latter wouldn't work well if you had a need to add a seconds processing button in here.

js/views/SocialBtns.js
```diff
@@ -0,0 +1,63 @@
+import app from '../app';
```
rmisio: Could we move this to the components folder?

js/views/SocialBtns.js
```diff
@@ -0,0 +1,63 @@
...
+    };
+  }
+
+  onClickMessage() {
+    // activate the chat message
+    app.chat.openConversation(this.options.targetID);
+  }
+
+  onClickFollow() {
+    const type = this._state.following ? 'unfollow' : 'follow';
```
rmisio: Can we use getState() rather than referencing the private _state variable?

js/templates/socialBtns.html
```diff
@@ -0,0 +1,7 @@
+<div class="<%= ob.stripClasses %>">
+  <a class="btn <%= ob.btnClasses %> js-messageBtn"><%= ob.polyT('userPage.message') %></a>
+  <%= ob.processingButton({
+  className: `btn ${ob.btnClasses} ${ob.processing ? 'processing' : ''} js-followUnfollowBtn`,
```
rmisio: Could we indent the inner two lines here?

js/views/SocialBtns.js
```diff
@@ -0,0 +1,63 @@
...
+  constructor(options = {}) {
+    if (!options.targetID) throw new Error('You must provide a targetID');
+
+    super(options);
+    this.options = options;
+
+    this._state = {
+      following: followedByYou(options.targetID),
+      processing: false,
+      ...options.initialState || {},
```
rmisio: It may not be a bad idea to set default values for btnClasses and stripClasses since I imagine those might be fine for a lot of implementations. I would also probably set them as state, rather than options.

js/views/modals/listingDetail/Listing.js
```diff
@@ -528,6 +529,16 @@ export default class extends BaseModal {
       this.$reviews = this.$('.js-reviews');
       this.$reviews.append(this.reviews.render().$el);
 
+      if (!this.model.isOwnListing) {
+        if (this.socialBtns) this.socialBtns.remove();
+        this.socialBtns = this.createChild(SocialBtns, {
+          targetID: this.vendor.peerID,
+          stripClasses: 'btnStrip clrSh3',
```
rmisio: It may not be a bad idea to pass in stripClasses and btnClasses in as state, rather than options. The benefit of that is you could dynamically change them then.

js/views/SocialBtns.js
```diff
@@ -0,0 +1,63 @@
+import app from '../app';
+import loadTemplate from '../utils/loadTemplate';
+import { followedByYou, followUnfollow } from '../utils/follow';
+
```
rmisio: Is this empty line needed?

PR #957, opened 2017-09-28
More translations

Kirvx:


js/languages/en-US.json
```diff
@@ -1020,6 +1027,10 @@
     "addQueryProviderBtn": "Add search provider",
     "chooseDefaultTitle": "Select a default search provider",
     "chooseDefaultMsg": "This provider will be used as the default for future seaches until you set a new one.",
+    "providers": {
+        "transactions": "Transactions",
+        "myPage": "My Page",
```
jjeffryes: In json, you can't have a comma on the last item in a set. This comma will prevent the app from starting.

Always check to make sure nothing breaks after editing any file.

js/languages/en-US.json
```diff
@@ -133,6 +133,13 @@
     "torOnTooltip": "Tor active",
     "discoverCalloutTitle": "Discover",
     "discoverCalloutBody": "Browse and shop hundreds of stores on OpenBazaar",
+    "myPage": "My page",
```
jjeffryes: The P in page should be capitalized, to match the other links.

PR #964, opened 2017-09-29
Search if more than 10 select options

jjeffryes:
Changes the parameter to show the select list search input if there are more than 10 options.

Currently none of the filters have more than 10 options, Tyler is adding more to the shipping filter.

### DIFF
--- a/js/views/search/Search.js
+++ b/js/views/search/Search.js
@@ -437,8 +437,7 @@ export default class extends baseVw {
     });
     const filterWrapper = this.$('.js-filterWrapper');
     filterWrapper.find('select').select2({
-      // disables the search box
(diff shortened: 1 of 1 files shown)


morebrownies:
> Currently none of the filters have more than 10 options, Tyler is adding more to the shipping filter.

Is there anyway to confirm this is working correctly? Or do we need to wait on Tyler to add more regions first?

jjeffryes:
You could edit the number to be 3, or some other small amount, to test it.

rmisio:
This one LGTM.

rmisio:
@morebrownies This is waiting on you're final OK. Can you take a look when you have a chance please?

morebrownies:
Looks good to me 👍

PR #968, opened 2017-09-29
960 - fixing uncaught exception stemming from the server allowing unc…

### LINKED ISSUE #960: Uncaught Wallet Error
When I have an input to the wallet with zero confirmations that is large enough to cover the price of a purchase, the pay from wallet button allows me to try to pay from the wallet, but Spend.js throws an unhandled error.

_Spend.js:132 Uncaught Error: amount: You have insufficient funds to send such an amount. Be sure to account for the transaction fee._

This leaves the button in the processing state, and the user has to close the modal to proceed.

image

image

image

rmisio:
…onfirmed spends and the client not

closes #960

### DIFF
--- a/js/views/modals/purchase/Payment.js
+++ b/js/views/modals/purchase/Payment.js
@@ -59,9 +59,6 @@ export default class extends BaseVw {
         }
       });
     }
-
-    console.log('sugar');
-    window.sugar = app;
   }
 
   set balanceRemaining(amount) {
--- a/js/views/modals/purchase/Payment.js
+++ b/js/views/modals/purchase/Payment.js
@@ -59,6 +59,9 @@ export default class extends BaseVw {
         }
       });
     }
+
+    console.log('sugar');
+    window.sugar = app;
   }
 
   set balanceRemaining(amount) {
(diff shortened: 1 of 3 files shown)


js/views/modals/purchase/Payment.js
```diff
@@ -89,26 +89,44 @@ export default class extends BaseVw {
   }
 
   clickPayFromWallet(e) {
-    this.spendConfirmBox.setState({ show: true });
-    this.spendConfirmBox.fetchFeeEstimate(this.balanceRemaining);
+    const insufficientFunds = this.balanceRemaining > app.walletBalance.get('confirmed');
+    this.spendConfirmBox.setState({
+      show: true,
+      fetchFailed: true,
```
jjeffryes: Wouldn't this render twice, once here, then again when fetchFeeEstimate is called?

Could the state be set as an else after the insufficientfunds check instead?

rmisio: > Wouldn't this render twice, once here, then again when fetchFeeEstimate is called?

Yes, it would.

> Could the state be set as an else after the insufficientfunds check instead?

No, it wouldn't be that simple because you want to show the tooltip in either case. If you really wanted to avoid the double render you could pass in a `renderOnChange: false` option with the setState call and then in an else call render manually.

But... that's somewhat risky because it only works if `fetchFeeEstimate` will trigger a render, which it does now, but if something about that function is changed and there's a scenario where it wouldn't re-render, it would introduce a bug. And especially with that function being in a separate view, it's doubtful the dev changing that function would realize it's affecting this view in that way. 

I think this is a scenario where a very slight inefficiency makes for less bug prone and cleaner code. But, it's a very perceptive catch.

(FWIW - this is also another less than ideal aspect of us having to manage renders. More modern frameworks take care of that for you. You just concern yourself with setting state and it's up to framework and the thousands of eyes on it to concern themselves with rendering efficiency (queuing things up, batching things, etc...))

PR #969, opened 2017-09-29
Search suggestions

jjeffryes:
This implements a row of suggestions below the search bar if any are provided.

Currently, if there are more than one row worth of suggestions they will just be trimmed to one line. Adding in logic to detect if they go over one line and add a button to show the rest will need to come in a future PR (the current library we use for this, Trunk8, didn't work for this scenario when I tried it).

If the search provider doesn't provide any suggestions, this will insert the following list:
'books',
        'clothing',
        'electronics',
        'food',
        'games',
        'health',
        'movies',
        'music',
        'sports',
        'toys',

### DIFF
--- a/js/templates/search/Suggestions.html
+++ b/js/templates/search/Suggestions.html
@@ -1,0 +1,6 @@
+<% if (ob.suggestions && ob.suggestions.length) { %>
+  <button class="btnTxtOnly txUnb clrT2 suggestionTitle">Suggestions:</button>
+  <% ob.suggestions.forEach(suggestion => { %>
+    <button class="btnTxtOnly txUnb js-suggestion" data-suggestion="<%= suggestion %>"><%= suggestion%></button>
+  <% }); %>
+<% } %>
--- a/js/templates/search/Suggestions.html
+++ b/js/templates/search/Suggestions.html
@@ -1,5 +1,5 @@
 <% if (ob.suggestions && ob.suggestions.length) { %>
-  <span class="clrT2">Suggestions:</span>
+  <span class="clrT2"><%= ob.polyT('search.suggestions') %></span>
   <% ob.suggestions.forEach(suggestion => { %>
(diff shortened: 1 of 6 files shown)


js/templates/search/Suggestions.html
```diff
@@ -0,0 +1,6 @@
+<% if (ob.suggestions && ob.suggestions.length) { %>
+  <span class="clrT2">Suggestions:</span>
```
rmisio: Can we please translate?

js/views/search/Suggestions.js
```diff
@@ -0,0 +1,46 @@
...
+    const suggestion = $(e.target).data('suggestion');
+    this.trigger('clickSuggestion', { suggestion });
+  }
+
+  render() {
+    super.render();
+    const state = this.getState();
+    loadTemplate('search/Suggestions.html', t => {
+      this.$el.html(t({
+        ...this.options,
```
rmisio: Is it necessary to pass the options in here? Sometimes you want one or two to get in, but it seems safer and cleaner to explicitly pass in what you need instead of completely passing in the view options to the template.

js/views/search/Search.js
```diff
@@ -27,6 +28,21 @@ export default class extends baseVw {
 
     super(opts);
     this.options = opts;
+
+    this.defaultSuggestions = this.options.defaultSuggestions ||
+      [
+        'books',
```
rmisio: hopefully @tyler-smith could send back some more useful suggestions, because none of these result in any results.

I recommend this wonderful restaurant for dinner tonight -- the restaurant is only open for breakfast and lunch :(

rmisio: nevermind t-sizzle. I was on testnet... lol!

js/views/search/Suggestions.js
jjeffryes: Oops. Leftover from when I passed the suggestions in over options. Thanks!

morebrownies:
Love it! this helps with the usability a lot. good thinking @jjeffryes on the default list.

PR #981, opened 2017-10-04
Listing save confirm

### LINKED ISSUE #104: If there are unsaved changes, and I attempt to close the Create Listing modal, a warning should trigger asking, "You have unsaved changes. Are you sure you want to close?"
This will prevent accidental clicks that result in a lot of lost work

### LINKED ISSUE #835: After saving a listing, display a "view listing" link next to the save button.
We've heard from at least a handful of people that finding the listing they just saved isn't super intuitive. 

After the `Save` button is clicked, and the listing saves succesfully, display a `View listing` link next to the save button. We'll want to be sure to do this next to both `Save` buttons.

screenshot 2017-09-11 09 56 28

rmisio:
Implements a confirmation dialog when closing or navigating away from the listing modal while you have unsaved changes in the form.

Also implements a View Listing link into the Edit Listing modal. When creating a listing, it will appear after first saving the listing.

closes #835
closes #104

### DIFF
--- a/js/router.js
+++ b/js/router.js
@@ -3,8 +3,10 @@ import { ipcRenderer } from 'electron';
 import { Router } from 'backbone';
 import { getGuid, isMultihash } from './utils';
 import { getPageContainer } from './utils/selectors';
+import { isPromise } from './utils/object';
 import './lib/whenAll.jquery';
 import app from './app';
+import { getOpenModals } from './views/modals/BaseModal';
 import UserPage from './views/userPage/UserPage';
 import Search from './views/search/Search';
 import Transactions from './views/transactions/Transactions';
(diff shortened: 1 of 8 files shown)


js/router.js
```diff
@@ -149,10 +166,71 @@ export default class ObRouter extends Router {
...
-    app.loadingModal.open();
+  execute(callback, args, name, options = {}) {
+    if (this.closeUnconfirmedRollBack) {
+      this.closeUnconfirmedRollBack = false;
+      return false;
+    }
+
     this.navigate(this.standardizedRoute(), { replace: true });
 
+    // We'll iterate through any open modal which have a confirmClose method
```
rmisio: The router work is where the bulk of the complexities lies with this feature. It's one thing to put up a confirm when you click close, but ideally we also want it if the user navigates away from the modal as well. Additional complexity comes into play with the fact that you could have multiple modals open that need confirmation (e.g. you may have begun editing a listing and then decided to create a listing on top of the one you are editing).

js/templates/modals/editListing/editListing.html
```diff
@@ -42,7 +42,14 @@ <h2 class="h3 clrT js-listingHeading"><% print(ob.createMode ? ob.polyT('editLis
...
               <div class="col12">
-                <label for="editListingTitle" class="required"><%= ob.polyT('editListing.title') %></label>
+                <div class="flexRow">
+                  <label for="editListingTitle" class="required"><%= ob.polyT('editListing.title') %></label>
+                  <div class="flexExpand js-viewListingWrap <% if (ob.createMode) print('hide') %>">
+                    <div class="flexHRight">
+                      <a class="txU js-viewListing"><%= ob.polyT('editListing.viewListingLink') %></a>
+                    </div>
+                  </div>
+                </div>
```
jjeffryes: You can simplify this by putting the flexExpand on the title. 

```
<div class="flexRow">
  <label for="editListingTitle" class="required flexExpand"><%= ob.polyT('editListing.title') %></label>
  <div class=" js-viewListingWrap <% if (ob.createMode) print('hide') %>">
    <a class="txU js-viewListing"><%= ob.polyT('editListing.viewListingLink') %></a>
  </div>
</div>
```

PR #986, opened 2017-10-06
Multiple

rmisio:
closes #972 
closes #970 
closes #966 
closes #926

PR #991, opened 2017-10-09
Patch radio btns

jjeffryes:
I missed that #986 caused some issues with radio buttons outside of forms. This corrects the problem, and changes how the alignment works to be more robust. 

### Before
image

### After
image


jjeffryes:
Fixing some more missed spots.

morebrownies:
Radio buttons looking good to me in all places @jjeffryes

PR #1002, opened 2017-10-11
Create uk-UA.json

6025c7:
Initial translation


jjeffryes:
Does this translation exist in transifex? 

If so, please just translate it there. We pull all the translations from transifex every release, and replace any files in the repo with them.

6025c7:
Okay, thanks for a reply

jjeffryes:
I saw your join request on Transifex and approved it. Thank you for helping with translations.

6025c7:
My pleasure!

PR #1005, opened 2017-10-13
Lang data

rmisio:
- updates the settings > general lang field to interface with the local settings
- beefs up the languages data file so it has both a full language list as well as a separate list of languages that are translatable (i.e. decouples a full list of languages from what we happen to have translations for)


js/languages/en-US.json
```diff
@@ -2117,5 +2119,235 @@
...
+    "uk-UA": "Ukrainian (Ukraine)",
+    "ur": "Urdu",
+    "ur-PK": "Urdu (Islamic Republic of Pakistan)",
+    "uz": "Uzbek (Latin)",
+    "uz-UZ": "Uzbek (Cyrillic) (Uzbekistan)",
+    "vi": "Vietnamese",
+    "vi-VN": "Vietnamese (Viet Nam)",
+    "xh": "Xhosa",
+    "xh-ZA": "Xhosa (South Africa)",
+    "zh": "Chinese",
```
rmisio: @jjeffryes I switched things up a little bit from what we discussed. I know we mentioned we were only going to use the base lang and not include a region. But... I remember from my days of working as a localization engineer that there is a substantial difference between simplified (zh-CN) and traditional chinese (zh-TW) - they may even have slightly different alphabets. So, I don't think we could get away with just using the base language as it might be critical to know which "flavor" of that language you mean.

I ended up basing out list off of this one:
http://www.lingoes.net/en/translator/langcode.htm

js/models/LocalSettings.js
```diff
@@ -3,8 +3,25 @@ import LocalStorageSync from '../utils/backboneLocalStorage';
...
 import app from '../app';
 
+/**
+ * Will convert a transifex style language tag to the standard format JS understands as
+ * best as it can. For example, `ca_ES` will be returned as  `ca-ES` and `ca@valencia` will
+ * be returned as `ca`. It's still probably a good idea to wrap any JS that deals with these
+ * codes in a try/catch (e.g. localeCompare) since even a standardized transifex code may not
+ * map to something JS understands.
+ */
+export function standardizedTranslatedLang(lang) {
```
rmisio: This was a bit of a rabbit hole... but since transifex language codes don't necessarily map to ones JS understands, it's important that when you want a language code for JS (e.g. using localeCompare() or Intl.NumberFormat()), rather than obtain it directly from localSettings (`app.localSettings.get('language')`), you retrieve it via `localSettings.standardizedTranslatedLang()`.

That method will try and convert the transifex style to JS style. Keep in mind, it's not a one to one mapping. There are some transifex langs that don't map to ones JS understands (e.g. `tzm`). That's why it's also important that you be cognizant of that fact when using the language in any JS functions (e.g. localeCompare() will throw an exception if it gets an unrecognized lang code, so I needed to explicitly default it to en-US in that case. Intl.NumberFormat() on the other hand was more forgiving and it looks like it defaulted it on its own).

js/models/LocalSettings.js
```diff
@@ -15,12 +32,22 @@ export default class extends Model {
   }
 
   defaults() {
+    let langDataObj;
+
+    try {
+      langDataObj = getTranslationLangByCode(navigator.language.replace('-', '_'));
```
rmisio: I am defaulting this to detect the user's language. If it maps to one that we have a translation for, then that lang will be used. Otherwise, we fall back to us english.

js/models/LocalSettings.js
```diff
@@ -15,12 +32,22 @@ export default class extends Model {
   }
 
   defaults() {
+    let langDataObj;
```
jjeffryes: Since this is a string, is it confusing to name it langDataObj?

jjeffryes:
@rmisio can you fix the merge conflicts?

js/models/LocalSettings.js
rmisio: I don't think it's a string. I think it returns the full data object, e.g.:

```
  {
    name: 'Français (French, Canada)',
    code: 'fr_CA',
  },
```

Are you seeing something different?

js/data/languages.js
```diff
@@ -1,9 +1,1049 @@
...
+  app.localSettings.get('language') || 'en-US', sort = true) {
+  if (!lang) {
+    throw new Error('Please provide the language the translated languages' +
+      ' should be returned in.');
+  }
+
+  let translated = languages.map(language => {
+    let translatedName;
+    if (!(app && app.polyglot)) {
+      console.warn('Unable to translate the name because the polyglot object is not available.');
```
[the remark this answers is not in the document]
rmisio: I'm not sure. It's really not a breaking change. It just means that we can't translate the lang name. I was thinking it's better without an exception which could cause weirdness in the app. At least this way, they would get the untranslated language back and the app would still be usable.

It's really a developer error if this is called before polyglot is available on the app object.

js/models/LocalSettings.js
rmisio: oh ok, i see the confusion. one sec.

rmisio: ok, I needed to use the code from that object rather than the object itself. Good catch that something funky was up.

PR #1018, opened 2017-10-16
Plethora

### LINKED ISSUE #994: UX: Show 'Free' instead of '$0.00' under Shipping
If an item does not charge for shipping, they have to input $0.00 as the price. This price is also shown to the user, as is a bit awkward to look at.

Instead, it may be nicer if it said 'Free', rather than a price of zero dollars and zero cents.

### LINKED ISSUE #1013: Missing red * for required field in Photos
image
Missing red * for required field in Photos

### LINKED ISSUE #1014: Two buttons with misaligned long translations
This is seen on long translations (v2.0.14)
"Receive Money" button on wallet:
image

"Pay with wallet" button:
image

### LINKED ISSUE #1016: Can't translate some UI elements
v.2.0.14

1)  Send button on order chat:
image

2) Edit/Clone/Delete listing buttons on Search Results and Store page:

image

image

3) Home, Store, Following, Followers titles and Follow/Unfollow button

image

### LINKED ISSUE #1017: Missing singular/plural text on listing Photos
v2.0.14

This should be "View 1 Photo" / "View 2 Photos"
image

rmisio:
closes #1014 
closes #1017
 
closes #1016 
---> I think this one also needs the latest `en-US` to be published. A few of the pieces called out as untranslatable did already have translations.

closes #1013 
closes #994


styles/components/_buttons.scss
```diff
@@ -19,6 +19,7 @@
   cursor: pointer;
   font-weight: 700;
   transition: border 0.15s ease;
+  text-align: center;
```
rmisio: We need to be on the look-out in case this one has some side-effects. I didn't see any, but I only did a cursory look through.

jjeffryes: Is this the intended effect?

image

If we change height to min-height, we'd get this

image

But that would be more disruptive to the layout in some places, because the button would push other elements out of the way.

@morebrownies I believe you asked for the text to truncate. That's not ideal for translations. Do you still want that, or would one of the above be ok?

PR #1029, opened 2017-10-18
Report button

jjeffryes:
This adds a report button to any search provider that includes a report link in their data.

The button will show a modal with 3 pre-populated options, and an other option with a text field. When submitted, the report will be sent to the provider's report end-point.

After submitting, a post-submit message will be shown explaining how reporting works so user's won't be confused by the reported listings not disappearing immediately.

In the future we can expand this to block reported nodes and grey out or otherwise obscure search results from them. That should be done as part of the block nodes work. 

This completes the report button portion of #569 

Note: this PR also fixes some layout issues with radio buttons, which requires any radio buttons that are horizontal to be inside a btnStrip container.

### DIFF
--- a/js/templates/listingCard.html
+++ b/js/templates/listingCard.html
@@ -126,4 +126,3 @@
     <%= ob.polyT('listingCard.deleted') %>
   </div>
 </div>
-<div class="js-report"></div>
--- a/js/templates/listingCard.html
+++ b/js/templates/listingCard.html
@@ -24,6 +24,8 @@
             </div>
           </div>
         </div>
+      <% } else if (ob.showReportBtn) { %>
+        <button class="btn js-report">report this</button>
       <% } %>
   </div>
   <div class="pad clrBr borderTop">
@@ -69,7 +71,7 @@
         <div class="listingImage posR js-listViewListingImage">
         <div class="js-listingImageLoadSpinner center">
           <%= ob.spinner({ className: 'spinnerSm' }) %>
-        </div>          
+        </div>
         </div>
       </div>
       <div class="flexExpand">
@@ -99,7 +101,7 @@
       <div class="overlayPanelInner"></div>
       <div class="flexHCent gutterHSm">
         <button class='iconBtnSm clrP clrBr toolTipNoWrap toolTipTop js-edit' data-tip="<%= ob.polyT('listingCard.editListingTooltip') %>"><span class="ion-edit"></span></button>
-        <button class='iconBtnSm clrP clrBr toolTipNoWrap toolTipTop js-clone' data-tip="<%= ob.polyT('listingCard.cloneListingTooltip') %>"><span class="ion-ios-copy"></span></button>          
+        <button class='iconBtnSm clrP clrBr toolTipNoWrap toolTipTop js-clone' data-tip="<%= ob.polyT('listingCard.cloneListingTooltip') %>"><span class="ion-ios-copy"></span></button>
         <div class="posR">
           <button class='iconBtnSm clrP clrBr toolTipNoWrap toolTipTop js-delete' data-tip="<%= ob.polyT('listingCard.deleteListingTooltip') %>"><span class="ion-trash-b"></span></button>
           <div class="js-deleteConfirmedBox confirmBox deleteConfirm tx5 arrowBoxBottom clrBr clrP clrT hide">
@@ -111,7 +113,7 @@
               <a class="btn clrBAttGrad clrBrDec1 clrTOnEmph js-deleteConfirmed"><%= ob.polyT('listingCard.confirmDelete.btnConfirm') %></a>
             </div>
           </div>
-        </div>          
+        </div>
       </div>
     </div>
   <% } %>
--- a/js/templates/listingCard.html
+++ b/js/templates/listingCard.html
@@ -24,8 +24,8 @@
             </div>
           </div>
         </div>
-      <% } else if (ob.showReportBtn) { %>
-        <button class="btn js-report">report this</button>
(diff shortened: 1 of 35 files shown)


js/templates/modals/settings/advanced/advanced.html
```diff
@@ -17,25 +17,27 @@ <h2 class="h3 clrT"><%= ob.polyT('settings.advancedTab.sectionName') %></h2>
...
-                <div class="btnRadio clrBr">
-                  <input type="radio"
-                         data-persistence-location="local"
-                         name="showAdvancedVisualEffects"
-                         value="false"
-                         id="advancedTabVisualEffectsOff"
-                         data-var-type="boolean"
-                  <% if(!ob.showAdvancedVisualEffects) { %>checked<% } %>>
-                  <label for="advancedTabVisualEffectsOff"><%= ob.polyT('settings.off') %></label>
+                <div class="btnStrip">
```
jjeffryes: The buttons here were put in a button strip container, radio buttons won't line up horizontally on their own any more since that was causing some issues.

js/views/search/AddProvider.js
```diff
@@ -89,7 +89,7 @@ export default class extends BaseView {
 
   render() {
     super.render();
-    loadTemplate('search/AddProvider.html', t => {
+    loadTemplate('search/addProvider.html', t => {
```
jjeffryes: I noticed the capitalization was off here and fixed it.

js/views/search/Results.js
```diff
@@ -140,7 +142,7 @@ export default class extends baseVw {
 
 
   render() {
-    loadTemplate('search/Results.html', (t) => {
+    loadTemplate('search/results.html', (t) => {
```
jjeffryes: The capitalization was off here too.

js/views/userPage/CategoryFilter.js
```diff
@@ -23,7 +23,7 @@ export default class extends BaseVw {
   }
 
   className() {
-    return 'clrP clrBr padMd clrT contentBox clrSh2 categoryFilter';
+    return 'clrP clrBr padMd clrT contentBox clrSh2 form veryCompact categoryFilter';
```
jjeffryes: Adding the form and veryCompact classes allows the store category radio buttons to use the normal radio button styling.

styles/components/_form.scss
```diff
@@ -1,4 +1,5 @@
-form {
+form,
+.form {
```
jjeffryes: There are various places where inputs need form styling, but aren't in forms, the .form class handles that situation.

styles/components/_form.scss
```diff
@@ -9,6 +10,14 @@ form {
...
     align-items: center;
   }
+
+  &.compact label {
+    line-height: $formElementHeight * 0.75;
+  }
+
+  &.veryCompact label {
+    line-height: $formElementHeight * 0.5;
+  }
```
jjeffryes: The compact and veryCompact classes are for input elements that should have form styling, but not as much vertical space.

styles/modules/_search.scss
```diff
@@ -179,13 +179,6 @@
       padding: 0.3em 0px;
     }
 
-    input[type="radio"] + label:before {
```
jjeffryes: This earlier fix isn't needed any more now that the radio button styling is cleaned up and unified.

rmisio:
Do we need a tooltip on the report button? It would be nice if the tooltip changed after the listing was reported (e.g. `Report listing` => `You have reported this listings`).

rmisio:
The reported button still gets a border on hover which is a little confusing because it seems like it's clickable:

image

morebrownies:
screenshot 2017-10-18 10 24 06

The report button should only be visible when hovering a listing card. No need to show it all the time.

morebrownies:
> Do we need a tooltip on the report button? It would be nice if the tooltip changed after the listing was reported (e.g. Report listing => You have reported this listing).

hmm I'm not seeing the tooltip on my end. Suggestion, the states of the tooltip could be: `Report listing` and `Reported`

morebrownies:
The `title` of the **Report Listing** overlay should be centered and smaller. See the `title` of the **Create Listing** overlay as an example

screenshot 2017-10-18 10 27 31

screenshot 2017-10-18 10 27 21

morebrownies:
Clicking `Other` would ideally autofocus the text input next to it

screenshot 2017-10-18 10 31 12

js/templates/listingCard.html
```diff
@@ -124,3 +126,4 @@
     <%= ob.polyT('listingCard.deleted') %>
   </div>
 </div>
+<div class="js-report"></div>
```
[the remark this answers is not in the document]
jjeffryes: Not any more. I forgot to remove it after refactoring.

jjeffryes:
@rmisio @morebrownies this is updated will all the changes suggested.

morebrownies:
Looking good. Maybe this is just me? But i'd still prefer shorter tooltips:

`Report this listing` => `Report listing`
`This listing was reported` => `Reported`

Will be easier to translate and will be shorter across languages

jjeffryes:
> Will be easier to translate and will be shorter across languages

I actually made them longer so they'd be easier to translate. I get a lot of issues on Transifex that short phrases are ambiguous and confusing.

In this instance, just "report" can mean "show me a report about this listing" as well as "take the action of creating and sending a report about this listing".

Less ambiguity is better in this case, at least in my opinion.

morebrownies:
Ok good to know. thanks

PR #1037, opened 2017-10-20
use translated language for stats

jjeffryes:
This fixes the wallet stats, which was still referencing the old language settings.


jjeffryes:
@morebrownies please confirm this fixes the bug for you. It does for me, but I want to make sure it's the same thing you were seeing.

morebrownies:
Yeah that fixed it 👍

PR #1038, opened 2017-10-20
Update the readme file

jjeffryes:
Fleshes out the readme file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ For full installable versions of the OpenBazaar app, with the server and client 
 To create a local development copy of the reference client, clone the client repository into a directory of your choice:
 - `git clone https://github.com/OpenBazaar/openbazaar-desktop`
 
-Make sure you have the latest versions of Node.js and NPM installed.
+Make sure you have Node.js 8.9.2 and NPM 5.5.1 installed.
 
 This client uses Babel to compile [ES6 JavaScript](https://github.com/lukehoban/es6features). You should be familiar with ES6 before modifying its code.
 
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ For full installable versions of the OpenBazaar app, with the server and client 
 To create a local development copy of the reference client, clone the client repository into a directory of your choice:
 - `git clone https://github.com/OpenBazaar/openbazaar-desktop`
 
-Make sure you have Node.js 8.9.2 and NPM 5.5.1 installed.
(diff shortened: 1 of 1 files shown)


jjeffryes:
This is still waiting for a review from @SamPatt and @hoffmabc

jjeffryes:
@hoffmabc @SamPatt any thoughts on this?

README.md
```diff
@@ -1,18 +1,26 @@
...
 [![Build Status](https://travis-ci.org/OpenBazaar/openbazaar-desktop.svg)](https://travis-ci.org/OpenBazaar/openbazaar-desktop)
 
-Installation
-------------
+## Getting Started
+
+To create a local development copy of the reference client, clone the client repository into a directory of your choice:
+- `git clone https://github.com/OpenBazaar/openbazaar-desktop`
+
+Make sure you have the latest versions of Node.js and NPM installed.
```
hoffmabc: Do we want to give a specific version? Do we know if it will just straight up work with latest node and npm?

jjeffryes: That's a good point, we should specify the latest versions we've tested.

rmisio:
@jjeffryes who is this one waiting on?

jjeffryes:
@rmisio I need to go back and make Brian's suggested change.

jjeffryes:
This has been updated with specific Node and NPM versions. Once #1137 is merged we can merge this.

README.md
```diff
@@ -11,7 +11,7 @@ For full installable versions of the OpenBazaar app, with the server and client
 To create a local development copy of the reference client, clone the client repository into a directory of your choice:
 - `git clone https://github.com/OpenBazaar/openbazaar-desktop`
 
-Make sure you have the latest versions of Node.js and NPM installed.
+Make sure you have Node.js 8.9.2 and NPM 5.5.1 installed.
```
rmisio: shouldn't this have "or greater"?

jjeffryes: My thought is those are the versions I've tested up to, and even if it's not likely later versions would break anything, we can't be sure.

rmisio: hmm... I'm not sure I've ever seen anything like this (although I certainly haven't seen everything :)).

The thing is, for example, if the latest version of node becomes 8.9.9, I would think the user should just install latest and not go through the pain of having to install an old version.

jjeffryes: We could say make sure you have the latest versions of node 8 and NPM 5, and then say it's been tested on 8.9.2 and 5.5.1 specifically. 

Though we should make a habit of updating both of them and checking to make sure nothing breaks periodically.

PR #1039, opened 2017-10-20
Fix clamp peek through and other issues

### LINKED ISSUE #1033: ^~´` of listing title is being displayed on top of second line
win7 64 bits v2.0.14
If the second line of a wrapped listing title has a character with ^~` or ´, this is not cut:
Example: test listing ÉÉÉÃÃÃÃÕÕÕÁÁÁÁÀÀÀÀÀÀÀÊÊÊÊÊÉÉÉÃÃÃÃÕÕÕÁÁÁÁÀÀÀÀÀÀÀÊÊÊÊÊÉÉÉÃÃÃ

image

jjeffryes:
This fixes an issue where the next line of characters could peek through the bottom of clamped text.

It also fixes a problem where very long notifications would push the thumbnail to the side in the notifications list, and a problem where very long listing titles were not being trimmed to avoid running beneath the avatar.

Closes #1033

### DIFF
--- a/js/templates/listingCard.html
+++ b/js/templates/listingCard.html
@@ -29,9 +29,9 @@
       <% } %>
   </div>
   <div class="pad clrBr borderTop">
-    <div class="rowSm <% if(ob.vendor) print('trimWidth') %>">
+    <div class="rowSm">
       <% const tooltipClass = ob.title.length > 60 ? 'toolTip' : 'toolTipNoWrap' %>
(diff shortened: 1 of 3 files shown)


jjeffryes:
I tried to go through the entire app and check all the text with clamp classes, but another look would be a good idea. There's a chance there's text that has a clamp class and a custom line height that isn't accounted for. If that happens, the text would have the bottom cut off.

PR #1042, opened 2017-10-20
Numeric error msg tweak

rmisio:
Updating the error messages when a field is expecting a number and allows decimal numbers to include the expected decimal format.

### DIFF
--- a/js/data/languages.js
+++ b/js/data/languages.js
@@ -976,7 +976,7 @@ function getIndexedLanguages() {
 }
 
 export function getLangByCode(code, translate = true, lang = app && app.localSettings &&
-  app.localSettings.get('language') || 'en-US') {
+  app.localSettings.standardizedTranslatedLang() || 'en-US') {
   if (!code) {
     throw new Error('Please provide a language code.');
   }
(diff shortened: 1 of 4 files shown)


js/data/languages.js
```diff
@@ -1029,7 +1029,26 @@ function getTranslatedLangs(lang = app && app.localSettings &&
   });
 
   if (sort) {
-    translated = translated.sort((a, b) => a.name.localeCompare(b.name, lang));
+    translated = translated.sort((a, b) => {
```
rmisio: This was missed in a previous PR. Without this one of the settings tabs was bombing since JS does not recognize the transifex lang code - it needs to be standardized.

js/views/modals/editListing/VariantInventoryItem.js
```diff
@@ -82,11 +82,15 @@ export default class extends BaseView {
   }
 
   calculateTotalPrice(surcharge) {
-    const updatedSurcharge = (typeof surcharge === 'number' && !isNaN(surcharge)) ?
-      surcharge : 0;
+    const listingPrice = this.options.getPrice();
```
rmisio: This is also an additional fix where the variant inventory table was bombing if the listing price was not a number. Now it just won't display the total price if it's unable to calculate it (if either the listing price or surcharge are not numbers). If the user saves the listing the offending field will get an error instructing the user to provide a proper number.

PR #1044, opened 2017-10-22
Fix Language Code in getTranslatedLangs

jjeffryes:
In languages, getTranslatedLangs was using localSettings.get('language", which returned the Transifex style language string.  That threw a `Invalid language tag: en_US` error.

This updates the code to use the standardized value instead.


rmisio:
This is handled in #1042

PR #1048, opened 2017-10-23
Position PageContent

### LINKED ISSUE #995: UI: Follow store loading spinner shows above top navigation bar
The loading spinner that shows after following/unfollowing a store is placed above the navigation bar layer, and thus overlaps it upon scrolling down the page.

This does not happen with the regular, non-loading 'Follow'/'Unfollow' button.

OpenBazaar2_2.0.13, Debian
Reproducible: Always

Steps to reproduce:
1. Wait for store page to load
2. Follow/Unfollow store
3. Scroll down and watch loading spinner incorrectly overlay on top of other UI elements

jjeffryes:
Add a z-index to the pageContent div so it is under the page tab.
Closes #995

### DIFF
--- a/styles/_base.scss
+++ b/styles/_base.scss
@@ -208,6 +208,7 @@ svg {
         margin: 0 calc((100% - #{$pageWidth + $padLg + $padLg}) / 2) 0 auto;
         padding: 0 $padLg $bar $padLg;
         position: relative;
+        z-index: 0; // position below the tab tab bar for scrolling
         @include chatOpeningTransition(margin);
 
         @include alignToChat {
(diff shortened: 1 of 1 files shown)


styles/_base.scss
```diff
@@ -208,6 +208,7 @@ svg {
         margin: 0 calc((100% - #{$pageWidth + $padLg + $padLg}) / 2) 0 auto;
         padding: 0 $padLg $bar $padLg;
         position: relative;
+        z-index: 0; // position below the tab tab bar for scrolling
```
rmisio: Did you mean `top tab bar`?

jjeffryes:
@rmisio comment typo fixed.

PR #1050, opened 2017-10-23
Change search placeholder

### LINKED ISSUE #937: Sales page search field has wrong prefilled text
On Sales page, user searches for "Buyer".
On Purchases page, user searches for "Vendor".
On Cases page, user searches for "Vendor/Buyer"? (I have no case, need to check this).

salesvendor

jjeffryes:
Changes the transaction search placeholder text so it applies to all 3 transaction tabs. 

Closes #937

### DIFF
--- a/js/languages/en_US.json
+++ b/js/languages/en_US.json
@@ -1279,7 +1279,7 @@
       "dateAsc": "Newest",
       "dateDesc": "Oldest"
     },
-    "placeholderSearch": "Search by Order #, Title or Vendor",
+    "placeholderSearch": "Search by Order #, Title or Person ID",
     "resetFilters": "Reset filters",
     "transactionsTable": {
       "headers": {
--- a/js/languages/en_US.json
+++ b/js/languages/en_US.json
@@ -1279,7 +1279,9 @@
       "dateAsc": "Newest",
       "dateDesc": "Oldest"
     },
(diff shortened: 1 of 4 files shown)


rmisio:
@morebrownies Are you find with using the generic `Person ID` here?

I know you had mentioned you wanted this:

```
The placeholder is correct when you're on the Purchases tab, but incorrect when on the Sales tab.

We'll probably want to change the placeholder text to Search by Order #, Title or Buyer when on the Sales tab.
```

jjeffryes:
It's different for each tab now.

jjeffryes:
@morebrownies this was updated to the less lazy solution. :)

rmisio:
> this was updated to the less lazy solution. :)

Oh... no `Person ID` anymore? :(

js/languages/en_US.json
```diff
@@ -1279,7 +1279,9 @@
       "dateAsc": "Newest",
       "dateDesc": "Oldest"
     },
-    "placeholderSearch": "Search by Order #, Title or Vendor",
+    "placeholderSearchsales": "Search by Order #, Title or Buyer",
```
rmisio: It's minor, but can we use proper camel-case here and then you could just use the `capitalize` utility function in `tab.html`?

jjeffryes: Will do.

PR #1068, opened 2017-10-26
Tweak emoji size and alignment

### LINKED ISSUE #1067: Text Getting Cut Off in Chat
image

OSX 2.0.15-rc1

The location is cut off on the bottom.

jjeffryes:
- only increase the size of emojis in the conversation, so the location icon doesn't throw off the line height
- align the emojis to the bottom so they are more centered.
- closes #1067

### DIFF
--- a/styles/modules/_chat.scss
+++ b/styles/modules/_chat.scss
@@ -394,8 +394,9 @@
   }
 }
 
-.chatConversation .emoji, #chatContainer .emoji {
-  width: 16px;
-  height: 16px;
+.chatConversation .emoji, #chatContainer .chatConvoMessages .emoji {
+  width: 1.3em;
+  height: 1.3em;
+  vertical-align: bottom;
 }
 


styles/modules/_chat.scss
```diff
@@ -394,8 +394,9 @@
   }
 }
 
-.chatConversation .emoji, #chatContainer .emoji {
-  width: 16px;
-  height: 16px;
+.chatConversation .emoji, #chatContainer .chatConvoMessages .emoji {
+  width: 1.3em;
+  height: 1.3em;
+  vertical-align: bottom;
```
rmisio: Why is the bottom vert align necessary here but not in all the other places we use emojis?

rmisio:
This is kind of a tangent, but I think at some point we should get rid of the right margin that's globally on all emojis. Most of our emojis are coming from user text and people will put a space character next to an emoji if their intention is for there to be space there. Us hard-coding a right margin on the emoji just results in what appears to be a double space to the right of it.

styles/modules/_chat.scss
jjeffryes: Because most of the other places they're center aligned using flex, here they're inline and without the bottom alignment they're really off center. 

I tried using middle instead, and it's still off due to the size of the emoji. This was the best compromise.

jjeffryes:
> I think at some point we should get rid of the right margin that's globally on all emojis

That's a good idea, I think that was originally in place because they were being injected without a space and it caused problems.

We'd have to check everything that uses emojis carefully and address those places when it gets removed.

PR #1079, opened 2017-10-31
Selectize

### LINKED ISSUE #978: Replacing tags for a listing
Brief Description: 
Replacing tags for listings does not work as expected. The replaced tag reappears.

Operating System (OS and version):  Mint 18
OpenBazaar version (shown on `About OpenBazaar` page in menu):  client 2.0.13, server 0.9.2

Reproducible (Always / Almost Always / Sometimes / Rarely / Couldn't Reproduce): 
Always

Steps to reproduce:
1. add a tag
2. remove the tag
3. add another tag
(check the video below)

Observed Behavior:
the removed tag reappears

Expected Behavior:
the removed tag does not reappear

**Demo video: https://streamable.com/f58sb**

rmisio:
This PR switches tagging fields across the app to use the selectize library instead of select2. At some point if we want, we could also do this for drop downs.

closes #978


js/templates/modals/editListing/variant.html
```diff
@@ -19,18 +19,7 @@
...
-        <select multiple name="variants" class="clrBr clrP select2Tags" style="width: 100%">
-          <% ob.variants.forEach(variant =>
-              print(`<option value="${variant._clientID}<===>${variant.name}" selected>${variant.name}</option>`)) %>
-        </select>
-        <div class="placeholder tagsPlaceholder js-variantChoicesPlaceholder"><%= ob.polyT('editListing.variants.choicesPlaceholder') %></div>
-        <!-- needed to properly hide the select2 dropdown which we don't want on a tagging field -->
-        <div class="tagsDropdown js-dropDownContainer"></div>
-      </div>
-    </div>
+    <select multiple name="variants" class="clrBr clrP clrSh2 hideDropDown flexExpand" placeholder="<%= ob.polyT('editListing.variants.choicesPlaceholder') %>"></select>
```
rmisio: Note the `hideDropDown` class. One of the quirks with selectize is if you delete an existing tag and then start typing it in again, it will appear in a drop down of available options. On most of the tagging fields we don't want such behavior - we essentially want a text field that contains tags. Use the `hideDropDown` class on these fields.

It will style the drop down to be hidden. It is also used by the hacked selectize lib to prevent certain related behavior (e.g. an enter press auto adding an option from the hidden menu).

js/utils/selectize.js
```diff
@@ -0,0 +1,15 @@
+/*
+  A wrapper for the selectize jquery element. A tweaked version of it is in the lib folder.
+*/
+
+export const tagsDelimiter = '<=!=>';
```
rmisio: When using selectize for tagging, you can trigger it from an input[type=text] element and, by default, you populate any existing tags inside a value attribute, e.g. `value="tag1,tag2,tag3"`. I found the comma there could be problematic as a separator because certain tags like categories may potentially contain commas. Our client doesn't allow them, but since the server allows them, they might get in there.

So, I've updating the selectize lib to use `<=!=>` as the delimiter, instead of a comma, e.g. `value="tag1<=!=>tag2<=!=>tag3<=!=>`.

When you're dealing with the data, you'll often need to Array.split() and String.join(). Make sure you import this delimiter form here and pass that in to each of those methods.

rmisio: FWIW, I've subsequently found there's no reason you HAVE to use an input[type=text]. You are free to use a \<select\> which allows you to embed the options in markup as option elements, or pass them in via code when initializing the jQuery plugin. <--- might be a slightly easier workflow, but definitely has a more robust api if you have to do something fancy.

rmisio:
@jjeffryes I had to tweak the selectize source to account for a couple bugs / inadequacies. If, at some point, you're trying to implement something using the lib and you run into problems, might not be a bad idea to ping me. I may be able to point you in the right direction.

js/utils/selectize.js
```diff
@@ -0,0 +1,20 @@
+/*
+  A wrapper for the selectize jquery element. A tweaked version of it is in the lib folder.
+*/
+
+// If triggering selectize from an input[type=text] the default delimiter has been changed
+// in the lib to use the following instead of a comma, since some of our tags may have
+// commas in them. If manipulating the data (e.g. Array.split(), String.join()), you'll
+// likely want to pass this variable in.
+export const tagsDelimiter = '<=!=>';
```
rmisio: When using selectize for tagging, you can trigger it from an input[type=text] element and, by default, you populate any existing tags inside a value attribute, e.g. value="tag1,tag2,tag3". I found the comma there could be problematic as a separator because certain tags like categories may potentially contain commas. Our client doesn't allow them, but since the server allows them, they might get in there.

So, I've updating the selectize lib to use <=!=> as the delimiter, instead of a comma, e.g. value="tag1<=!=>tag2<=!=>tag3<=!=>.

When you're dealing with the data, you'll often need to Array.split() and String.join(). Make sure you import this delimiter form here and pass that in to each of those methods.

rmisio: FWIW, I've subsequently found there's no reason you HAVE to use an input[type=text]. You are free to use a <select> (potentially with a hideDropDown class) which allows you to embed the options in markup as option elements, or pass them in via code when initializing the jQuery plugin. <--- might be a slightly easier workflow, but definitely has a more robust api if you have to do something fancy.

jjeffryes:
I noticed that after you add a country, you can't add any more until you click out of the input and back in unless you delete an existing country. Is that fixable?

styles/components/_selectize.scss
```diff
@@ -0,0 +1,48 @@
...
+      font-size: inherit;
+      padding-top: 7px;
+      min-height: $formElementHeight;
+
+      input {
+        font-size: inherit;
+      }
+
+      &:not(.has-items) input {
+        top: 3px;
```
jjeffryes: This should have position: relative. Otherwise the top property doesn't get applied until after the input is in focus, which applies an inline position:relative, and that causes the placeholder to jump down.

jjeffryes:
It's also a little weird that deleting a tag opens the dropdown options, but that's not a deal breaker.

styles/components/_selectize.scss
rmisio: fixed

rmisio:
> I noticed that after you add a country, you can't add any more until you click out of the input and back in unless you delete an existing country. Is that fixable?

Yes, there is a `closeAfterSelect` field I made to false on certain fields. Should work as you expect now.

rmisio:
> It's also a little weird that deleting a tag opens the dropdown options, but that's not a deal breaker.

That's a known issue. There are a couple proposed hacks on the interwebs with work-arounds, but supposedly the hacks to not solve every case.

The ideal would be to hack the source to not do this, but that could easily escalate into a half day or day long thing and out of some self love, I will not put myself through that, at least at this time :)

PR #1081, opened 2017-10-31
Various

### LINKED ISSUE #1009: When Saving Moderator Settings Fields Required Even if Not Moderator
You are prevented from saving Moderator settings if you don't fill out the fields even if moderation is set to Off. It should only validate and require the moderator fields if you are choosing to be a moderator.

### LINKED ISSUE #1025: Missing translations
NSFW (Edit listing):
image

Delete button (Addresses):
image

Cancel and Add (Add search provider dialog):
image

Notifications title:
image

Load more button (connected peers):
image

Inventory control:
image

Primary photo tag:
image

Moderation title on configurations:
image

rmisio:
closes #1025 
closes #1009

### DIFF
--- a/js/templates/modals/editListing/uploadPhoto.html
+++ b/js/templates/modals/editListing/uploadPhoto.html
@@ -1,6 +1,6 @@
 <li style="background-image: url('<%= ob.getServerUrl(`ob/images/${ob.small}`) %>')" class="tile">
   <div class="primaryIndicatorWrap">
-    <div class="clrE1 clrTOnEmph phraseBox padTn">PRIMARY PHOTO</div>
+    <div class="clrE1 clrTOnEmph phraseBox padTn"><%= ob.polyT('primaryPhoto').toUpperCase() %></div>
   </div>
   <a class="closeIcon tx2 <%= ob.closeIconClass || '' %>">
     <span class="ion-ios-close-empty clrBr clrP clrT"></span>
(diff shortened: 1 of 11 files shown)


js/templates/modals/editListing/uploadPhoto.html
```diff
@@ -1,6 +1,6 @@
 <li style="background-image: url('<%= ob.getServerUrl(`ob/images/${ob.small}`) %>')" class="tile">
   <div class="primaryIndicatorWrap">
-    <div class="clrE1 clrTOnEmph phraseBox padTn">PRIMARY PHOTO</div>
+    <div class="clrE1 clrTOnEmph phraseBox padTn"><%= ob.polyT('primaryPhoto').toUpperCase() %></div>
```
jjeffryes: I think we should leave the uppercase in the translation instead of forcing it here. Uppercase letters may mean something different in other languages.

js/templates/modals/editListing/uploadPhoto.html
```diff
@@ -1,6 +1,6 @@
 <li style="background-image: url('<%= ob.getServerUrl(`ob/images/${ob.small}`) %>')" class="tile">
   <div class="primaryIndicatorWrap">
-    <div class="clrE1 clrTOnEmph phraseBox padTn">PRIMARY PHOTO</div>
+    <div class="clrE1 clrTOnEmph phraseBox padTn"><%= ob.polyT('primaryPhoto') %></div>
```
jjeffryes: This should be editListing.primaryPhoto

PR #1088, opened 2017-11-01
Tweak selectize

rmisio:
Reverting back selectize so the menu closes after selecting an option and fixing the issue that required you to click away and then back into the control to re-show the menu.


js/lib/selectize.js
```diff
@@ -851,7 +851,18 @@ import { _tagsDelimiter } from '../utils/selectize';
...
+      // If the input has focus, we'll let a click of it open the menu, which
+      // might be closed. This prevents the user from having to click out of the
+      // input and then click back in, just to show the menu.
+      if (self.isFocused && !self.isDisabled) {
+        if (!self.$activeItems.length) {
+          self.showInput();
+          self.setActiveItem(null);
+          self.refreshOptions(!!self.settings.openOnFocus);
+        }
+      }
```
rmisio: If we ever use selectize for regular drop-downs, we'll want to check this plays nice with that (I think it will, but something to keep in mind in case we see some weirdness).

PR #1102, opened 2017-11-09
Offline Warning [WIP]

jjeffryes:
This PR adds a check to see if a vendor is online, and if so, shows a warning that the transaction may fail if the fee reduces the vendor's payout to zero or less.

The purchase button is disabled until the online check is completed.

Vendor nodes that cannot accept incoming messages will show as offline, which is correct for this usage, as the transaction will be an offline transaction.

image

image

image

image


hoffmabc:
@TheButterZone I think this is a separate issue altogether so we should open another one on GitHub to discuss this idea.

@jjeffryes I think this looks fine and it works as described when I was testing it.
image

This is a LGTM.

jjeffryes:
@hoffmabc yes, still WIP, though it shouldn't take much to finish it.

hoffmabc:
ok ACK on concept and looks good so far. 👍

jjeffryes:
I'll close this PR, it's served it's purpose, which was to test whether we could pull the online status and calculate the effects on the fee in a reasonable way.

hoffmabc:
wait why is this being closed?

jjeffryes:
@hoffmabc pending discussion of a better solution. If we decide something like this is the way to go, we can always reuse some of the code.

hoffmabc:
Ok so maybe we need a Proposal tag or something. I was under the impression that everyone was kind of working together on this.

jjeffryes:
Generally we use WIP for things that aren't done or won't be merged that we want early feedback on.

rmisio:
My main concern with this approach is that it is a problem that only affects a small subset of transactions (offline low priced items), yet we're inundating all purchases (perhaps the most important user experience in the app) with the "overhead" of this. By overhead I mean a.) a whole lot of extra cognitive load on the user b.) a purchase button that will be disabled for 30 seconds + when it may turn out the vendor is online or the order is for $5000.

==================

My understanding of the problem is when an offline order is made, the funds go into a 1 of 2 multi-sig which means that when the vendor accepts the order, there is an extra fee that will be taken out of the order total, which reduces the amount of money the vendor will get. There are two main problems this causes:

a.) The vendor is getting a lot less money than they thought they'd get. For example, they sell an item for 4$. If the fees at the time they accept the order are $3, they will only get $1.

I think this is something that needs to be solved with vendor education. We already have documents that explain this issue... perhaps we need to give some examples of different scenarios in those docs? Even if a vendor ignores reading the docs, as soon as they receive their first offline payment, it will be clear to them they are receiving less than they thought they were and at that point they could educate themseleves, be it by reading the docs or reaching out to support. Since this really only happens on lower priced items, it's not as if they'll be out a huge sum of money. Relatively speaking, it will be an inexpensive lesson for them.

This is really on the vendor to know that if they are selling low cost products they either need to a.) be online all the time b.) use an alternate currency once we have that available c.) if they're willing to wait a while for payments to confirm, they could set the fees to a low amount in the config file (although there is possibly some risk payments would never confirm).

b.) At the time the vendor accepts the payment, the fees are larger than the payment amount and therefore the payment will never go through.

This is a worse scenario. Ideally, if the vendors are properly educated, this wouldn't happen. But... to mitigate this I think we could probably do a tweaked version of what this PR implements. I don't think we should disable the pruchase button for everybody while we do the online check (which is going to take 10 - 60 seconds or more). I think we should figure out an algorythm to determine when it's likely this will be a problem - for example if the current fees are within 75% (or whatever) of the order total, it could behind the scenes make the call to determine whether the vendor is online. If the user does press the "Pay" button, then a spinner comes and if the online call still hasn't come back, when it does and it's determined the vendor is offline we could prevent the purchase with a clear explanation. I think it would be better to have that come in an alert modal rather than complicating the purchase flow layout with some extra messaging. If it's determined the vendor is online, then the spinner stays and we make the purchase call to the server.

Another option there is if we are quite certain the payment will fail, we could prevent it, whereas if we think it is at risk of failing, but not sure, we could explain that to the user in a dialog and ask them to confirm they want to risk it and send the payment anyway.

rmisio:
image

This IMHO is irrelevant messaging for a buyer. Why does the buyer care how much the vendor is getting? That's a vendor concern.

SamPatt:
>This IMHO is irrelevant messaging for a buyer. Why does the buyer care how much the vendor is getting? That's a vendor concern.

Vendors don't need to accept offline orders, they can reject them. If they do, the buyer pays a second fee back to themselves. If the vendor is left with a small amount after the purchase they may choose to reject all offline orders, and the buyer would pay two fees for nothing.

It's possible the best way to solve this is by allowing vendors to disable offline purchasing or another method, but I don't think presenting this information to the buyer is irrelevant.

rmisio:
@SamPatt I see your point, but that's not at all conveyed in that message.

jjeffryes:
> Vendors don't need to accept offline orders, they can reject them. If they do, the buyer pays a second fee back to themselves. If the vendor is left with a small amount after the purchase they may choose to reject all offline orders, and the buyer would pay two fees for nothing.

I think it would be good to be fully transparent about all fees, even if they're estimates. 

Show the fee you're going to pay when you purchase, show the amount you may get back if the order is cancelled (ie: if you or the vendor cancel, you only get back X), etc. etc.

I think that's a much better approach than what we're doing now, which is hiding critical information from the user that they need to make informed decisions. Even if the fee is tiny, concealing it leads to confusion and potentially serious negative consequences.

PR #1106, opened 2017-11-13
FEATURE: Add Block Booth search to default provider list.

ghost:


jjeffryes:
I noticed the BlockBooth API returns a filter for ratings, but it looks like all listings come back with an averageRating of 0.

The filter seems to be working, the problem is just the listing data not showing the ratings.

I'm not sure if the minimum sales amount filter does anything, as only 5+ and any returns any results. I'm pretty sure there are more than 4 listings with 5 or more sales.

PR #1118, opened 2017-11-24
Index abbr num test

andyjdee:
Noticed there was no test for js/utils/indexjs abbrNum() function so I added in the following test scenarios:

- Test for negative number
- Test for 0
- Test for n < 1000
- Test for n < 1000000
- Test for n < 1000000000
- Test for n < 1000000000000
- Test for n > 1000000000000
- Test for rounded numbers up to 6 decimal places

**(Please Note: This should not be merged until #1119 is merged)**

### DIFF
--- a/test/utils/index.js
+++ b/test/utils/index.js
@@ -78,7 +78,7 @@ describe('our index utilities', () => {
   });
 
   describe('have a abbrNum utility that', () => {
-    it('returns abbreviated a rounded number based on (1000(n) & maximum trillion)', () => {
+    it('returns abbreviated a rounded number for (1000(n) & max trillion)', () => {
       const sourceNumberSet = [123, 1266, 125468,
                              77547959, 388475766449,
                              48242478968789, 482424789687893242];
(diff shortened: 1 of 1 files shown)


test/utils/index.js
```diff
@@ -76,5 +76,23 @@ describe('our index utilities', () => {
...
+  describe('have a abbrNum utility that', () => {
+    it('returns abbreviated a rounded number based on (1000(n) & maximum trillion)', () => {
+      var sourceNumberSet = [123,1266,125468,77547959,388475766449,48242478968789,482424789687893242];
+      var formattedNumberSet = [123,'1.3k','125.47k','77.548m','388.476b','48.242t','482,424.79t'];
+      for (let i = 0; i < sourceNumberSet.length;i++) {
+        expect(abbrNum(sourceNumberSet[i],i)).to.be.equal(formattedNumberSet[i]);
+      }
+    });
+
+    it('returns no abbreviation if a negative number or 0', () => {
```
rmisio: I don't think this second test is necessary. It was a limitation in the utility function that it didn't abbreviate negative numbers. I've fixed that [here](https://github.com/OpenBazaar/openbazaar-desktop/pull/1119).

I don't think we need to test that it doesn't abbreviate on zero. What it's really doing is not abbreviating on < 1000, since the minimum abbreviate threshold is 1000.

test/utils/index.js
```diff
@@ -76,5 +76,23 @@ describe('our index utilities', () => {
...
+      var sourceNumberSet = [-12341235,0];
+      var formattedNumberSet = [-12341235,0];
+      for (let i = 0; i < sourceNumberSet.length;i++) {
+        expect(abbrNum(sourceNumberSet[i],i)).to.be.equal(formattedNumberSet[i]);
+      }
+    });
+  });
 });
```
rmisio: There are 23 linting errors introduced by this PR. Can you please run `npm run lint` and fix the errors?

andyjdee: Fixed lint errors.

See Commit: 20c49043f37c6a2aaf68864cb693d330d15046ac

test/utils/index.js
andyjdee: Thanks for the feedback!

I have removed the test for 0 as per your comments. Although, given your PR #1119 - you've added functionality to check negatives on line 63, therefore I feel its necessary to test for it. Also, there is a small bug in your code for numbers in range of -1 to -999 because you're using absolute numbers. See my comments in your pull request for patch.

PR #1119, opened 2017-11-28
patching the abbrNum utility function to handle negative numbers

rmisio:


js/utils/index.js
```diff
@@ -58,10 +58,11 @@ export function splitIntoRows(items, itemsPerRow) {
...
 // http://stackoverflow.com/a/2686098/632806
 // todo: unit test
-export function abbrNum(_number, _decPlaces = 1) {
+export function abbrNum(_number, _maxDecPlaces = 1) {
   // 2 decimal places => 100, 3 => 1000, etc
-  const decPlaces = Math.pow(10, _decPlaces);
-  let number = _number;
+  const decPlaces = Math.pow(10, _maxDecPlaces);
+  const isNegative = _number < 0;
+  let number = Math.abs(_number);
```
andyjdee: Change line 65 to the code mentioned below. With the changes you made would introduce a bug for negatives between -999 to -1. See comments in PR #1118.

`let number = _number > -1000 && _number < 0 ? _number : Math.abs(_number);`

With my changes to latest test and your changes, plus the above line change everything seems good.

rmisio: @andyjdee Thanks for catching that!

I fixed it in a slightly different way since I wanted the return type to consistently be a string whether the number needed to be abbreviated or not.

andyjdee:
Great work - I noticed now n < 1000 returns a string now. Well spotted.

I've added an assertion to ensure that it is of type string in PR #1118 to test this properly.

I've re-tested with your latest changes and its looking good from my end.

PR #1122, opened 2017-11-29
1100

### LINKED ISSUE #965: Make http/https ursl in chat into links
For example `https://bitcoinshirt.co/` should be a link there:

image

### LINKED ISSUE #1100: Don't count non-multihash as links in chat
In models/chat/ChatMessgage.js, some words that are not hashes are being treated as hashes by the isMultiHash function.

Examples appear to include all 5 letter words starting with "Th" followed by "a", "i" or "e".

rmisio:
closes #1100 
closes #965

### DIFF
--- a/js/models/chat/ChatMessage.js
+++ b/js/models/chat/ChatMessage.js
@@ -57,9 +57,9 @@ export function processMessage(message) {
 
           if ((w.startsWith('@') && w.length > 1) ||
             (w.startsWith('ob://') && w.length > 5) ||
-            (w.startsWith('http://') && w.length > 7) ||
-            (w.startsWith('https://') && w.length > 8) ||
-            (w.startsWith('www.') && w.length > 4) ||
+            (w.startsWith('http://') && w.length >= 11) ||
+            (w.startsWith('https://') && w.length >= 12) ||
+            (w.startsWith('www.') && w.length >= 8) ||
             isGuid) {
             wordsToAnchorify.push(w);
           }
--- a/js/models/chat/ChatMessage.js
+++ b/js/models/chat/ChatMessage.js
@@ -57,7 +57,9 @@ export function processMessage(message) {
 
           if ((w.startsWith('@') && w.length > 1) ||
             (w.startsWith('ob://') && w.length > 5) ||
-            is.url(w) ||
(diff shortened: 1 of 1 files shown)


js/models/chat/ChatMessage.js
```diff
@@ -50,9 +51,16 @@ export function processMessage(message) {
...
           if (wordsToAnchorify.includes(w)) return;
 
+          // Certain words are valid multihashes (e.g. that)
+          // const isGuid = isMultihash(w) && w.length > 15;
+          const isGuid = isMultihash(w) && w.length > 15;
+
           if ((w.startsWith('@') && w.length > 1) ||
             (w.startsWith('ob://') && w.length > 5) ||
-            isMultihash(w)) {
+            (w.startsWith('http://') && w.length > 7) ||
```
jjeffryes: Since the shortest url is the prefix plus one character plus a dot plus 2 characters (like http://a.io), should the length checks be 11, 12, and 8 instead?

js/models/chat/ChatMessage.js
```diff
@@ -66,10 +74,20 @@ export function processMessage(message) {
...
   wordsToAnchorify.forEach(word => {
-    const w = word.startsWith('ob://') ? word.slice(5) : word;
+    let href = !is.url(word) && !word.startsWith('ob://') ?
+      `#ob://${word}` : word;
+
+    if (is.url(word)) {
+      const link = document.createElement('a');
+      link.setAttribute('href', word);
+      if (link.protocol === 'file:') {
+        href = `http://${word}`;
```
jjeffryes: If the protocol was file, wouldn't it only work locally?

js/models/chat/ChatMessage.js
rmisio: If that is indeed the shortest url, then yes.

js/models/chat/ChatMessage.js
rmisio: file is what the protocol comes up as when you don't give a protocol. For example if the user types `www.espn.com`, the protocol (using this method) comes up as `file` and that to us is an indicator that we need to add in the `http`.

PR #1127, opened 2017-11-30
Reputation tab

### LINKED ISSUE #1109: Reputation Tab
## User Story

As a buyer, I want to be able to look at a vendor's page and see all of their reviews in one place, so I can get an idea of how trustworthy they are independent of the reviews on a single listing.

As a seller, I want an easy way for buyers to see all of my reviews, so they can see that I can be trusted even if the listing they are interested in has no or few reviews.

## Functional Requirements
- [x] Show the Reputation Tab when the user's reputation score is clicked in the User Page top bar.
- [ ] When the user's reputation score is clicked in a User Card open the User Page with the Reputation Tab open.
- [x] There is no visible tab in the top bar of the User Page for the Reputation Tab. When it is active, no tab is marked as active.

### Reputation Screen
- [x] There will be a top area with the total number of reviews, the average for the past 90 days, and the all time average.
- [x] When loaded, the first 5 reviews will be loaded and shown. 
- [x] There will be a Load More button. When clicked it will load 5 more reviews.

### Review
- [x] Each review should have the current data returned from the fetchReviews API (date, optional user name, review text and star ratings)
- [x] Each review should also have a thumbnail of the listing. This should be given as an object with all the image sizes. 
- [x] If the image doesn't exist or can't load, a placeholder will be shown.
- [x] Each review should also have the title of the review. 
- [x] The image and title of the listing should be links that when clicked on load the listing.
- [ ] When the title is hovered over, it should have an underline.
- [ ] Update the user link to match the title link on hover.

### Ratings API
- [ ] An API will need to be created to get the rating hashes for a node.
- [ ] It should be very similar to the ratings API.
    - or the existing rating api can be modified, to return rating hashes from all reviews if no slug is provided.
- [ ] The API should accept a peerID to return ratings for.
- [ ] The API should return an array of hashes of reviews.
- [ ] The API should return the total average rating.

### Individual Reviews API
- [ ] It should be very similar to fetchratings (or fetchratings can be used, if possible)
- [ ] When given an array of review hashes, it should return the data for those hashes.
- [ ] It should have an asyc option, when true it returns an id for websocket messages
- [ ] The format of the returned messages should be identical to those returned from ratings.

## Notes
- the listing title and image will need to be saved when the review is made
- this may break compatibility with older nodes making purchases, because the purchase data will be different.

jjeffryes:
This adds a reputation tab that shows all of the user's reviews to the user page.

It can be accessed by clicking on the star rating next to their name at the top of the page.

To test this, use it with the ratings server branch.
https://github.com/OpenBazaar/openbazaar-go/pull/768

Reviews won't have titles or thumbnails unless you leave a new review with that branch.

You can see an example of a review with a title and thumbnail at ob://QmczPH7Pf7EC67eahRHXLYxTwo3nxYx4W7qEWznsNtCbm4/reputation

Note: I moved the reviews and review files to their own folder, since they're used both in the listing details and in the reputation tab.

Closes #1109


js/templates/modals/moderatorDetails.html
```diff
@@ -27,14 +27,14 @@
...
-      <div class="rowDivV clrBr TODO"></div>
+      <div class="rowDivV clrBrBk TODO"></div>
       <div class="flexExpand TODO">
         <div class="flexCol flexCent">
           <%= ob.parseEmojis('👍') %> XX<% // placeholder for reputation %>
           <div class="tx5b clrT2"><%= ob.polyT('moderatorDetails.recommendations') %></div>
         </div>
       </div>
-      <div class="rowDivV clrBr"></div>
+      <div class="rowDivV clrBrBk"></div>
```
jjeffryes: This is fixing a wrong color on the dividing lines I noticed while making other changes (the same lines are used in the reputation page).

js/templates/reviews/review.html
```diff
@@ -0,0 +1,73 @@
+<% if (ob.error) { %>
+  <p class="clrTErr txCtr"><i class="ion-alert-circled"></i> <%= ob.polyT('listingDetail.errors.loadReview', { id: ob.ratingId, error: ob.error }) %></p>
```
jjeffryes: If the review fails to load, and error will be shown in it's place. This fixes various situations where the count would be off, since the user can see that there should be X reviews (ie: we don't want to show 4 for the number of reviews but only 3 load because 1 failed silently).

js/templates/reviews/review.html
```diff
@@ -0,0 +1,73 @@
...
+<% } else { %>
+
+  <%
+    let title = ob.vendorSig.metadata.listingTitle ? ob.vendorSig.metadata.listingTitle : ob.vendorSig.metadata.listingSlug;
+    title = title ? title : ob.polyT('reputation.noTitle');
```
jjeffryes: If the title fails to load because it's an old review, the slug will be shown instead. That's more user friendly than showing "title not found" in my opinion.

styles/components/_layout.scss
```diff
@@ -62,7 +62,8 @@
   //a vertical divider. Use when you need to separate things without adding a border to one of them
   //use a color class for the background color
   width: 1px;
-  height: 100%;
+  flex-shrink: 0;
+  align-self: stretch;
```
jjeffryes: This fixes the vertical divider.

styles/components/_misc.scss
```diff
@@ -71,6 +71,10 @@
       opacity: 0.4;
       pointer-events: none;
 
+      .emoji {
+        margin-right: 0.3em;
```
jjeffryes: This is only in the rating strip, the margin right was removed from the other emoji instances.

styles/components/_misc.scss
```diff
@@ -91,9 +95,8 @@
 }
 
 .emoji {
-  height: 1em;
-  width: 1em;
-  margin-right: 0.3em;
+  height: 1em !important; // don't allow parent text sizes to override this
+  width: 1em !important;  // don't allow parent text sizes to override this
```
jjeffryes: This fixes some issues where the emoji would be off center if the parent text size was in ems.

morebrownies:
screen shot 2017-12-05 at 1 24 25 pm

I noticed all of my reviews have the slug displaying. You mentioned titles won't display on older reviews. However, could we capitalize the first character and remove the dashes to make it appear as a title?

morebrownies:
screen shot 2017-12-05 at 1 23 45 pm

Also, my fault on this one... Can we push "overall rating" and "reviews" under the values (as pictured above). Not really feeling the other way i designed it.

morebrownies:
Do reviews display in any order? Can we not order by most recent at the top?

jjeffryes:
> could we capitalize the first character and remove the dashes to make it appear as a title?

That's a good idea.

> Can we push "overall rating" and "reviews" under the values (as pictured above). 

Sure.

> Do reviews display in random order? Can we not order by most recent at the top?

They come in randomly over web sockets. We could order them by date, but then they'd jump around visually as new ones come in and change the order.

jjeffryes:
@morebrownies the changes you suggested were made, aside from ordering by date (because it would cause jumping).

morebrownies:
Nice looks good to me

rmisio:
image

Should "review" be singular in this case?

jjeffryes:
> Should "review" be singular in this case?

Yes, good catch.

rmisio:
image

@morebrownies should that spinner have a border around it?

morebrownies:
No border should be needed on the spinner @rmisio

rmisio:
@jjeffryes the spinner doesn't show up until the fetchRatings call comes back. Until then it shows text indicating the user has no reviews:

image

Can the spinner precede the fetchRatings call as well?

rmisio:
> No border should be needed on the spinner @rmisio

@jjeffryes also notice, not only is there a border around the spinner, there's another detached border above it (which looks out of place).

js/views/userPage/Reputation.js
```diff
@@ -1,27 +1,65 @@
...
 import BaseVw from '../baseVw';
 import loadTemplate from '../../utils/loadTemplate';
+import app from '../../app';
+import Reviews from '../reviews/Reviews';
+import { openSimpleMessage } from '../modals/SimpleMessage';
+import Profile from '../../models/profile/Profile';
 
 export default class extends BaseVw {
   constructor(options = {}) {
+    if (!options.model && !(options.model instanceof Profile)) {
```
rmisio: should this be an 'or'?

PR #1133, opened 2017-12-05
Help in nav menu

jjeffryes:
This adds the help link to the nav menu for #1125

### DIFF
--- a/js/views/modals/about/About.js
+++ b/js/views/modals/about/About.js
@@ -31,7 +31,7 @@ export default class extends BaseModal {
       Help,
     };
 
-    this.currentTabName = this.tabViews[opts.initialTab] ? opts.initialTab : 'Story';
+    this.currentTabName = opts.initialTab;
     this.isBundledApp = remote.getGlobal('isBundledApp');
     this.updatesSupported = remote.getGlobal('updatesSupported');
   }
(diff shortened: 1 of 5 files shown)


js/views/modals/about/About.js
```diff
@@ -30,7 +31,7 @@ export default class extends BaseModal {
       Help,
     };
 
-    this.currentTabName = 'Story';
+    this.currentTabName = this.tabViews[opts.initialTab] ? opts.initialTab : 'Story';
```
rmisio: If we're already defaulting the initalTab up in the opts, can't this just be?

`this.currentTabName = opts.initialTab`

rmisio: I think I see why you did that, but I think the protection would need to be in the `selectTab` function since selecting a tab may not always go through the constructor.

```
  selectTab(tabViewName) {
    let tabView = this.tabViewCache[tabViewName];

    if (!tabView && !this.tabViews[tabViewName]) {
      throw new Error('You are attempting to select an invalid tab.');
    }
    ...
```

jjeffryes: Yeah, good point.

morebrownies:
Thanks, looks good

PR #1139, opened 2017-12-09
search checkbox filter

premek:
The checkboxes didn't work accoring to https://github.com/OpenBazaar/obips/blob/master/obip-0002.md
but I think they were not used by any search engine yet(?).

My change uses just one of many possible ways of sending multi-value form field in query params. I used dot-notation for minimal client code change.

Example: 
My test search engine defines "color" filter with "red", "green", "blue" checkboxes.
The client sends "color.green=true" query param if Green checkbox is checked and "false" value or missing query param if it is not checked.
`/search/listings?q=*&network=mainnet&p=0&ps=24&nsfw=true&color.green=true&color.blue=false`

Note other ways of doing the same
- `...&color=green&color=blue`
- `...&color[]=green&color[]=blue` (PHP way?)
- `...&color[green]=true&color[blue]=true`
- `...&color[green]=on&color[blue]=on`

Also it would be nice to clarify this in obip-0002.


jjeffryes:
@premek thank you for submitting this. The current client code should work with checkboxes as-is if the search provider provides data in the expected format. We're reviewing that format now, if we revise it so the current code no longer works we can take a look at changing the client to match the new format.

premek:
Sure, I was just playing with making my search engine and maybe I did understand the format wrong, or maybe the specification needs updating?

tyler-smith:
@jjeffryes @premek After thinking about it and testing some things out I think I like `...&color=green&color=blue` best. If there's no object I'll update OBIP2.

tyler-smith:
I tried to make the change in the client but it just sends "true" instead of the value.

premek:
@tyler-smith Have you tried adding value="..." attribute to the <input> tag? But I remember maybe there was some other code actually sending the value. I cannot test it right now but will try to take a look later.

jjeffryes:
@tyler-smith that's fine. I have a PR ready that will format it that way.

@premek I had to refactor a few things to handle multiple values for single keys. I'll add the changes to this PR and submit it as a new PR.

rmisio:
closing in lui of #1161

PR #1141, opened 2017-12-11
No purchase block [WIP]

jjeffryes:
This removes the code that blocks a user from completing a moderated purchase if the estimated fee is too high.

***Do not merge*** this until the server PR to block moderated and offline purchases when the fee is too high has been merged.

This PR also refactors most of the files in the purchase flow to use getCachedEl and state changes instead of jquery based changes.

### DIFF
--- a/js/views/baseVw.js
+++ b/js/views/baseVw.js
@@ -250,9 +250,9 @@ export default class baseVw extends View {
       newState = _.extend({}, this._state, state);
     }
 
-    if (opts.renderOnChange && !_.isEqual(this._state, newState)) {
+    if (!_.isEqual(this._state, newState)) {
       this._state = newState;
-      this.render();
+      if (opts.renderOnChange) this.render();
     }
 
     return this;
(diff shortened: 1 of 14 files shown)


js/views/baseVw.js
```diff
@@ -250,9 +250,9 @@ export default class baseVw extends View {
       newState = _.extend({}, this._state, state);
     }
 
-    if (opts.renderOnChange && !_.isEqual(this._state, newState)) {
+    if (!_.isEqual(this._state, newState)) {
       this._state = newState;
-      this.render();
+      if (opts.renderOnChange) this.render();
```
jjeffryes: This fixes a bug where the state was not updated if renderOnChange was false.

js/views/modals/purchase/Purchase.js
```diff
@@ -104,14 +111,17 @@ export default class extends BaseModal {
       selectFirst: true,
       radioStyle: true,
     });
+    // render the moderators so it can start fetching and adding moderator cards
+    this.moderators.render();
+    this.moderators.getModeratorsByID();
```
jjeffryes: The render has to happen here because getModeratorsByID adds moderators to the DOM, and we don't want to re-fetch them by calling this in the render.

The moderator view could probably be refactored to fix that, but it should be a separate PR.

jjeffryes:
This has gotten out of date enough untangling the merge conflicts is a problem, I'm going to close it and make a fresh PR.

PR #1143, opened 2017-12-14
fix wrong var in template

jjeffryes:
The moderator information under the user card in the home tab was not being shown due to the variable being checked having been changed and not updated in the template.


jjeffryes:
Closing this, there may be additional related issues that need to be fixed, I'll add them to this PR.

jjeffryes:
Looks like that was the only issue after more testing. This should be the only change needed.

PR #1144, opened 2017-12-14
Altcoin switching in the bundled app

### LINKED ISSUE #987: Feature request: Add Zcash support to front-end
Given the ability for OpenBazaar to be pseudonymous and run over Tor, it would be fitting to adopt native support for [Zcash](https://z.cash) with its strong privacy protections. In my mind, this would complete the privacy circle to create a truly anonymous marketplace.

As @cpacia has already written server-side support for Zcash, and @SamPatt indicated only client-side integration was lacking, the [challenge has been issued](https://twitter.com/zooko/status/908886341584334848) for assistance from the community to implement the front-end.

I'll add 1 ⓩ to a bounty for the person who submits a pull request which is merged that adds full ZEC support alongside BTC.

### LINKED ISSUE #1121: Altcoin support for bundled app
The following are some preliminary designs for handling alt coins in the bundled app, notably how to configure your server or servers to work with a particular coin.

So the first time you run the client you will need to select what crypto you want your wallet in:

image

Then if you want to support any other coins, you would do so by creating a new server configuration. The first question will be the server type (built in or stand-alone).

image

If you select built-in, then there will be a “wallet coin type” field as part of the data you need to enter for the server configuration.

image

If you selected stand-alone, the same form as is currently in the app is presented where you basically need to fill out various things (e.g. ip, port, etc..). In this case, you will not get a wallet coin type. You will need to configure that via the stand alone server’s config file:

image

This is what multiple built-in servers (for different coins) would look like:

image

rmisio:
The PR incorporates the [UI](https://app.zeplin.io/project/5a2865874a2a4b075cd422c5) to select a Wallet Currency for your bundled server and create a seperate bundled server configs for each coin.

If you are running this branch from source, you will need to tweak your code so the app thinks it's running the bundled app. This can be done as follows:

1.) Make sure your server is on the latest [altcoins](https://github.com/OpenBazaar/openbazaar-go/pull/775) branch.
2.) In the server dir, compile a binary via `go build openbazaard.go`.
3.) back in the client source... in main.js, change the following lines using the path to your server branch in step 1:
```
const serverPath = '/Users/rob/develop/openbazaar/v2/server/';
```

closes #1121 
closes #987


jjeffryes:
When I followed the instructions for testing, the only way to create a new server was the run the app, let it fail to connect to the created server, then close the client and start it again, at which point it would work normally with the new server. The same happens when creating a new server from the server management modal. It fails to connect, and clicking retry throws a `Uncaught Error: The provided model is not in this collection and must be to set it as the active config.` error on line 39 of ServerConfigs.js

I also see an error when trying to switch between servers, `Uncaught Error: Could not call remote function 'stop'. Check that the function signature is correct. Underlying error: Cannot read property 'spawn' of undefined
Error: Could not call remote function 'stop'. Check that the function signature is correct.`

I don't know that it's worth looking into, since obviously this isn't using the code as intended, just something to be aware of when testing.

jjeffryes:
For anyone testing on Windows, I had to use a server path with escaped slashes like this:
`this.serverPath = 'C:\\Go\\src\\github.com\\OpenBazaar\\openbazaar-go\\';`

js/templates/modals/connectionManagement/configurationForm.html
```diff
@@ -22,94 +20,180 @@ <h2 class="h3 clrT txUnl txCtr"><%= ob.title %></h2>
...
+          <div class="btnStrip">
+            <div class="btnRadio">
+              <input type="radio"
+                     name="serverType"
+                     value="BUILT_IN"
+                     id="serverConfigBuiltin"
+                     data-field-builtin
+                     data-field-standalone
+                     <% if (ob.builtIn === true) { %>checked<% } %>>
+              <label class="gutterHSm" for="serverConfigBuiltin"><span><%= ob.polyT('connectionManagement.configurationForm.serverTypeBuiltin') %></span><span class="toolTip" data-tip=""><i class="ion-help-circled"></i></span></label>
```
jjeffryes: Why is the data-tip value blank? No tooltip appears when the help icon is hovered over.

js/templates/modals/connectionManagement/configurationForm.html
```diff
@@ -22,94 +20,180 @@ <h2 class="h3 clrT txUnl txCtr"><%= ob.title %></h2>
...
+            </div>
+            <div class="btnRadio">
+              <input type="radio"
+                     name="serverType"
+                     value="STAND_ALONE"
+                     id="serverConfigStandalone"
+                     data-field-builtin
+                     data-field-standalone                     
+                     <% if (ob.builtIn === false) { %>checked<% } %>>
+              <label class="gutterHSm" for="serverConfigStandalone"><span><%= ob.polyT('connectionManagement.configurationForm.serverTypeStandalone') %></span><span class="toolTip" data-tip=""><i class="ion-help-circled"></i></span></label>
```
jjeffryes: The data-tip is blank here too. Should something be giving it a value?

js/templates/modals/connectionManagement/configurationForm.html
rmisio: just waiting on copy here. Whoever wants is welcome to provide it.

js/models/ServerConfig.js
```diff
@@ -31,6 +34,31 @@ export default class extends BaseModel {
...
   }
 
+  set(key, val, options = {}) {
+    // Handle both `"key", value` and `{key: value}` -style arguments.
+    let attrs;
+    let opts = options;
+
+    if (typeof key === 'object') {
+      attrs = key;
+      opts = val || {};
```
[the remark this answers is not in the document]
rmisio: lol... this is the second time you've flagged this code and this is actually code right from the Backbone source :)

The way set() works is that you could either call it like `model.set(attr, value, options)` or `model.set({ attr1: val, attr2: val}, options)`. This code facilitates that and ensures that even if the user is passing it in the first way, it will convert it to an object so the rest of the method just needs to handle the object format.

js/views/modals/WalletSetup.js
```diff
@@ -0,0 +1,90 @@
...
+
+  render() {
+    super.render();
+    loadTemplate('modals/walletSetup.html', t => {
+      this.$el.html(t({
+        cryptoCurs: this.cryptoCurs,
+        errors: this.model.validationError || {},
+        ...this.model.toJSON(),
+      }));
+      super.render();
```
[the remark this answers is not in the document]
rmisio: no, good catch - the top one should go away.

js/models/ServerConfig.js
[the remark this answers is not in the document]
jjeffryes: I see. I'll make sure to read the entire Backbone source code and cross reference all previous reviews before each new code review. It may slow down the review process a little, but I'm sure it's worth it! ;)

js/templates/modals/connectionManagement/configurationForm.html
jjeffryes: Built-in: "This server is created and controlled by this app on this computer, and only runs when this app is running. This is recommended for most users."

Stand-Alone: "This server is manually installed and controlled from the command line. It might be installed on a remote computer. It must already be running before you can connect to it."

rmisio: @morebrownies for any of the "i" icons that don't show a tip on hover, feel free to provide any copy you think should be in them.

rmisio: Oh sorry wolf, didn't see that Josh kind of did that already.

PR #1145, opened 2017-12-15
making selectize create a tag on blur by default

### LINKED ISSUE #1128: Categories Disappear When Blurred
When entering a category in the edit listing screen, if the category has not been changed to a tag by clicking enter or tab, and the user clicks outside the input element or clicks the save button, the category is deleted and not saved.

The input should convert any text in it to a tag when the text input portion loses focus.

rmisio:
closes #1128


rmisio:
closing in liu of #1151

PR #1149, opened 2017-12-17
Patch cur check

rmisio:
Fixes issue resulting from the fact that the exchange rate API stopped including BTC.


hoffmabc:
Should we merge this? It LGTM.

PR #1152, opened 2017-12-23
Fixed typos

BillStrait:
Corrected typos in readme.


BillStrait:
Sorry, I'm pretty newbish with Git. This seemed like a simple change, not sure why the CI tool would fail it. Did I do something wrong?

hoffmabc:
It fails because there is some encryption of keys in the process that don't work on a submitted branch. I plan to refactor that in the future.

Issue #417, opened 2017-03-24
Variant inventory table header should stay fixed as I scroll


morebrownies:
I find myself frequently scrolling back up to view the column titles in the variant inventory section. It gets especially complicated when there are a lot of variants and choices.

Currently, when I scroll, I see: (table header scrolls off screen)
![screenshot 2017-03-24 10 35 32](https://cloud.githubusercontent.com/assets/13066886/24306542/3091afc6-107e-11e7-9eda-3260c89a046e.png)

Ideally, the table header would stay fixed as I scroll: 
![screenshot 2017-03-24 10 47 14](https://cloud.githubusercontent.com/assets/13066886/24306848/454beed0-107f-11e7-8e43-07cc220932ee.png)

BillStrait:
I have this done. I did a bad job with my change management, so it's waiting on another PR before I submit it.

morebrownies:
great! thank you @BillStrait!

PR #1155, opened 2017-12-27
Sticky headers for varients table

BillStrait:
Updated css/corrected table structure to allow for sticky table headings.

This should resolve issue  #417

### DIFF
--- a/js/templates/modals/editListing/variantInventory.html
+++ b/js/templates/modals/editListing/variantInventory.html
@@ -3,15 +3,15 @@
 <% } else { %>
 <div class="inventoryTableWrap">
   <table>
-    <tr>
+    <thead>
       <% ob.columns.forEach(column => { %>
-      <th class="clrBr"><%= column %></th>
+      <td class="clrBr"><div class="stickyDiv"><%= column %></div></td>
       <% }) %>
-      <th class="clrBr">Surcharge</th>
-      <th class="clrBr">Total Price</th>
-      <th class="clrBr">Sku</th>
-      <th class="clrBr quantityCol">Quantity</th>
-    </tr>
+      <td class="clrBr"><div class="stickyDiv">Surcharge</div></td>
+      <td class="clrBr"><div class="stickyDiv">Total Price</div></td>
+      <td class="clrBr"><div class="stickyDiv">Sku</div></td>
+      <td class="clrBr quantityCol"><div class="stickyDiv">Quantity</div></td>
+    </thead>
   </table>
 </div>
 <div class="clrT2 txSm padSm"><%= ob.polyT(`editListing.variantInventory.helperText`) %></div>
--- a/js/templates/modals/editListing/variantInventory.html
+++ b/js/templates/modals/editListing/variantInventory.html
@@ -5,12 +5,12 @@
   <table>
     <thead>
       <% ob.columns.forEach(column => { %>
-      <td class="clrBr"><div class="stickyDiv"><%= column %></div></td>
+      <td class="clrP"><div class="stickyDiv border clrBr"><%= column %></div></td>
       <% }) %>
-      <td class="clrBr"><div class="stickyDiv">Surcharge</div></td>
-      <td class="clrBr"><div class="stickyDiv">Total Price</div></td>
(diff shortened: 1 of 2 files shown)


js/templates/modals/editListing/variantInventory.html
```diff
@@ -3,15 +3,15 @@
 <% } else { %>
 <div class="inventoryTableWrap">
   <table>
-    <tr>
```
rmisio: A couple things:

1.) The sizing seems a bit off in that the variant table is unnecessarily getting a horizontal scrollbar it wasn't getting before. There are certain situations where it's unavoidable (e.g. if you're tracking inventory since that adds in an extra checkbox), but by default there wasn't one there before.

2.) The header row has a different border than it used to have - I'm not sure that's necessary. Also, the header labels used to be bold and they no longer are.

Before:

image

After:

image

styles/modules/modals/_editListing.scss
```diff
@@ -240,13 +239,33 @@
...
           border-spacing: 0;
           border-collapse: collapse;
           width: 100%;
 
+          thead tr td {
+            position: -webkit-sticky;
+            position: sticky;
+            top: 0px;
+            padding: 0px;
+            margin:0px;
```
rmisio: missing space.

styles/modules/modals/_editListing.scss
```diff
@@ -240,13 +239,33 @@
...
+            border: none;
+            z-index:5;
+          }
+
+          thead tr td .stickyDiv {
+            width: inherit;
+            height: inherit;
+            background-color: white;
```
rmisio: When it comes to colors, we have planned the ability for the user to be able to switch themes. It hasn't been implemented yet, but the general idea is to not hard code colors in the css (outside of the themes file).

Basically, instead of hard-coded the colors here, you want to add certain classes to the respective elements.

So, instead of declaring the following in the css:

```
background-color: white;
border: 1px solid grey;
```

In the markup, you would want to add the `border clrBr clrP` classes to the element your targeting.

js/templates/modals/editListing/variantInventory.html
```diff
@@ -3,15 +3,15 @@
...
       <% ob.columns.forEach(column => { %>
-      <th class="clrBr"><%= column %></th>
+      <td class="clrBr"><div class="stickyDiv"><%= column %></div></td>
       <% }) %>
-      <th class="clrBr">Surcharge</th>
-      <th class="clrBr">Total Price</th>
-      <th class="clrBr">Sku</th>
-      <th class="clrBr quantityCol">Quantity</th>
-    </tr>
+      <td class="clrBr"><div class="stickyDiv">Surcharge</div></td>
```
rmisio: From the examples I've seen, I think these should be ths, even when they are in a thead. That would give you the bolding.

BillStrait: Chrome does not like the `position: sticky;` attribute on TH. THEAD can contain any valid TR. TR can contain either TH or TD. TH is used to mark the cells as header cells specifically, but using TD is not invalid. Source: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/thead 

Will restore bolding by applying correct classes/adding a bold attribute.

jjeffryes:
@BillStrait were you planning to update this with the requested changes?

BillStrait:
That's the plan, yes. I apologize, my day job has gotten a bit busy after everyone came back from vacation. I hope to have some time to correct it this week.

BillStrait:
I believe all requested changes (except replacing td's with th's, which I hope I justified adequately) are complete.

jjeffryes:
@BillStrait I apologize, I haven't looked closely at this until now.

Using position sticky isn't the best way to solve this. It creates a layout issue because the sticky table cells size themselves independently, and you can get something like this:

image

Have you looked at putting the overflow: auto on the tbody so it scrolls within that and the thead stays at the top?

BillStrait:
Sounds like you've got a solution in mind. I'll see if I have time to implement it later.

jjeffryes:
I'm not sure what would be involved with setting the tbody to scroll. It may behave weirdly without fixed widths and heights. 

I think there's also a way to do this with absolutely positioned th elements.

PR #1158, opened 2017-12-29
Hot fixes

rmisio:
The main fix here is shutting down the OB server in the bundled app using the API as opposed to the previous way using `taskkill` which doesn't allow the server to properly shutdown.

to test:
1.) Compile the latest server source via `go build openbazaard.go`.
2.) In the client source... in main.js, change the following lines using the path to your server binary in step 1:

const serverPath = '/Users/rob/develop/openbazaar/v2/server/';

It's probably not a bad idea to clear your local storage first.

### DIFF
--- a/js/utils/serverConnect.js
+++ b/js/utils/serverConnect.js
@@ -325,15 +325,41 @@ export default function connect(server, options = {}) {
     if (connectAttempt) connectAttempt.cancel();
   };
 
-  // If we're not connecting to the local bundled server or it's running for a different coin,
-  // then let's ensure it's stopped.
+  if (server.get('builtIn') && !localServer) {
+    // This should never happen to normal users. The only way it would is if you are a dev
+    // and mucking with localStorage and / or fudging the source for the app to masquerade
+    // as a bundled app.
+    throw new Error('A configuration for a built-in server should only be used on ' +
+      ' the bundled app.');
+  }
+
+  const curLocalServerCoin = serverStartArgsToCoin();
+  let curLocalServerZecBinaryPath = null;
+
+  if (localServer) {
+    if (curLocalServerCoin === 'ZEC') {
+      const zecIndex = localServer.lastStartCommandLineArgs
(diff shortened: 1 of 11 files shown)


js/utils/serverConnect.js
```diff
@@ -325,15 +325,41 @@ export default function connect(server, options = {}) {
...
-  // If we're not connecting to the local bundled server or it's running for a different coin,
-  // then let's ensure it's stopped.
+  if (server.get('builtIn') && !localServer) {
+    // This should never happen to normal users. The only way it would is if you are a dev
+    // and mucking with localStorage and / or fudging the source for the app to masquerade
+    // as a bundled app.
+    throw new Error('A configuration for a built-in server should only be used on ' +
+      ' the bundled app.');
+  }
+
```
rmisio: This has grown a bit more in complexity than I would have liked. In determining whether we need to bounce an already running local server, it not only needs to factor in the coin the existing server is running in, but also the zcash binary path (if zec) and also (further down) the tor settings.

It's very manual right now in that if we add any additional server start args, some code will be needed to check for them. I'd like to simplify by just comparing what start args the existing server is running with vs. what we need and if they differ, bounce. But... I need to refactor a thing or two to make that happen so I'm going to save that as a down-time task.

jjeffryes:
If you try to delete the last server configuration in the configuration manager modal, it throws an unhandled error in Configurations.js on line 106.

```
Uncaught TypeError: Cannot read property 'remove' of undefined
    at _class.<anonymous> (C:/Users/J/ob/openbazaar-desktop/js/views/modals/connectionManagement/Configurations.js:106:7)
```

jjeffryes:
Sometimes when creating a new server I see an unhandled error in configurations.js

```
Uncaught TypeError: Cannot read property 'get' of undefined
    at _class.<anonymous> (C:/Users/J/ob/openbazaar-desktop/js/views/modals/connectionManagement/Configurations.js:62:23)
```

PR #1161, opened 2018-01-02
Search checked and currency

### LINKED ISSUE #1177: Setting for Accepted Currency in Search Doesn't Stick
If I'm running a BCH or ZEC server and I search I want to find listings that I can purchase so I select the correct currency in Accepted Currency. If I leave search to a listing page or profile and come back, the setting is cleared and is back to BTC.

jjeffryes:
This does the following:

- implements checkboxes according to the method planned for the server data
- fixes filters so the checked value is used, then the default value if no checked value is present
- refactors some code to remove the use of the filters object in the template, as it's only needed for the initial query, when the template is rendered better data is available from the search provider.
- sends an acceptedCurrency parameter to the search provider based on the crypto currency being used by the user's node.
- reduces the size of the filter text here and in the userpage store tab
- closes ##1162
- closes #1177


jjeffryes:
@rmisio it's actually built on top of #1139, so merging this should merge that.

js/templates/search/search.html
```diff
@@ -46,30 +46,31 @@ <h3 class="txUnl rowSm"><%= ob.name %></h3>
...
-                  }
-                  print(`<option ${selected} value="${option.value}">${option.label}</option>`);
-                }); %>
+                <%
+                  // if any option has a selected value use the first one. Otherwise use the first default.
+                  let selectedIndex = val.options.findIndex(opt => opt.checked);
+                  selectedIndex = selectedIndex === -1 ? val.options.findIndex(opt => opt.default) : selectedIndex;
+                  _.each(val.options, function(option, ind) {
+                    const selected = selectedIndex === ind ? 'selected' : '';
+                    // parsing the label for emojis isn't needed here because the select list is replaced by selectize.js
```
rmisio: Actually, selectize is used for tagging fields, but we're still using select2 for regular dropdowns and that's what this view uses for this dropdown.

Also, in the Search.js, I noticed that select2 is not declared as a dependency, but should be. The reason it's not bombing is because some other module imported it, which added it to the jQuery object, but we shouldn't depend on that.

js/templates/search/search.html
```diff
@@ -79,13 +80,23 @@ <h4 class="rowSm"><%= val.label %></h4>
...
-                  print(`<input type="checkbox" ${checked} id="${option.name}" name="${option.name}">`);
-                  print(`<label for="${option.name}">${parsedLabel}</label>`);
-                }); %>
+                <%
+                  // find all the values that should be checked
+                  const checkedVals = new Set();
+                  val.options.forEach(opt => {
+                    if (opt.checked) checkedVals.add(opt.value);
+                  });
+                  val.options.forEach((option, index) => {
```
rmisio: Why the second separate forEach, instead of just adding this code to the first one?

js/templates/userPage/store.html
```diff
@@ -27,10 +27,10 @@ <h3><%= ob.polyT('userPage.store.unableToFetchListings') %></h3>
   <div class="flex">
     <div class="col storeFilters <% if (!ob.listingCount) print('disabled') %>">
       <div class="clrP clrBr padMd clrT clrSh2 contentBox">
-        <div class="txB row"><%= ob.polyT('userPage.store.shippingFilter.heading') %></div>
-        <div class="flexVCent tx5 rowSm">
+        <div class="txB rowSm"><%= ob.polyT('userPage.store.shippingFilter.heading') %></div>
+        <div class="flexVCent rowSm">
           <label class="margRSm" for="shipsToSelect"><%= ob.polyT('userPage.store.shippingFilter.shipsTo') %>:</label>
-          <select class="tx6 select2Small js-shipsToSelect" style="width: 133px">
+          <select class="tx6 select2Small js-shipsToSelect" style="width: 133px" id="shipsToSelect">
```
rmisio: Is the id necessary? It was added in, but I'm not seeing anything added in that uses it.

js/views/search/Search.js
```diff
@@ -152,16 +157,24 @@ export default class extends baseVw {
...
+    // if the DOM is ready, use the form values. Otherwise use the values set in the constructor.
+    let filters = this.$filters ? this.$filters.serialize() : $.param(this.filterParams);
+    filters = !filters || reset ? '' : `&${filters}`;
+    const newURL = new URL(`${this.providerUrl}?${query}${network}${sortBy}${page}${filters}`);
+    const newParams = newURL.searchParams;
+    if (!newParams.has('nsfw')) {
+      newParams.append('nsfw', app.settings.get('showNsfw'));
+    }
+    if (!newParams.has('acceptedCurrencies')) {
+      newParams.append('acceptedCurrencies', app.serverConfig.cryptoCurrency);
```
rmisio: Didn't we fall on always sending the non-testnet value, in which case we should be using `getServerCurrency().code`?

main.js
```diff
@@ -84,7 +84,9 @@ if (handleStartupEvent()) {
   console.log('OpenBazaar started on Windows...');
 }
 
-const serverPath = `${__dirname}${path.sep}..${path.sep}openbazaar-go${path.sep}`;
+
+//const serverPath = `${__dirname}${path.sep}..${path.sep}openbazaar-go${path.sep}`;
+const serverPath = 'C:\\GoWork\\src\\github.com\\OpenBazaar\\openbazaar-go\\';
```
rmisio: Let's not forget to revert this back.

js/templates/search/search.html
jjeffryes: Because first it's finding out if any of the options are checked. The second loop only applies the default if none of the options are checked.

It could be done more concisely with a filter.

js/templates/userPage/store.html
jjeffryes: The label uses it for the "for" attribute. Previously the for wasn't pointing to anything.

js/views/search/Search.js
jjeffryes: Good catch. I updated it in the initial filters and forgot to here.

main.js
jjeffryes: Whoops!

js/views/search/Search.js
jjeffryes: Actually, this should be removed, since we're using no value for acceptedCurrencies to mean search for all of them.

PR #1173, opened 2018-01-10
Fix Shapeshift

jjeffryes:
- fix the payment address given to Shapeshift
- add a parameter for whether currencies can use Shapeshift to purchase
- change the name of the canShapeShiftInto to canShapeShiftIntoWallet

### DIFF
--- a/test/data/cryptoCurrencies.js
+++ b/test/data/cryptoCurrencies.js
@@ -114,6 +114,12 @@ describe('the crypto currencies data module', () => {
       });
     });
 
+    it('that requires the canShapeShiftIntoPurchase to be a boolean', () => {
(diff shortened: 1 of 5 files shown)


test/data/cryptoCurrencies.js
```diff
@@ -108,9 +108,9 @@ describe('the crypto currencies data module', () => {
       });
     });
 
-    it('that requires the canShapeShiftInto to be a boolean', () => {
```
rmisio: Would you mind adding a similar test for `canShapeShiftIntoPurchase` to ensure it's being provided as a boolean?

jjeffryes: Updated.

PR #1180, opened 2018-01-12
Blocked

### LINKED ISSUE #1062: Block Users [Rough Draft]
# User Story
As a user, when I see a user node I don't want to see again for any reason, I want to be able to take an action to prevent them from appearing in lists of users. If I navigate directly to their node, I want the screen to hide their content, with an option to temporarily show it, or to unblock them. I should be able to look at a list of all the nodes I have blocked, and have the ability to unblock nodes on that list, or look at them to decide if I should unblock them.

# Functional Requirements

## User Cards
- [ ] Each user card should have a block button. When clicked, this button will block that node.

If a node is blocked:
- [ ] all content other than its name should be obscured. 
- [ ] There should be a message stating the node is blocked.
- [ ] There should be a button to temporarily show the node (show the content until the next refresh).

- [ ] If the node is temporarily shown, the block button should instead be an unblock button, allowing the user to unblock that node.

## Listing Cards
- [ ] Each listing card should have a block button. When clicked, this button will block that node.

If a listing's node is blocked:
- [ ] All content other than the node's name should be obscured. 
- [ ] There should be a message stating the listing is blocked.
- [ ] There should be a button to temporarily show the listing (show the content until the next refresh).

- [ ] If the listing is temporarily shown, the block button should instead be an unblock button, allowing the user to unblock that node.

## Search Report Modal
 - [ ] When reporting a listing, there should be an option to send the report and block the node. This would block the node in addition to sending the report. There should also be a button to just report the listing.

## Moderator Cards
- [ ] Each moderator card should have a block button. When clicked, this button will block that node.

If a moderator is blocked:
- [ ] All content other than the node's name should be obscured. 
- [ ] There should be a message stating the listing is blocked.
- [ ]  There should be a button to temporarily show the moderator (show the content until the next refresh).

- [ ] If the moderator is temporarily shown, the block button should instead be an unblock button, allowing the user to unblock that node.

## User Page
- [ ] Each user page should have a block button. When clicked, this button will block that node.

If a user page's node is blocked:
- [ ] All content other than the node's name should be obscured. 
- [ ] There should be a message stating the node is blocked.
- [ ] There should be a button to temporarily show the page (show the content until the next refresh).

- [ ] If the page is temporarily shown, the block button should instead be an unblock button, allowing the user to unblock that node.

## Settings
- [ ] There should be a tab for blocked node management.
- [ ] That tab should have a list of blocked nodes. Each item on the list should:
    - [ ] Show the peer ID of the node.
    - [ ] The ID should be a link that navigates to that user page.
    - [ ] Have an unblock button
QUESTION: does it make sense to have an unblock button in the settings tab, when there is no information available about the blocked node? Would it be better to only have the links to the nodes, so the user can look at the node's page, de-obscure it, and then decide to unblock it or not?

image

# Data

The list of blocked nodes is stored in the settings as "blockedNodes". It is an array of peerIDs.

To update the list, send a PUT with a new value for the blockedNodes to the server. This should be done by setting the new value on the settings model and saving it with type PATCH.

# Notes
- the obscuring of blocked nodes should work in a similar way to obscuring NSFW nodes if the user has opted to not see NSFW content. The buttons to temporarily show NSFW content should be similar to the buttons to temporarily show blocked content.

rmisio:
closes #1062 

known issues:
- Chat messages from blocked users are still going through. My understanding is that the server should be preventing this ([bug here](https://github.com/OpenBazaar/openbazaar-go/issues/820)).


js/models/settings/Settings.js
```diff
@@ -67,4 +67,17 @@ export default class extends BaseModel {
 
     return super.sync(method, model, options);
   }
+
+  parse(response = {}) {
+    if (Array.isArray(response.blockedNodes)) {
+      // de-dupe
+      response.blockedNodes = Array.from(new Set(response.blockedNodes));
```
rmisio: Since the items in a Set must be unique, it's a really cool way to de-dupe an array.

js/views/userPage/Home.js
```diff
@@ -12,11 +12,8 @@ export default class extends BaseVw {
   constructor(options = {}) {
     super(options);
     this.options = options;
-
     this.ownPage = options.ownPage;
-
-    this.userCard = new UserCard({ model: this.model, hideControls: true });
-
+    this.userCard = this.createChild(UserCard, { model: this.model, hideControls: true });
```
rmisio: @jjeffryes Do you know why we are hiding the controls on the user card on the home page?

I understand to not show them if it's your own page, but I'm not sure why we'd otherwise be hiding the Moderator, Blocked and Follow toggles.

jjeffryes: I don't think the design had them on the home page.

@morebrownies should that be revised?

rmisio: @morebrownies any thoughts on this?

morebrownies: Yeah we can show them on the home page too

styles/components/_buttons.scss
```diff
@@ -238,19 +238,28 @@
   &:not(.allowOverflow) {
     overflow: hidden; // clear the floats to avoid layout issues.
   }
-
-  & > .btn {
+  
+  .btn {
```
jjeffryes: I think there was a situation where we had a nested button in some button strips, and that's why it had the > selector on it. 

But I don't think we do any more. I'm just putting this down as a reminder to myself if we notice a problem later.

jjeffryes:
I noticed that if you open a new chat with someone, with no chat messages in it (it's the first time you've opened the chat for that person), and then you block them from inside the chat menu, nothing seems to happen. It successfully blocks them (you can see the button change on their user page), but the chat message area stays open instead of disappearing like it does when you have chat messages.

js/languages/en_US.json
```diff
@@ -26,6 +26,27 @@
...
+  "blockedWarning": {
+    "heading": "You blocked this user",
+    "paragraph1": "This user is currently in your list of blocked users. You can’t view their page or listings while they’re blocked. Blocked users also cannot message you or place orders on your store.",
+    "paragraph2": "Would you like to unblock and view their page?",
+    "btnCancel": "Cancel",
+    "btnUnblock": "Yes, unblock"
+  },
+  "nsfwWarning": {
+    "heading": "Mature content",
+    "paragraph1": "This listing contains photos and/or content that may not be suitable for children or for people in work environments. If you’re sensitive to mature content, please click cancel to return to the previous screen.",
```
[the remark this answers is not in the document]
rmisio: @wolf any objection?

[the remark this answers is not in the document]
morebrownies: Nope. What @jjeffryes suggested makes sense. I like it.

js/views/modals/NsfwWarning.js
```diff
@@ -0,0 +1,63 @@
...
+  onProceedClick() {
+    if (this.getCachedEl('.js-checkboxNsfw').is(':checked')) {
+      this.stopListening(app.settings, null, this.onChangeNsfw);
+      app.settings.set('showNsfw', true);
+      $.ajax({
+        type: 'PATCH',
+        url: app.getServerUrl('ob/settings/'),
+        data: JSON.stringify({ showNsfw: true }),
+        dataType: 'json',
+      });
```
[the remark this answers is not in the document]
rmisio: I was being optimistic and it just doesn't seem worth it to me to handle an error here. All it means is that the client will have `nsfw: true` until the next time the user starts the app, in which case it will come back as `false` when the settings are fetched. Once the user notices that it's false and they want it true, they could try again.

Also, this is local call that doesn't involve any type of communication with other nodes, so it's quite unlikely it will fail.

styles/components/_buttons.scss
rmisio: Oh yeah, sorry... meant to point this one out to you. I didn't want to limit it to immediate children because if you render in a backbone view in there (like I had to), the button will be further nested.

If we do run into the problem you mention, we'll need to come up with some way to manage that case and this case.

rmisio:
> I noticed that if you open a new chat with someone, with no chat messages in it (it's the first time you've opened the chat for that person), and then you block them from inside the chat menu, nothing seems to happen. It successfully blocks them (you can see the button change on their user page), but the chat message area stays open instead of disappearing like it does when you have chat messages.

Good catch! It also did the same behavior with Delete Conversation. I just pushed a fix.

rmisio:
@jjeffryes @morebrownies I've incorporated in all the changes - should be ready for re-review.

Please do your best to not come up with an additional-additional day of work on this one ;)

(jk josh - I know we've sent you on many an extra day after reviewing a PR... lol)

morebrownies:
haha ok will test again now

morebrownies:
all the additional changes look really good to me. Nice work. Thanks for adding shadows to the buttons too :)

PR #1183, opened 2018-01-15
Show 0 ratings

### LINKED ISSUE #1181: Show 0 stars if no reviews on user page
On the user page and listings and listing cards, instead of hiding the reviews if there aren't any, show star: 0 (0).

jjeffryes:
Always show the star rating, even if it's 0 and the count is 0.

Closes #1181

### DIFF
--- a/js/utils/templateHelpers.js
+++ b/js/utils/templateHelpers.js
@@ -55,9 +55,11 @@ export function parseEmojis(text, className = '', attrs = {}) {
  */
 export function formatRating(average, count) {
   const avIsNum = typeof average === 'number';
-  const ratingAverage = avIsNum ? average.toFixed(1) : 0;
-  const ratingCount = typeof count === 'number' ? ` (${count})` : '';
-  return `${parseEmojis('⭐')} ${ratingAverage}${ratingCount}`;
+  const countIsNum = typeof count === 'number';
+  const ratingAverage = avIsNum ? average.toFixed(1) : '?';
+  const ratingCount = countIsNum ? ` (${count})` : '(?)';
+  const error = !avIsNum || !countIsNum ? ` ${app.polyglot.t('ratings.invalid')}` : '';
+  return `${parseEmojis('⭐')} ${ratingAverage}${ratingCount}${error}`;
 }
 
 export const getServerUrl = app.getServerUrl.bind(app);
--- a/js/utils/templateHelpers.js
+++ b/js/utils/templateHelpers.js
@@ -52,14 +52,17 @@ export function parseEmojis(text, className = '', attrs = {}) {
  * If the average is a number, show the last 2 digits and trim any trailing zeroes.
  * @param {number} average - the average rating
  * @param {number} count - the number of ratings
+ * @param {boolean) skipCount = a count wasn't sent, don't show it or test it for validity
  */
-export function formatRating(average, count) {
+export function formatRating(average, count, skipCount) {
   const avIsNum = typeof average === 'number';
   const countIsNum = typeof count === 'number';
   const ratingAverage = avIsNum ? average.toFixed(1) : '?';
-  const ratingCount = countIsNum ? ` (${count})` : ' (?)';
-  const error = !avIsNum || !countIsNum ? ` ${app.polyglot.t('ratings.invalid')}` : '';
-  return `${parseEmojis('⭐')} ${ratingAverage}${ratingCount}${error}`;
+  let ratingCount = countIsNum ? ` (${count})` : ' (?)';
+  if (skipCount) ratingCount = '';
+  const error = !avIsNum || (!countIsNum && !skipCount) ? ' ⚠' : '';
+  const cl = !avIsNum || (!countIsNum && !skipCount) ? 'class="clrTErr"' : '';
+  return `<span ${cl}>${parseEmojis('⭐')} ${ratingAverage}${ratingCount}${error}</span>`;
 }
 
 export const getServerUrl = app.getServerUrl.bind(app);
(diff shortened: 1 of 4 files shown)


js/utils/templateHelpers.js
```diff
@@ -50,16 +50,14 @@ export function parseEmojis(text, className = '', attrs = {}) {
...
- * If the average is invalid or 0, don't show anything.
  * @param {number} average - the average rating
  * @param {number} count - the number of ratings
  */
 export function formatRating(average, count) {
   const avIsNum = typeof average === 'number';
-  const ratingAverage = avIsNum ? average.toFixed(1) : '';
-  const ratingCount = typeof count === 'number' ? ` (${count})` : '';
-  return avIsNum && average > 0 ? `${parseEmojis('⭐')} ${ratingAverage}${ratingCount}` : '';
+  const ratingAverage = avIsNum ? average.toFixed(1) : 0;
```
rmisio: I understand to show 0 if the counts are coming back as 0, but do we really want to show 0 if the counts aren't provided? Isn't that really misleading, and potentially shocking to a vendor, who may have 100 reviews with 5 stars, but for some reason a search provider doesn't provide that field and we end up representing the listing as having no reviews?

morebrownies:
Looks good 👍

jjeffryes:
@rmisio updated not to show invalid counts.

@morebrownies is it fine to show 0 stars if the server returns bad data? I could have it show an error icon or something instead.

morebrownies:
If the serve returns bad data we should display something else, probably something more along the lines of: ⭐  ? (?) failed

jjeffryes:
Question marks will work. The word "Failed" or "Error"?

morebrownies:
`Error` works. Do we need a tooltip to explain the error? Or is there no information on why it failed?

jjeffryes:
There isn't any information why it failed, in that situation the data was loaded somewhere else, so it's not a call failing it's just data formatted wrong.

jjeffryes:
@morebrownies & @rmisio this is updated with the error state.

Mike, there isn't a good way to test it without changing the code. It will replace the number with ? and add the translatable word "Error". so you could get any of the following:
?(4) Error
3(?) Error
?(?) Error

morebrownies:
Sounds good, thanks

rmisio:
I'm not sure if this necessarily needs to be fixed now because it required some extreme data for me to cause the issue, but we do have a potential overflow issue:

image

Some suggestions to mitigate:
1.) Instead of the text `Error` which is variable length depending on the language, we could use an error icon, for example (could even make the icon smaller):

image

2.) If the reviews have a legitimate number, run it through the `abbrNum` util function so large numbers like 10,567 could be shortened to 10.5k or 10k.

rmisio:
If the count is invalid, looks like we have a missing space:

image

Same if both count and average are invalid:

image

jjeffryes:
Good catch on the space.

@morebrownies what do you think about using an icon instead?

I can also make it use the error color class.

morebrownies:
hmm yeah could we possibly display an icon (❗) with a tooltip on it (Failed to load reputation data)?

jjeffryes:
A tooltip gets tricky, since this is a utility function that injects the text into various places. I'd prefer to avoid the complexity of making sure the tooltip overlaps things right everywhere a rating is used.

How about a ⚠ icon instead of a ⭐ and make it red?

Then you'd have 
⚠3.5 (?)
⚠ ? (2)
⚠ ? (?)

morebrownies:
I'd prefer to keep the star and display the warning icon next to the rating. That way it's always clear what the numbers and/or question marks represent. 

⭐ ? (?)❗

rmisio:
FWIW, each of us has thrown out a different icon, but keep in mind this one is used throughout the app:

image

I think the one Mike proposed is ok too, since it would probably conserver even more space and looks really similar to the one we are using.

I wouldn't use the one Josh showed, for consistency, but also because it's not them-able (which is the reason I used the flat color instead of that one in the first place).

(Josh - I do understand you probably just threw it in there as an example)

morebrownies:
I'm fine with that icon @rmisio, consistency would be nice.

jjeffryes:
> (Josh - I do understand you probably just threw it in there as an example)

The icon I pasted is the emoji that gets translated into the one you pasted. So we are actually agreeing.

jjeffryes:
> I'd prefer to keep the star and display the warning icon next to the rating.

Ok. That's how I have it now, I was worried it'd be too many icons close to each other.

jjeffryes:
This is updated again. Screenshot attached of what it looks like when both numbers are invalid.

I added a skipCount parameter, since there is one place where the function is called without a count in the reputation tab. It seemed better than having a second function for just the count.

image

js/utils/templateHelpers.js
```diff
@@ -50,16 +50,19 @@ export function parseEmojis(text, className = '', attrs = {}) {
...
 /**
  * If the average is a number, show the last 2 digits and trim any trailing zeroes.
- * Don't show the count if the count is invalid, x (0) would be inaccurate and confusing.
- * If the average is invalid or 0, don't show anything.
  * @param {number} average - the average rating
  * @param {number} count - the number of ratings
+ * @param {boolean) skipCount = a count wasn't sent, don't show it or test it for validity
  */
-export function formatRating(average, count) {
+export function formatRating(average, count, skipCount) {
```
rmisio: Normally the best practice with options is to us an options object, otherwise if more options come, it could make for a long signature to pass them in individually. Also, if you happen to only want to pass in the last option, you have to throw in a bunch of nulls.

No big deal to change this. More as an FYI for going forward.

jjeffryes: Agreed. I mostly didn't want to go back and change every place it's called, since every other place passes in the first two parameters.

jjeffryes:
Ok, appearance updated again, and the stray unused translation is taken out.

image

@morebrownies does this look good?

morebrownies:
Yeah looks good

PR #1190, opened 2018-01-17
Fix view listing

jjeffryes:
This fixes a weird bug where the override of the open method in the edit listing modal view on line 829 of EditListing.js caused this.editModal to be undefined in line 179 of Listing.js when the modal view was created.

I'm not entirely sure why that was happening, but adding a timeout to the render and open of the modal fixed it.

Removing the override of the open method also fixes it, but we can't just remove that.

### DIFF
--- a/js/utils/modalManager.js
+++ b/js/utils/modalManager.js
@@ -32,14 +32,9 @@ export function launchEditListingModal(modalOptions = {}) {
     }
   }
 
-  const editListingModal = new EditListing(modalOptions);
-  // use a timeout to prevent the editModal from being undefined when the view listing
-  // function is set due to the override of the open method
-  setTimeout(() => {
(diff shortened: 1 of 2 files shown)


rmisio:
@jjeffryes How can I reproduce the problem this solves?

jjeffryes:
@rmisio in master make any edit to a listing, then click the view listing link.

rmisio:
@jjeffryes The issue is that the overridden open isn't returning itself. If you just add a `return this` as the last line, it'll work.

rmisio:
Also `return this` if the exception happens:

```
      try {
        cur = this._origModel.unparsedResponse.listing.metadata.pricingCurrency;
      } catch (e) {
        return this; // added a "this" here
      }
```

PR #1192, opened 2018-01-18
Don't show the refund link if disputed

jjeffryes:
Allowing the vendor to issue a refund while an order is in a disputed state isn't being handled on the server, and shouldn't be an option. 

This PR does not show the refund link when the order is in a disputed state. There will be a separate PR for the server to prevent the refund API from allowing a call when the order is disputed.

### DIFF
--- a/js/views/modals/orderDetail/summaryTab/Summary.js
+++ b/js/views/modals/orderDetail/summaryTab/Summary.js
@@ -72,11 +72,8 @@ export default class extends BaseVw {
         const acceptedState = {
           showFulfillButton: false,
           infoText: app.polyglot.t('orderDetail.summaryTab.accepted.vendorReceived'),
+          showRefundButton: false,
         };
-
-        if (state !== 'DISPUTED') {
-          acceptedState.showRefundButton = false;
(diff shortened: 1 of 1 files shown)


pinheadmz:
Why shouldn't it be an option? What if the moderator goes MIA?

https://www.reddit.com/r/OpenBazaar/comments/7tlmv4/moderator_mia/

jjeffryes:
Because it isn't currently supported on the server, as mentioned above.

pinheadmz:
Client can't just send refund request to server?
https://api.openbazaar.org/#4ae1a3cc-2cb5-6a26-772f-c243304dff59

jjeffryes:
No. That has been disable when the order is in a disputed state.

https://github.com/OpenBazaar/openbazaar-go/pull/835

jjeffryes:
Before it was disabled, allowing a refund while the order was being disputed sent the order into a state of limbo, where the vendor saw it as refunded, but the moderator and buyer saw it as disputed, and none of the parties could move forward.

Clearly that's not something we want to allow.

PR #1194, opened 2018-01-19
Add line for shipping method on order detail modal

### LINKED ISSUE #1193: Order shipping method does not appear on sellers' orders detail page
The fulfilled order shipping method does not display to the seller.

When multiple shipping options are available, the short name of the shipping method selected by the buyer should be clearly shown on the order detail page on the seller side. Otherwise the sell does not know how to ship the item.

ab10460ef3:
Added line for shipping method on order detail modal

Added translations for English

Closes #1193

### DIFF
--- a/js/languages/en_AU.json
+++ b/js/languages/en_AU.json
@@ -1483,6 +1483,8 @@
         "moderatorHeading": "Moderator",
         "copyAddress": "Copy",
         "viewOnMap": "View on map",
+        "shippingOption": "Shipping Option",
+        "shippingService": "Shipping Service",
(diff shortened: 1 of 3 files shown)


jjeffryes:
@ab10460ef3 thank you for catching this, it was an oversight. 

@morebrownies this looks good to me, any thoughts on if the layout needs to be tweaked?

morebrownies:
screenshot 2018-01-19 09 47 15

Yes looks good to me @jjeffryes 

Thank you for adding this @ab10460ef3

PR #1196, opened 2018-01-22
[WIP] Share this store on Twitter

hoffmabc:
Twitter Share Button on Home tab

image

### DIFF
--- a/js/templates/userPage/home.html
+++ b/js/templates/userPage/home.html
@@ -64,6 +64,15 @@
             </div>
           </div>
         <% }); %>
+        <div class="listItem">
+          <div class="flex">
+            <span class="flexExpand">Share this Page</span>
+          </div>
+          <div class="clrT2">
+            <a class="ion-social-twitter" href="https://twitter.com/intent/tweet?text=Check%20out%20<%= ob.name %>%20on%20%23OpenBazaar%20at%20ob://<%= ob.peerID %>">
+             Tweet</a>
+          </div>
+        </div>
       <% } %>
     </div>
   </div>


jjeffryes:
Do we want to design something more robust, with support for multiple social networks, like we had in v1?

@morebrownies ?

morebrownies:
I think that's probably a good idea to put this task through a design phase @jjeffryes and @hoffmabc. The `information` section on the user `home` page is probably very rarely viewed. I would imagine we would want to expose the share button(s) in areas that get more eye traffic? 

For example, I would imagine the most valuable placement for share buttons would be somewhere high up on the listing detail screens.  Users visit listing details a lot more frequently and are more likely to share a listing than a user's `home` page.

hoffmabc:
- I think a design phase would be fine since I'm springing this on you guys. I'm trying to pitch out some easy little tweaks to inspire a few updates here and there when I have a chance.
- The home page is rarely shared, which is true, but the idea is that there is literally nowhere to share out a page via button at the moment, which kind of stinks. I'm supportive of moving it higher up somewhere. 
- I think we probably need to offer both store sharing and listing sharing. I see check out my store way more often than check out my product. I think both are probably needed.

christroutner:
Re: "there is literally nowhere to share out a page via button at the moment"

Have you guys checked out the app on a 1920x1080 monitor? The OB client does not scale to fit different resolutions.  There is a tone of wasted space, and monitor resolution is only going up. I do web site design, which is why I have a laptop such very high definition. 

Next time you're doing some road-mapping, you should checkout Bootstrap 3 or 4, and talk about how you'll approach a mobile-first design. Changes could be adapted iteratively to the Backbone templates over the course of a year.

That would kill two birds with one stone by also moving you towards a smartphone friendly interface. A smartphone UI is a topic I want to talk about with you all further, maybe on the next dev call?

jjeffryes:
@christroutner we know the app isn't responsive. The CSS framework we wrote is responsive ready, we just haven't added any responsive designs yet.

rmisio:
@hoffmabc Since this is been kind of dormant for a while, can we close this PR and potentially re-vive it after we go through the aforementioned design phase?

hoffmabc:
I think you misspelled never. Sure that’s fine.

PR #1207, opened 2018-01-26
Update Trumbowyg

jjeffryes:
- remove links, since they are stripped by the server
- add in some missing options
- fix the list formatting


jjeffryes:
Closing this while investigating if links work or not.

PR #1211, opened 2018-01-29
Don't allow adding moderators without compatible coins

### LINKED ISSUE #1195: Block Adding Moderators with the Wrong Coin
When a user is looking at the user page of a moderator, they can add them as a moderator even if the moderator is running a node that is using a different coin.

The user card also allows this.

The user about page, the user card, and the moderator details view should check the accepted currency of the moderator and user and not allow the user to select an incompatible moderator. Ideally it will just hide the option.

jjeffryes:
This adds a check to make sure users don't add moderators that don't support their current crypto coin.

The moderator button on the user card is hidden if the moderator doesn't support the user's coin.

On the moderator details modal, the moderator's supported currencies are shown, and the add button is replaced with a message stating the user's coin isn't supported by this moderator if that's the case. 

Closes #1195

### DIFF
--- a/js/templates/modals/moderatorDetails.html
+++ b/js/templates/modals/moderatorDetails.html
@@ -44,10 +44,25 @@
       </div>
     </div>
     <div class="contentBox mDetailBox padMd clrP clrBr clrSh3 tx5">
+      <div class="rowMd">
+        <h4><%= ob.polyT('moderatorDetails.currenciesSupported') %></h4>
+        <ul class="unstyled">
+          <% ob.moderatorInfo.acceptedCurrencies.forEach(cur => {
+            const curData = ob.currencyMod.getCryptoCurByCode(cur);
+            if (curData) { %>
+              <li class="flexVCent gutterHSm clrBr">
+                <span class="ion-ios-checkmark-empty clrTEm tx2"></span>
+                <%= ob.currencyMod.cryptoIcon({ code: curData.code }) %>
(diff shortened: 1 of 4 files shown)


js/templates/modals/moderatorDetails.html
```diff
@@ -57,13 +72,19 @@ <h4><%= ob.polyT('moderatorDetails.termsOfService') %></h4>
...
       <div class="flexCol flexHCent gutterV">
         <% if (ob.cardState !== 'selected' && (!ob.ownMod || ob.purchase)) { %>
-          <button class="btn clrP clrBr js-addAsModerator">
-            <% if (ob.purchase) { %>
-              <%= ob.polyT('purchase.chooseModerator') %>
-            <% } else { %>
-              <%= ob.polyT('moderatorDetails.addAsModeratorButton') %>
-            <% } %>
-          </button>
+          <% if (ob.moderatorInfo.acceptedCurrencies.indexOf(ob.currencyMod.getServerCurrency().code) !== -1) { %>
```
rmisio: For this check and the one below, we can't just use getServerCurrency().code because that doesn't take testnet into account and we'll end up with no mods being available on testnet.

It would have something like:

```
ob.moderatorInfo.acceptedCurrencies.includes(
  ob.currencyMod.getServerCurrency().isTestnet ?   ob.currencyMod.getServerCurrency().testnetCode :   ob.currencyMod.getServerCurrency().code
)
```

jjeffryes: Good catch, thanks.

PR #1217, opened 2018-02-01
Remove Moderated Order Price Limit

jjeffryes:
This removes the code to prevent moderation on orders based on the estimated fee. That will now be handled by the server when the purchase is attempted.

This PR also has extensive refactoring of the purchase code to use our current methods of setting state. 

The original code was written some time ago, I reapplied it manually to master because merging it was a mess, this should be checked carefully to make sure nothing old accidentally was written over something new (I've checked several times myself).

### DIFF
--- a/js/views/baseVw.js
+++ b/js/views/baseVw.js
@@ -250,9 +250,9 @@ export default class baseVw extends View {
       newState = _.extend({}, this._state, state);
     }
 
-    if (opts.renderOnChange && !_.isEqual(this._state, newState)) {
+    if (!_.isEqual(this._state, newState)) {
       this._state = newState;
-      this.render();
+      if (opts.renderOnChange) this.render();
     }
 
     return this;
--- a/js/views/modals/purchase/Payment.js
+++ b/js/views/modals/purchase/Payment.js
@@ -170,7 +170,7 @@ export default class extends BaseVw {
 
   get amountDueLine() {
     return app.polyglot.t('purchase.pendingSection.pay',
-      { amountBTC: formatCurrency(this.balanceRemaining, 'BTC') });
+      { amountBTC: formatCurrency(this.balanceRemaining, getServerCurrency().code) });
   }
 
   get qrDataUri() {
--- a/js/views/modals/purchase/Payment.js
+++ b/js/views/modals/purchase/Payment.js
@@ -67,8 +67,8 @@ export default class extends BaseVw {
   set balanceRemaining(amount) {
     if (amount !== this._balanceRemaining) {
       this._balanceRemaining = amount;
-      this.$amountDueLine.html(this.amountDueLine);
-      this.$qrCodeImg.attr('src', this.qrDataUri);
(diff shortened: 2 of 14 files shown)


js/views/baseVw.js
```diff
@@ -250,9 +250,9 @@ export default class baseVw extends View {
       newState = _.extend({}, this._state, state);
     }
 
-    if (opts.renderOnChange && !_.isEqual(this._state, newState)) {
+    if (!_.isEqual(this._state, newState)) {
       this._state = newState;
-      this.render();
+      if (opts.renderOnChange) this.render();
```
jjeffryes: This fixes a bug we discussed where the state is not updated if renderOnChange is false.

js/views/modals/purchase/Purchase.js
```diff
@@ -105,14 +111,17 @@ export default class extends BaseModal {
       selectFirst: true,
       radioStyle: true,
     });
+    // render the moderators so it can start fetching and adding moderator cards
+    this.moderators.render();
+    this.moderators.getModeratorsByID();
```
jjeffryes: The moderators view is being rendered here because getModeratorsByID will add the moderators to the DOM, which fails if it isn't rendered yet. We don't want to call that in the render of this view or it would refetch all the moderators on each render.

jjeffryes:
Temporarily closing for a few fixes.

jjeffryes:

js/views/modals/purchase/Payment.js
```diff
@@ -84,8 +84,8 @@ export default class extends BaseVw {
     return {
       'click .js-payFromWallet': 'clickPayFromWallet',
       'click .js-payFromAlt': 'clickPayFromAlt',
-      'click .js-copyAmount': 'copyAmount',
-      'click .js-copyAddress': 'copyAddress',
+      'click .js-amountRow': 'copyAmount',
+      'click .js-addressRow': 'copyAddress',
```
jjeffryes: The click action was changed to the containing element so all parts of it are clickable.

js/views/modals/purchase/Payment.js
```diff
@@ -145,23 +145,23 @@ export default class extends BaseVw {
   copyAmount() {
     clipboard.writeText(String(this.balanceRemaining));
 
-    this.$copyAmount.addClass('active');
+    this.getCachedEl('.js-copyAmount').addClass('active');
     if (this.hideCopyAmountTimer) {
       clearTimeout(this.hideCopyAmountTimer);
     }
     this.hideCopyAmountTimer = setTimeout(
```
rmisio: Can we clear this timer in remove?

rmisio:
image

We should probably get an error code from the server on this one so that we could translate.

I'm not sure why we have to keep asking for that. IMHO, any messages which could become user facing use have a code in addition to the text version. If it's coming from a lower level lib and that's problematic... ok, but I'm quite sure that's not the case with this one.

jjeffryes:
Error code request here https://github.com/OpenBazaar/openbazaar-go/issues/855

PR #1219, opened 2018-02-05
Replaced Bitcoin with cryptocurrency

SamPatt:
After the 2.1.0 release I updated the Terms of Service to say "cryptocurrency" instead of "Bitcoin."


rmisio:
looks like this is already updated in master.

PR #1230, opened 2018-02-09
Add Japanese locale

camelmasa:
I added Japanese locale 😃 

Almost words were translated by auto-translate service.
And then, I checked those words on development environment. Looks good.

It's not perfect for now. But, I hope those will be improved by other Japanese contributors.

openbazaar


jjeffryes:
@camelmasa we use Transifex for all our translations, if you want to help with translations can you ask to join there as a translator on the Japanese translation resource? It's already 40% translated. 

https://www.transifex.com/ob1/openbazaar

When we do a release, we copy all the latest translations for the languages we currently support, plus any new ones that have reached 80% completion or more, from Transifex. 

If we merged this PR, it would be overwritten the next time we pulled the files from Transifex. It's much better to make any changes there.

camelmasa:
@jjeffryes Thanks for comment 😃 Got it. I'll join Transifex when I have time.

PR #1234, opened 2018-02-16
fixing the positioning of the spend estimate fee confirmation tip

rmisio:
before:

screen shot 2018-02-16 at 10 02 07 am

screen shot 2018-02-16 at 10 49 28 am

after:

image

image

### DIFF
--- a/js/views/ListingCard.js
+++ b/js/views/ListingCard.js
@@ -372,7 +372,7 @@ export default class extends baseVw {
   remove() {
     if (this.fullListingFetch) this.fullListingFetch.abort();
     if (this.destroyRequest) this.destroyRequest.abort();
-    $(document).off(null, this.boundDocClick);
+    $(document).off('click', this.boundDocClick);
     super.remove();
   }
 
--- a/styles/modules/modals/_wallet.scss
+++ b/styles/modules/modals/_wallet.scss
@@ -226,7 +226,7 @@
 
   &.centeredBelow {
     // The container that you want the spendConfirmBox to be
-    // centered below must be relatively positioned.
+    // centered below must be non-statically positioned.
     .sendConfirm {
       margin-top: 10px;
       position: absolute;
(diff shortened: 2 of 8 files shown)


js/views/ListingCard.js
```diff
@@ -372,7 +372,7 @@ export default class extends baseVw {
   remove() {
     if (this.fullListingFetch) this.fullListingFetch.abort();
     if (this.destroyRequest) this.destroyRequest.abort();
-    $(document).off(null, this.boundDocClick);
```
rmisio: Didn't realize jQuery's off requires this param. It's optional on Backbone's version, in which case the given handler would be removed for all event types.

styles/modules/modals/_wallet.scss
```diff
@@ -224,8 +215,23 @@
...
+      position: relative;
+      top: -10px;
+    }
+  }
+
+  &.centeredBelow {
+    // The container that you want the spendConfirmBox to be
+    // centered below must be non-statically positioned.
+    .sendConfirm {
+      margin-top: 10px;
```
jjeffryes: I don't think you want the 10px margin here. On my Windows computer it pushes the tooltip down too far:

image

Without the margin it looks like this:
image

jjeffryes: The margin is needed in the order details though.

It looks like purchase.scss changes the top value to calc(100% + 10px), you could either take that out and use the 10px margin instead, or apply the same top value in the purchase and in the order details.

rmisio:
@jjeffryes Didn't we take the stance we're not going to ry and be pixel perfect to zoomed environments?

jjeffryes:
> Didn't we take the stance we're not going to try and be pixel perfect to zoomed environments?

This isn't related to zooming, there's an extra 10px being applied to the tooltip in the purchase screen for all OSes.

PR #1223, opened 2018-02-07
Order escrow timeout

rmisio:
Implements the [UI for the 45 day escrow timeout](https://github.com/OpenBazaar/openbazaar-desktop/issues/923):

Currently this runs best with this server branch: ` 843-test-w-short-aging-timeouts` (it is tweaked so escrow related notifications come in a matter of minutes - instead of days)

(as server bugs are fixed, they often happen in a bunch of different branches, so you will likely want to either ask the server devs to merge "in review" branches into the above branch or ask them to identify the branches and merge them yourself)

PR #1237, opened 2018-02-23
Order error state

### LINKED ISSUE #1202: Show Orders with an ERROR state
This is dependent on an enhancement to the server https://github.com/OpenBazaar/openbazaar-go/issues/838

## User Story

The server will sometimes put an order into an ERROR state and not provide it to the client. As a user, I want to know when my orders are in in an error state, so I know they happened and don't think the application is just broken or missing data.

As a user, when I see an order in an error state, I want some guidance on why it happened and what to do to resolve it.

As a seller, if an order has been rejected due to the listing being updated and a stale listing being purchased, I would prefer to have the order go through so I can choose whether to honor it.

## Functional Requirements

- [ ] When an order is stale due to an out of date listing being purchased, the order should work as normal. No indication needs to be given that the order was out of date.
- [ ] When an order has a state of ERROR_PENDING, an "Error" status will be shown.
    - [ ] The buyer can cancel the order the same as if it was in a PENDING state, if it is not moderated.
    - [ ] If the order is moderated, the buyer will have to dispute it to retrieve their funds.
    - [ ] The seller cannot take any action on an order in the ERROR_PENDING state.
    - [ ] The server will return a error_reason parameter, it should be shown in the Order Details.

rmisio:
closes #1202


js/views/modals/orderDetail/OrderDetail.js
```diff
@@ -526,8 +529,12 @@ export default class extends BaseModal {
     let showDisputeOrderButton = false;
 
     if (this.buyerId === app.profile.id) {
+      // TODO todo ToDo !!! TODO todo ToDo !!! TODO todo ToDo !!!
+      // todo: when escrow timeout code is ready, include this in the timeoutInfoView
+      // also ensure order is funded if processing error
       showDisputeOrderButton = this.moderatorId &&
```
rmisio: I'll have to circle back to this once #1223 is merged (waiting on server changes) since some of the stuff related to the dispute button was refactored.

I didn't want to merge that PR into this one and have it be some huge diff report.

js/views/modals/orderDetail/summaryTab/Summary.js
```diff
@@ -743,6 +791,46 @@ export default class extends BaseVw {
...
+      isBuyer,
+      isModerator: !!(this.moderator && this.moderator.id),
+      isOrderCancelable: this.isOrderCancelable,
+      isModerated: !!this.moderator,
+      isCase: this.isCase,
+      // TODO todo ToDo !!! TODO todo ToDo !!! TODO todo ToDo !!!
+      // TODO todo ToDo !!! TODO todo ToDo !!! TODO todo ToDo !!!
+      // TODO todo ToDo !!! TODO todo ToDo !!! TODO todo ToDo !!!
+      // TODO: factor in escrow timeout
+      isDisputable: isBuyer &&
```
rmisio: Same as above comment about #1223. Once that one is merged, I'll need to tweak this to account for the escrow timeout.

js/views/modals/orderDetail/contractTab/ContractTab.js
```diff
@@ -53,19 +53,33 @@ export default class extends BaseVw {
...
+        msg = `<p class="clrTEm flexVCent">${icon}${msgHtml}</p>`;
       } else if (!this.model.get('vendorContract')) {
-        msg = '<p class="flexVCent"><span class="ion-android-warning margRSm clrTAlert">' +
-          `</span>${app.polyglot.t('orderDetail.contractTab.vendorContractNotArrived')}</p>`;
+        const icon = `<span class="${iconBaseClass} clrTAlert ion-android-warning"></span>`;
+        const processingErrorKey =
+          'orderDetail.contractTab.vendorContractNotArrivedPotentialProcErr';
+        const buyerContract = this.model.get('buyerContract');
+        const buyerShowsVendorProcErr =
+          buyerContract && Array.isArray(buyerContract.get('errors'));
```
jjeffryes: Does the array check here need to check length, or is an empty array not possible?

js/models/order/Order.js
```diff
@@ -38,6 +38,16 @@ export default class extends BaseModel {
...
   }
 
+  /**
+   * Returns a boolean indicating whether the vendor had an error when processing
+   * the order. This is different from just checking for the PROCESSING_ERROR state,
+   * in that it will return true even after the order moves on from that state.
+   */
+  get vendorProcessingError() {
+    const contract = this.get('contract');
+    return contract && Array.isArray(contract.get('errors'));
```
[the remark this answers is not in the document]
rmisio: > Or does the error object have a value of undefined when there are no errors?

Yes, the error Array will only be added to the contract when there is a processing error.

js/views/modals/orderDetail/contractTab/ContractTab.js
rmisio: Same as above, based on what the current server behavior is (and CP confirmed that we should be able to rely on this behavior) the error field is never added to the contract until the point the vendor has an error processing the order. At that point, the Array will be added in and it will remain in the contract, which is important because it allows us to know whether there was a processing error even after the state moves on.

I don't know if there will ever be a situation where there is an empty array, but if there is that would be considered a processing error without a reason, which the client handles with generic messaging.

PR #1240, opened 2018-02-27
Verified mods

### LINKED ISSUE #1123: Verified Moderators
## User Story

- As a buyer, when browsing listings I want a way to see which listings have verified moderators so I can make informed choices when looking at listings.
- As a buyer, when looking at a listing, I want to know if I can pick a verified moderator when I purchase it, so I can make an informed choice whether to click the buy button.
- As a buyer, when I am purchasing a listing, if it has moderators I want to have the verified moderators indicated clearly, so I can easily choose one of them if I want to.
- As a buyer, when I am purchasing a listing, I should see a clear and very noticeable explanation of what a verified moderator is, and why I should choose one. The explanation should make the risks of not choosing a verified moderator clear, so I am warned that purchasing from vendors with no verified moderators comes at a risk.
- As a vendor, I want a way to show buyers browsing listings that my products use verified moderators, so they will be more likely to buy from me.
- As a vendor, I want buyers looking at my listing to see I have verified moderators, so they trust my listing more and are more likely to click the buy button.
- As a vendor, I want buyers to be able to see the verified moderators in the purchase process, so they can select one if they want and avoid selecting a risky moderator that will result in a bad experience. 
- As a vendor,  when I am selecting moderators for my store, I want to be able to easily find verified moderators. They should be clearly labeled, and I should have the ability to see them together, either first in the list or sorted.
- As a moderator, I want a way to show buyers and sellers that they can trust me when choosing a moderator.
- As a user, I want to have the default list downloaded automatically.
- As a user, I want to be able to change the list to a different list at a http or IPFS location.
- As a user, if I change the list I want an easy way to reset it to the default list.

## Functional Requirements
- [x] The client should call an http or IPFS address after starting to download the list of verified moderators.
- [x] The results should include a graphic to show (as a badge or something similar)
- [x] There should be a default or loading image to show while the badge is still loading.
- [x] The list should refresh periodically.
- [ ] On startup, the list should load from a cache instead of holding up startup.
- [ ] Show a loading message at the bottom of the page, letting the user know the list is loading and the current results are out of date (maybe show how old the list is). 
- [ ] Critical pages should show a warning if the current list is old or no list has loaded yet.

### Listing Card
- [x] (tentative) The listing card should show whether the listing has one or more moderator on the verified list.

### Listing Details
- [x] The listing screen should show whether the listing has one or more moderators on the verified list.

### Purchasing
- [x] There should be an explanation/warning letting users know why verified moderators are important, and what the risks are of purchasing without one.
- [x] Verified moderators should be clearly indicated. Maybe they should be shown first, or there should be a way to filter/sort by verified or not (TBD).
- [x] Show placeholders when loading so the moderators don't jump around when loading.
- [x] Show error states if a moderator doesn't load.
- [ ] Allow the user to hide errored moderators.
- [x] There should be a "show only verified" button that only shows verified moderators.
- [x] The button should be active by default.
- [x] It should be clear you can see the other moderators by de-activating the filter.

### Settings/Store
- [x] There should be an explanation/warning letting vendors know why verified moderators are important, and why they should choose them.
- [x] Verified moderators should be clearly indicated. Maybe they should be shown first, or there should be a way to filter/sort by verified or not (TBD).
- [x] Show placeholders when loading so the moderators don't jump around when loading.
- [x] Show error states if a moderator doesn't load.
- [ ] Allow the user to hide errored moderators.
- [x] There should be a "show only verified" button that only shows verified moderators.
- [x] The button should be active by default.
- [x] It should be clear you can see the other moderators by de-activating the filter.

### Settings/Advanced
- [ ] There should be an input for the list url.
- [ ] The default url should be intput.
- [ ] There should be a link or button to restore the default, so the user can recover it if they changed it to something else.

### Moderator Details
- [x] If the moderator is verified, it should be clearly indicated.

### Order Details
- [x] If the moderator is verified, it should be clearly indicated.

### User Card
- [x] If the user is a verified moderator, it should be clearly indicated

## Data
- the list should have an array of peerIDs.
- the list should include a link to the image for the badge
- if we want to show listings with verified moderators, the moderators would need to be included in search results and listings data. They are already in the data for individual listings and purchasing.

### LINKED ISSUE #1218: Show Controls on Errored UserCards
When user cards fail to load, the controls aren't shown, which makes it impossible to unfollow a node that can't be reached.

The usercard should show the appropriate controls when it fails to load.

jjeffryes:
This adds a verified moderators system to the client.

The endpoint is currently fixed, there is a commented out section in the settings for changing it we can make visible when other providers of endpoints exist.

Closes #1123 
Closes #1218


js/models/LocalSettings.js
```diff
@@ -50,7 +50,7 @@ export default class extends Model {
       language,
       listingsGridViewType: 'grid',
       bitcoinUnit: 'BTC',
-      searchProvider: 'https://search.ob1.io/search/listings',
```
jjeffryes: This was no longer used.

js/start.js
```diff
@@ -274,6 +277,39 @@ function onboard() {
   return onboardDeferred.promise();
 }
 
+function fetchVerifiedMods() {
```
jjeffryes: This function is run on a loop to update the list of verified moderators once an hour. I'm open to suggestions if there's a better way to do the loop, and/or to incorporate the verified moderator fetch into the startup sequence below.

js/start.js
```diff
@@ -305,44 +346,63 @@ function fetchStartupData() {
...
+      setInterval(() => fetchVerifiedMods(), 360000);
     })
     .fail((jqXhr) => {
       const curConn = getCurrentConnection();
 
-      if (!jqXhr || !curConn || curConn.status !== 'connected') {
+      // Check to see if the server is not connected. the search providers and verified moderators
+      // can return an undefined jqXhr if their external endpoints are down while the app itself is
+      // still functional.
+      if ((ownFollowingFailed || walletBalanceFetchFailed) &&
```
jjeffryes: If the search providers fail to load from localStorage or the verified moderators fail to load from an external source, this logic shouldn't trigger, since the node might be running normally and can proceed without the search providers or the verified moderators.

js/start.js
```diff
@@ -305,44 +346,63 @@ function fetchStartupData() {
...
       }
 
-      if (ownFollowingFailed || walletBalanceFetchFailed || searchProvidersFetchFailed) {
+
+      if (ownFollowingFailed || walletBalanceFetchFailed || searchProvidersFetchFailed ||
+        verifiedModsFetchFailed) {
         let title = '';
+        let msg = jqXhr && jqXhr.responseJSON && jqXhr.responseJSON.reason || '404';
+        let btnText = app.polyglot.t('startUp.dialogs.btnManageConnections');
+        let btnFrag = 'manageConnections';
```
jjeffryes: The search and verified fail states should have a continue button instead of a manage connections button, since neither are dependent on the connection to the node.

js/templates/modals/settings/general.html
```diff
@@ -128,6 +128,23 @@ <h2 class="h3 clrT"><%= ob.polyT('settings.generalTab.sectionName') %></h2>
           </div>
         </div>
       </div>
+      <div class="flexRow gutterH TODO">
```
jjeffryes: This is hidden since I wrote the code then we decided not to include it. When/if another verified moderator provider exists we could turn this back on.

js/views/components/Moderators.js
```diff
@@ -84,22 +89,60 @@ export default class extends baseVw {
...
+      initialState: {
+        mode: opts.method === 'POST' ? 'loading' : 'loaded',
+        showLoadBtn: opts.showLoadBtn,
+        showSpinner: opts.showSpinner,
+      },
+    });
+    this.listenTo(this.moderatorsStatus, 'browseMore', () => this.onBrowseMore());
+
+    // listen to the websocket for moderator data
+    this.serverSocket = getSocket();
```
jjeffryes: Moved out of the getModeratorsByID because that can be called multiple times.

js/views/components/Moderators.js
```diff
@@ -115,55 +158,62 @@ export default class extends baseVw {
     const buyerCur = app.serverConfig.cryptoCurrency;
     const modCurs = profile.moderatorInfo && profile.moderatorInfo.acceptedCurrencies || [];
     const validCur = modCurs.includes(buyerCur);
-    // if the moderator is on the list of IDs to exclude, or is this user, remove them
-    const excluded = this.excludeIDs.includes(profile.peerID) || profile.peerID === app.profile.id;
```
jjeffryes: Handling excluded peerIDs was moved to the .done handler for getModeratorsByID, since it makes more sense not to process excluded IDs at all.

jjeffryes:
@morebrownies one thing to look at is the list view of the store page, I made an educated guess for where the badge should go.

morebrownies:
Ok, I'll walk through the PR later today.

morebrownies:
Here's some feedback @jjeffryes 

**1. Settings > General is broken at the bottom of the screen**

screenshot 2018-02-28 13 13 52

**2. Can the Learn more link be on the same line as the rest of the text?**
screenshot 2018-02-28 13 14 23

**3. Text color #777777 appear to light on the orange background. I believe in the designs I'm using black text here.**
screenshot 2018-02-28 13 16 01

**4. I'm not seeing the Verified badge show up on the user card**
screenshot 2018-02-28 13 44 03


Checkout experience from a design perspective is looking good. I haven't completed a order yet and viewed the order details screen yet. Anyone have a store setup with verified moderator that I can order from?

jjeffryes:
@morebrownies if you look at a verified moderator's user card, it shows the badge.
ob://QmVFNEj1rv2d3ZqSwhQZW2KT4zsext4cAMsTZRt5dAQqFJ/home
ob://QmeSyTRaNZMD8ajcfbhC8eYibWgnSZtSGUp3Vn59bCnPWC/home

jjeffryes:
@morebrownies The learn more link is on a new line because the description text is sent from the verified moderator server, and could be 100s of lines, so I have it clamped to 3 lines with an ellipses if it goes over that.

If we wanted to let it be as big as other providers want, I could change the link to be inline with it. That's a little more fragile as it'd be up to other providers (I think one is being built now) not to make that text too long.

morebrownies:
> @morebrownies if you look at a verified moderator's user card, it shows the badge.ob://QmVFNEj1rv2d3ZqSwhQZW2KT4zsext4cAMsTZRt5dAQqFJ/homeob://QmeSyTRaNZMD8ajcfbhC8eYibWgnSZtSGUp3Vn59bCnPWC/home

I'm not seeing it on my own card when I go to user pages that follow me. Maybe that's where the issue is?

morebrownies:
> @morebrownies The learn more link is on a new line because the description text is sent from the verified moderator server, and could be 100s of lines, so I have it clamped to 3 lines with an ellipses if it goes over that.

If we wanted to let it be as big as other providers want, I could change the link to be inline with it. That's a little more fragile as it'd be up to other providers (I think one is being built now) not to make that text too long.

Not a big deal either way. I didn't realize we were planning to support additional copy in the tooltip. Not sure when that type of functionality will actually be relevant.

js/views/modals/Settings/Store.js
```diff
@@ -11,74 +10,74 @@ import { openSimpleMessage } from '../SimpleMessage';
 
 export default class extends baseVw {
   constructor(options = {}) {
-    super({
+    const opts = {
       className: 'settingsStore',
       ...options,
-    });
+      initialState: {
```
rmisio: I don't think you want to use the setState functionality here. It doesn't really play nice with forms since it triggers a full re-render and existing form data will be lost. For example, put a peerId in the Find text field and then check/uncheck the Verified Mods filter. You'll lose that peer id. It's just way too heavy handed to re-render this entire store view.

The setState approach also generally doesn't play well with views that modify their state via jQuery (like this one). You really want to pick a path and either go the jQuery approach or setState approach. For example, if you saved this tab and the save was taking a while and you check/uncheck the Verified Mods filter, you'd lose the "processing" state of the Save button.

As far as handling the `showVerifiedOnly`, you'd want to do that in a way in which it just reloads the available moderators and doesn't trigger a full re-render of everything else since this view wasn't structured in a way to re-create it's state on re-renders.

(dead horse alert... without a virtual dom to help us to only selectively re-render specific parts that change, it's a fine game between using a jQuery / setState approach and/or being much more granular with your sub views.)

js/templates/modals/orderDetail/modFragment.html
```diff
@@ -9,3 +9,4 @@
   <% } %>
   <div><%= ob.name %><% print(modLink ? ` ${modLink}` : '') %></div>
 </div>
+<div class="js-verifiedMod"></div>
```
rmisio: image

1.) That text doesn't really make sense in this context. You can't choose a verified moderator at this point. Perhaps the tooltip isn't needed here at all?
2.) Looks like some spacing is missing between that error icon and the text.
3.) Can we wrap the user name content and the verified mod content in a div and give it gutterVTn class so there is some vertical spacing between the name and the verified mod text?
4.) Instead of the emoji, can we use `ion-alert-circled clrTAlert`. It's a.) theme-able b.) consistent with what we use in the rest of the app and it's used a good amount. (I know the emoji might match the designs better, but everytime I bring up the consistency issue with @morebrownies, he opts for consistency over being 100% true to the designs).

js/views/components/ModeratorsStatus.js
```diff
@@ -0,0 +1,55 @@
...
+      'click .js-browseMore': 'clickBrowseMore',
+    };
+  }
+
+  clickBrowseMore() {
+    this.trigger('browseMore');
+  }
+
+  remove() {
+    clearTimeout(this.loadingTimer);
```
rmisio: Do you mean `this.spinnerTimout`?

js/views/components/ModeratorsStatus.js
```diff
@@ -0,0 +1,55 @@
...
+    super.remove();
+  }
+
+  render() {
+    loadTemplate('components/moderatorsStatus.html', (t) => {
+      this.$el.html(t({
+        ...this.getState(),
+      }));
+      if (this._state.showSpinner && this._state.spinnerTimer) {
```
rmisio: On any state change render is called, so with this code you'd be potentially setting multiple timeouts which turn off the spinner. I think you'd want to clear the timeout at the start of render() to avoid that.

js/collections/VerifiedMods.js
```diff
@@ -0,0 +1,71 @@
...
+       Embed the type in each moderator so it's easier to use elsewhere. It should look like:
+       peerID: string,
+       type: {
+         name: string,
+         description: string,
+         badge: url string,
+         }
+     */
+    parsedResponse.forEach((mod) => {
+      mod.type = {};
```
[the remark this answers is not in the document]
jjeffryes: Yes, this was a last minute change that broke things, I reverted it yesterday but didn't push it.

jjeffryes:
@rmisio this is ready for review. I found a weird issue in the start fetches we should probably discuss, where if one fetch fails on a retry the successful fetches still have arguments and look like they failed.

js/views/modals/purchase/Purchase.js
```diff
@@ -192,39 +200,38 @@ export default class extends BaseModal {
...
-
   onNoValidModerators() {
     this.order.moderated = false;
-    this.setState({ noValidModerators: true });
+    this.render();
+  }
+
+  togVerifiedModerators(bool) {
+    // if hiding unverified mods, deselect the selected mod if it is unverified and switch to
+    // direct payment
```
rmisio: Hmmm... instead of kicking them back to a direct payment, should we maybe select nothing?

My concern is the user selects a mod and they may have no idea that toggling a filter kicks them back to a direct payment and they may unknowingly purchase something directly when they thought they had a mod (but maybe where just curious and wanted to see the list of verified mods).

Would love to hear @morebrownies thoughts on this when he comes back.

jjeffryes: The interaction here is pretty confusing, I agree. If you select an unverified mod and then hide the unverified mods, it looks like no mod is selected, so we don't want to keep the hidden mod selected if you click purchase.

But like you said, the user may not realize the unverified mod they selected was de-selected. 

Since the UI uses radio buttons (even if in the HTML they aren't really radio buttons) I think we should stick with how radio buttons work, which means one should always be selected. Or we could change to checkboxes or something else.

I could also see disabling the ability to hide unverified moderators if you have an unverified moderator selected as an option. Maybe that's the simplest?

js/views/modals/Settings/Store.js
```diff
@@ -144,7 +141,7 @@ export default class extends baseVw {
   }
 
   onClickVerifiedOnly(e) {
```
rmisio: Would this be better as a 'change' instead of a 'click' handler?

js/views/components/ModeratorsStatus.js
```diff
@@ -42,9 +42,10 @@ export default class extends BaseVw {
       this.$el.html(t({
         ...this.getState(),
       }));
+      if (this.spinnerTimeout) clearTimeout(this.spinnerTimeout);
```
rmisio: As an fyi, clearTimeout handles pretty much anything being passed to it, so you really don't need that if.

PR #1241, opened 2018-03-01
refactor the fail handler of the whenall startup data fetch

rmisio:
@jjeffryes I was misunderstanding how $.whenAll passes in args to its fail handler. This is a refactor based on the correct understanding.

Basically, let's say I call `$.whenAll(a, b, c, d)` and let's say b and d fail, then the fail handler will be called with `(undefined, [jqXhr, textStatus, errorThrown], undefined, [jqXhr, textStatus, errorThrown])`.

So, if at least one call fails, the fail handler will be called and it will be passed an array that maps to the original array passed into whenAll. For any call that didn't fail, `undefined` will be in it's corresponding array slot. For any call that failed, their array slot will have the arguments that would otherwise go to jquery's `error` handler.

FWIW, this merges into your verifiedMods branch.

### DIFF
--- a/js/start.js
+++ b/js/start.js
@@ -312,76 +312,62 @@ function fetchVerifiedMods() {
 
 const fetchStartupDataDeferred = $.Deferred();
 let ownFollowingFetch;
-let ownFollowingFailed;
 let exchangeRatesFetch;
 let walletBalanceFetch;
-let walletBalanceFetchFailed;
 let searchProvidersFetch;
-let searchProvidersFetchFailed;
 let verifiedModsFetch;
-let verifiedModsFetchFailed;
 
 function fetchStartupData() {
-  ownFollowingFetch = !ownFollowingFetch || ownFollowingFailed ?
+  ownFollowingFetch = !ownFollowingFetch || ownFollowingFetch.state() === 'rejected' ?
     app.ownFollowing.fetch() : ownFollowingFetch;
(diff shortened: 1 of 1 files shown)


js/start.js
```diff
@@ -312,76 +312,62 @@ function fetchVerifiedMods() {
...
+      const ownFollowingFailArgs = args[0];
+      const exchangeRatesFailArgs = args[1];
+      const walletBalanceFailArgs = args[2];
+      const searchProvidersFailArgs = args[3];
+
+      // Find any that failed aside from the exchangeRateFetch. We don't care if the
+      // exchange rate fetch failed, because the exchangeRateSyncer will display a
+      // status message about it and the app will gracefully handle not having exchange
+      // rates.
+      const failed = args.filter(failArgs => failArgs && failArgs !== exchangeRatesFailArgs);
```
jjeffryes: Is there any chance one of the other calls could fail at the same time as the exchange rates call, and this would discard it because the data is coincidentally the same as the exchange rate fail data?

Could it be done more explicitly with something like args.splice(1,1).filter?

rmisio: > Is there any chance one of the other calls could fail at the same time as the exchange rates call, and this would discard it because the data is coincidentally the same as the exchange rate fail data?

No, it wouldn't be discarded. The only discarded one is the exchange rate one. Any other one would be returned as true by the filter and satisfy the `if (failed.length)` condition below.

PR #1246, opened 2018-03-05
Center purchase box [DO NOT MERGE BEFORE 1240]

### LINKED ISSUE #1235: Purchase confirmation tooltip needs centering
Observed:

image

Expected:
The arrow is pointing to the center of the button.

jjeffryes:
This is based on #1240, which made some changes to the arrowbox classes. Once 1240 is merged, the change set for this should be very small.

Closes #1235

### DIFF
--- a/js/templates/modals/purchase/actionBtn.html
+++ b/js/templates/modals/purchase/actionBtn.html
@@ -15,7 +15,7 @@
     </button>
   <% } %>
   <% if (ob.confirmOpen) { %>
-    <div id="confirmPay" class="confirmBox arrowBoxTop clrBr clrP clrT clrSh1 js-confirmPay">
+    <div id="confirmPay" class="confirmBox arrowBoxCenteredTop clrBr clrP clrT clrSh1 js-confirmPay">
       <div class="flexColRows gutterVSm padLg">
         <h3>
           <%= ob.polyT('purchase.confirmPayment.title') %>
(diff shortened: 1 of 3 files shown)


rmisio:
@jjeffryes I think this may have been introduced by the verified mods work... the icons used to be on the top left. Where they intentionally moved?

image

Unless the designs changed, I think the edit ones should be on the top left, whereas the report/block are bottom left.

image

jjeffryes:
@rmisio yes, @morebrownies moved the icons in his design, because the badge is now where the report button used to be.

rmisio:
@jjeffryes I'll merge since I don't want to hold up the release, but @morebrownies, are we sure we want the edit buttons to be centered whereas the report ones are left align (see screenshot above)? Just wondering if that inconsistency is intentional?

morebrownies:
Is there a reason to center them? I had them left aligned in the designs. 

screenshot 2018-03-06 09 14 49

rmisio:
> Is there a reason to center them? I had them left aligned in the designs.

@jjeffryes ^^^ ?

Also, is there a reason the verified mod top arrow is not centered, like in the designs? I think it's more problematic for it to be like it is below in case a listing card is ever placed near the edge of the screen (there's a lot of over-hang if you're pushing it that far to the left).

image

jjeffryes:
> Also, is there a reason the verified mod top arrow is not centered, like in the designs? 

Because it overlaps the chat on the right if it's centered, I think it might have ended up with the right end off the screen. 

I can adjust it to be more centered, I'll have to take a look at what happens.

Re: edit buttons, I thought I had it matching the designs, either it was and some other CSS change altered it, or I misread the designs. I can adjust it.

jjeffryes:
Now I remember. I think the badge was on the right side of the listing at some point, which caused the overlap, and I didn't update the tooltip when I moved the badge to the left.

PR #1248, opened 2018-03-05
fix issue where listing inventory is not saved on first save

### LINKED ISSUE #1244: Inventory doesn't seem to work
I'm on Windows 10 with the latest release. 

Reproduce:

1. Create or Clone a listing.
2. Fill out the form as it is displayed.
3. Click the dropdown for 'Inventory Management' and select 'Track Inventory'
4. Press tab and enter a number. Press tab to leave the box or click out of it.
5. Click Save.

Quantity will reset to 0.

Moreover, when an item with quantity of 1 is sold, this listing still appears purchasable to visitors. It may prevent the purchase from going through, I'm not 100% sure.

rmisio:
closes #1244

### DIFF
--- a/js/views/modals/editListing/EditListing.js
+++ b/js/views/modals/editListing/EditListing.js
@@ -862,7 +862,7 @@ export default class extends BaseModal {
 
     // If the inventoryManagement has been rendered, we'll let it's drop-down
     // determine whether we are tracking inventory. Otherwise, we'll get the info
-    // form the model.
+    // from the model.
     if (this.inventoryManagement) {
       trackBy = this.inventoryManagement.getState().trackBy;
     } else {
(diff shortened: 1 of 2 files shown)


js/views/modals/editListing/EditListing.js
```diff
@@ -893,14 +893,13 @@ export default class extends BaseModal {
     const excludes = '.js-sectionShipping, .js-couponsSection, .js-variantsSection, ' +
       '.js-variantInventorySection';
 
-    return this._$formFields ||
```
rmisio: The issue was that the quantity field is being dynamically rendered in after making a changing in its adjacent drop down and the form fields selector is already cached before that. Just going to turn off the cache for now.

jjeffryes:
@rmisio can you fix the lint error in InventoryManagement? It doesn't use the _ any more. 

Otherwise this looks good to go.

PR #1257, opened 2018-03-08
Addresses need to be URL encoded before being used for map lookup

EchterAgo:
The view on map link would not work with my address. The reason for this is because my address includes a slash. The patch URL encodes the address before being used.


rmisio:
thanks @EchterAgo!

PR #1258, opened 2018-03-08
Postion list view buttons

jjeffryes:
This will correctly position the block and edit buttons in list view mode on listing cards, and fixes a visual bug where the empty wrapper for the report button had a margin and pushed the block button out of place when no report button was being shown.


js/views/ListingCard.js
```diff
@@ -405,6 +405,8 @@ export default class extends baseVw {
       this.listenTo(this.reportBtn, 'startReport', this.startReport);
       this.getCachedEl('.js-reportBtnWrapper').append(this.reportBtn.render().el);
     }
+    // hide unneeded report buttons so their wrapper doesn't affect the button layout
+    this.getCachedEl('.js-reportBtnWrapper').toggleClass('hide', !this.reportsUrl);
```
rmisio: Would it be better to just style the wrapper to hide itself when't empty? `.wrapper:empty { display: none }`

The advantage to that is if you ever need to fill it based on some event, it would be ready for you (even though that's probably not needed in this case). Also saves a dom write.

We could even create a generic `hideIfEmpty` class, since this pattern is used a time or two in the code already.

styles/components/_listingCard.scss
```diff
@@ -292,6 +287,14 @@
       display: none;
     }
 
+    .additionalOverlay,
+    .editOverlay {
+      height: 100%;
+      box-sizing: border-box;
+      align-items: center;
+      padding-left: 13px; // position the first button in the middle of the thumbnail
```
rmisio: Is this padding super necessary? When I toggle it, it only adjusts things by a few pixels and I'm not noticing a difference?

PR #1259, opened 2018-03-09
Mod tweaks

rmisio:
- Slight tweak to the global link handler because it's previous incarnation would result in links not working if you stopped propagation on any element which contains links. With the tweak, anytime you have such an element you should be sure to explicitly call handleLinks and pass in that element.

@jjeffryes Not only had this been an issue on the listing card (which you fixed), I also found it to be an issue on the moderator component. Since propogation was being prevented on that, the learn more link was not working on the verified mod tip within it.

- Tweak to how verified mods are added to the moderators component. Previously they were being added to the top as they came in. The problems with that were:

a.) A few load, you want to select one of them and suddenly the one you want to select jumps down further as a new one comes in.
b.) The first one was being auto selected, but after a few more loaded, it would no longer be the first one. It looked random to just have the, for example, 5th one be selected by default.
 
To fix the two issues, verified mods will be added after the last verified mod. So, they still will collectively remain on top, it's just they won't be added to the tip-top as they come in.

### DIFF
--- a/js/views/components/ModeratorCard.js
+++ b/js/views/components/ModeratorCard.js
@@ -3,9 +3,9 @@ import loadTemplate from '../../utils/loadTemplate';
 import app from '../../app';
 import Profile from '../../models/profile/Profile';
 import VerifiedMod from './VerifiedMod';
+import { handleLinks } from '../../utils/dom';
 import { launchModeratorDetailsModal } from '../../utils/modalManager';
 import { getLangByCode } from '../../data/languages';
-
 
 export default class extends BaseVw {
   constructor(options = {}) {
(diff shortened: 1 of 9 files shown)


js/views/components/ModeratorCard.js
```diff
@@ -36,6 +36,8 @@ export default class extends BaseVw {
     if (!this.model || !(this.model instanceof Profile)) {
       throw new Error('Please provide a Profile model.');
     }
+
+    handleLinks(this.el);
```
rmisio: If you're going to be stopProgogation'ing on an element, you would want to pass that element into handleLinks so any links will still be handled.

This could be on any element. Doesn't need to be the main view element. It just happens to be slightly more convenient if you are stopping propagation on the main view element, because then you don't have to call handleLinks() in render.

PR #1260, opened 2018-03-09
fix deleting unloaded mods

jjeffryes:
If mods load very slowly, saving was deleting unloaded mods. This PR changes the code to only remove loaded mods that have been de-selected.

### DIFF
--- a/js/views/modals/Settings/Store.js
+++ b/js/views/modals/Settings/Store.js
@@ -1,4 +1,5 @@
 import $ from 'jquery';
+import _ from 'underscore';
 import app from '../../../app';
 import loadTemplate from '../../../utils/loadTemplate';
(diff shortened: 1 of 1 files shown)


js/views/modals/Settings/Store.js
```diff
@@ -149,7 +150,9 @@ export default class extends baseVw {
   }
 
   getSettingsData() {
-    const selected = this.modsSelected.selectedIDs;
+    let selected = app.settings.get('storeModerators');
+    // The mods may not have loaded in the interface yet. Subtract only explicitly de-selected ones.
+    selected = _.without(selected, ...this.modsSelected.unselectedIDs);
```
rmisio: This could also be accomplished just as concisely with the native Array.filter().

PR #1262, opened 2018-03-09
remove extra asterisk, center text, no shrink on verified

jjeffryes:
This makes the layout in the moderator details card a little nicer when the fee amount is very large, and removes an extra asterisk.

**Before:**
image

**After:**
image


rmisio:
@jjeffryes Since all three of these columns have variable width content, can we put some type of max-width to prevent one from dominating the others (or flat out break the layout)?

image

PR #1268, opened 2018-03-14
marged

liray-unendlich:


jjeffryes:
Hi @liray-unendlich. Do you want to tell us about your project?

liray-unendlich:
sorry, i miss submitted pull request.

jjeffryes:
No worries. We're just curious what you're working on. It's always great to see new people working on the OpenBazaar platform.

It looks like you're working on a version of OB that uses Phore. Would it be easier to just make an SPV wallet and submit it to our repo, so everyone in the network could have the option of using PHR, instead of restricting it to a forked client and server?

PR #1271, opened 2018-03-15
Add data-open-external attribute to verified mods

### LINKED ISSUE #1266: Add a data-open-external attr to verified mod Learn More links
This will force them to be opened as external links. Without that, they'll still be opened externally as long as the proper protocol is provided, but since these urls are provided by an outside party, we shouldn't assume they'll be in a certain format.

jjeffryes:
Adds a data-open-external to the link in the verified moderator tooltip, to make sure it opens externally.

Closes #1266


rmisio:
I fixed this in the pending pr #1269

jjeffryes:
I just saw this is in #1269

jjeffryes:
Jinx. :)

PR #1277, opened 2018-03-16
Opt In Anonymous Metrics tracking

jjeffryes:
This integrates an opt-in metrics option using Countly.

### DIFF
--- a/js/router.js
+++ b/js/router.js
@@ -174,6 +174,16 @@ export default class ObRouter extends Router {
       return false;
     }
 
+    if (window.Countly) {
+      console.log(name)
+      window.Countly.q.push(['add_event', {
+        key: 'asyncButtonClick',
+        segmentation: {
+          id: name,
+        },
+      }]);
+    }
+
     this.navigate(this.standardizedRoute(), { replace: true });
 
     // We'll iterate through any open modal which have a confirmClose method
--- a/js/router.js
+++ b/js/router.js
@@ -523,7 +523,7 @@ export default class ObRouter extends Router {
       new Transactions({ initialTab }).render()
     );
 
-    recordEvent('TransactionsPageLoad', {
+    recordEvent('Transactions_PageLoad', {
       tab: initialTab,
     });
   }
--- a/js/router.js
+++ b/js/router.js
@@ -174,16 +174,6 @@ export default class ObRouter extends Router {
       return false;
     }
 
-    if (window.Countly) {
-      console.log(name)
-      window.Countly.q.push(['add_event', {
-        key: 'asyncButtonClick',
-        segmentation: {
-          id: name,
-        },
-      }]);
-    }
-
     this.navigate(this.standardizedRoute(), { replace: true });
 
     // We'll iterate through any open modal which have a confirmClose method
--- a/js/router.js
+++ b/js/router.js
@@ -52,6 +52,7 @@ export default class ObRouter extends Router {
 
     $(window).on('hashchange', () => {
       this.setAddressBarText();
+      window.Countly.q.push(['track_pageview', location.pathname + location.hash]);
     });
 
     ipcRenderer.on('external-route', (e, route) => {
--- a/js/router.js
+++ b/js/router.js
@@ -499,9 +499,9 @@ export default class ObRouter extends Router {
         this.off(null, onWillRoute);
         endEvent('UserPageLoad', {
           ownPage: guid === app.profile.id,
-          pageState,
+          tab: pageState,
           listing: !!listingFetch,
-          error: userPageFetchError,
+          errors: userPageFetchError || 'none',
         });
       });
   }
--- a/js/router.js
+++ b/js/router.js
@@ -16,7 +16,7 @@ import TemplateOnly from './views/TemplateOnly';
 import Profile from './models/profile/Profile';
 import Listing from './models/listing/Listing';
 import BlockedWarning from './views/modals/BlockedWarning';
-import { recordEvent } from './utils/metrics';
+import { startEvent, endEvent } from './utils/metrics';
 
 export default class ObRouter extends Router {
   constructor(options = {}) {
@@ -408,8 +408,9 @@ export default class ObRouter extends Router {
     let profileFetch;
     let listing;
     let listingFetch;
-    const userPageFetchTime = Date.now();
     let userPageFetchError = '';
+
+    startEvent('UserPageLoad');
 
     if (guid === app.profile.id) {
       // don't fetch our own profile, since we have it already
@@ -496,8 +497,7 @@ export default class ObRouter extends Router {
     })
       .always(() => {
         this.off(null, onWillRoute);
-        recordEvent('UserPageLoad', {
-          timeInSeconds: (Date.now() - userPageFetchTime) / 1000,
+        endEvent('UserPageLoad', {
           pageState,
           listing: !!listingFetch,
           error: userPageFetchError,
--- a/js/router.js
+++ b/js/router.js
@@ -16,7 +16,7 @@ import TemplateOnly from './views/TemplateOnly';
 import Profile from './models/profile/Profile';
 import Listing from './models/listing/Listing';
 import BlockedWarning from './views/modals/BlockedWarning';
-import { startEvent, endEvent } from './utils/metrics';
+import { startEvent, endEvent, recordEvent } from './utils/metrics';
 
 export default class ObRouter extends Router {
   constructor(options = {}) {
@@ -516,9 +516,15 @@ export default class ObRouter extends Router {
       this.navigate('transactions/sales');
     }
 
+    const initialTab = tab || 'sales';
+
     this.loadPage(
-      new Transactions({ initialTab: tab || 'sales' }).render()
+      new Transactions({ initialTab }).render()
     );
+
+    recordEvent('TransactionsPageLoad', {
+      tab: initialTab,
+    });
   }
 
   connectedPeers() {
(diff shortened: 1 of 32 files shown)


rmisio:
image

Looks likes some padding is needed to account for the scrollbar. Also, the Decline button is disabled, which makes for a tricky opt out ;)

rmisio:
Should we make the data no-closable, other than via Decline or Share, forcing a decision to be made?

```
      dismissOnEscPress: false,
      showCloseButton: false,
```

js/utils/metrics.js
```diff
@@ -0,0 +1,111 @@
...
+let metricsRestartNeeded = false;
+
+export function isMetricRestartNeeded() {
+  return metricsRestartNeeded;
+}
+
+export function addMetrics() {
+  function loadMetrics() {
+    // Reverse the countly opt out
+    window.localStorage.setItem('cly_ignore', 'false');
```
rmisio: What is this purpose of this? It doesn't seem to be used anywhere. Is this a flag that countly is reading?

js/utils/metrics.js
```diff
@@ -0,0 +1,111 @@
...
+    metricsRestartNeeded = !!window.Countly;
+
+    //  If Countly has already been added, it won't run again until the app is restarted.
+    if (window.Countly) return;
+
+    const sVer = app.settings.get('version');
+    const serverVersion = sVer.substring(sVer.lastIndexOf(':') + 1, sVer.lastIndexOf('/'));
```
rmisio: Looks like this is also duplicated in the about modal. Can we instead add a method to the settings model to compute this...maybe a `prettyServerVer()` getter?

js/views/modals/purchase/Payment.js
```diff
@@ -99,7 +100,15 @@ export default class extends BaseVw {
...
         fetchError: 'ERROR_INSUFFICIENT_FUNDS',
       });
+      recordEvent('payFromWallet', {
+        currency: getServerCurrency().code,
+        sufficientFunds: false,
+      });
     } else {
+      recordEvent('payFromWallet', {
+        currency: getServerCurrency().code,
+        sufficientFunds: true,
```
rmisio: At this point, it doesn't necessarily mean there are sufficient funds. It just means the client can't confirm that there are. Based on the number of inputs there might not be. The estimate fee api will determine it definitively.

js/views/modals/purchase/Payment.js
```diff
@@ -114,15 +123,27 @@ export default class extends BaseVw {
...
       spend({
         address: this.paymentAddress,
         amount: this.balanceRemaining,
-        currency: getServerCurrency().code,
+        currency,
       })
+        .done(() => {
+          recordEvent('walletSpend', {
+            currency,
+            err: '',
```
rmisio: We seem to be using error in other countly events. Can we be consistent and either use "err" or "error" consistently? I think it would make for a cleaner review of the countly data.

js/views/modals/purchase/Purchase.js
```diff
@@ -322,13 +323,18 @@ export default class extends BaseModal {
 
     this.setState({ phase: 'processing' });
 
+    const segmentation = { error: 'None' };
```
rmisio: This is another place we're not so consistent. In most place success is defined as an empty error string, whereas here we went with 'None'.

js/views/modals/purchase/Purchase.js
```diff
@@ -348,24 +354,33 @@ export default class extends BaseModal {
             this.listenTo(this.payment, 'walletPaymentComplete',
               (pmtCompleteData => this.completePurchase(pmtCompleteData)));
             this.$('.js-pending').append(this.payment.render().el);
+            segmentation.id = data.orderId;
```
rmisio: Is it ok to track orderId's?

@SamPatt @morebrownies

morebrownies:
Hey @jjeffryes I finally had a chance to check out this PR. Here's a few pieces of feedback:

I find the term 'share' to be strange here. I think it would be better suited to offer 'Decline' or 'Accept' actions.
screenshot 2018-04-26 10 22 57

Once I clicked 'Share' I expected the window would close automatically on click. Having to manually close after hitting 'share' felt unnecessary. 
screenshot 2018-04-26 10 25 47

Other than that, looks good to me. 👍

js/views/modals/purchase/Purchase.js
morebrownies: Not sure @rmisio, seems like a business decision to me.

js/utils/metrics.js
jjeffryes: That's a countly flag. Their code assumes if you opt out you can't change your mind, we have to change the flag to re-activate the tracking if the user wants to turn it back on.

js/views/modals/purchase/Purchase.js
jjeffryes: I am taking it out, we thought we could use it to connect state change events in orders so we could see which percentage of them made it through what steps, but even though countly has the data I haven't found a way to use it usefully.

js/router.js
```diff
@@ -485,11 +490,21 @@ export default class ObRouter extends Router {
...
       }
     })
-      .always(() => (this.off(null, onWillRoute)));
+      .always(() => {
+        this.off(null, onWillRoute);
+        endEvent('UserPageLoad', {
+          ownPage: guid === app.profile.id,
+          tab: pageState,
+          listing: !!listingFetch,
+          errors: userPageFetchError || 'none',
```
jjeffryes: Countly doesn't record segments with blank values, so any event with an error will record "none" if it's successful. Otherwise you can't show a comparison of events with errors vs. events without errors.

js/start.js
```diff
@@ -622,6 +631,9 @@ function start() {
               }
             });
           }
+
+          // add the feedback mechanism
+          addFeedback();
```
jjeffryes: The feedback was moved because previously it was created before the profile and settings were available, leaving all of that data unreported.

js/templates/modals/metricsModal.html
```diff
@@ -0,0 +1,41 @@
...
+        <% } %>
+      </div>
+    <% } %>
+    <% if (ob.showUndecided) { %>
+      <a class="btn clrP clrBr js-decline"><%= ob.polyT('metrics.declineBtn') %></a>
+      <a class="btn clrP clrBr js-share"><%= ob.polyT('metrics.acceptBtn') %></a>
+    <% } else { %>
+      <a class="btn clrP clrBr js-decline <% if (!ob.shareMetrics) print('disabled') %>"><%= ob.polyT('metrics.deactivateBtn') %></a>
+      <a class="btn clrP clrBr js-share <% if (ob.shareMetrics) print('disabled') %>"><%= ob.polyT('metrics.activateBtn') %></a>
+    <% } %>
```
jjeffryes: If the modal is opened for the first time, the buttons will be "Decline" and "Accept", and the modal won't have a close button.

If the modal is opened from settings, the buttons will be "Deactivate" and "Activate", and will be disabled as appropriate. The modal will have a close button so the user can close it without changing anything.

js/router.js
rmisio: I'm not going to hold up merge for this, but I think it would have been more efficient to centralize this error: 'none' handling in the metric util module. That way you only write it once instead of every caller to endEvent, recordEvent, etc... having to follow this convention.

For example (metrics.js):

```
function parseSegmentation(segmentation = {}) {
  return {
    error: segmentation.error || 'none';
  }
}

export function endEvent(key, segmentation = {}) {
  const seg = parseSegmentation(segmentation);
  ...
}
```

Then instead of having to repeat something like this:

```
endEvent('Purchase', {
  errors: 'none',
 });
```

you could simplify to:

```
endEvent('Purchase');
```

js/start.js
```diff
@@ -583,7 +583,16 @@ function start() {
           }
 
           localStorage.serverIdAtLastStart = curConn && curConn.server && curConn.server.id;
-          Backbone.history.start();
+
+          if (app.localSettings.get('shareMetrics') === undefined) {
+            showMetricsModal({
```
rmisio: You also need `dismissOnEscPress: false` here.

js/start.js
rmisio: Definitely not holding up merge for this... just musing...

Now that it's moved, doesn't that mean a user can't give feedback during the startup portion of the app (which could be lengthy)? I would have thought it would be fine to allow feedback earlier and if we have the profile and settings, they are sent over, otherwise not. I'm assuming that if the user didn't submit feedback until after they became available, then they would be sent?

js/views/modals/Settings/advanced/Advanced.js
```diff
@@ -242,7 +243,6 @@ export default class extends baseVw {
   }
 
   render() {
-    super.render();
```
rmisio: This render should probably be up top. It's on modals that it needs to be after the content is added to `this.el`.

js/router.js
jjeffryes: That's a good idea, though we might not always want an error segment on events. If that happened we could have a endEventNoError function to separate out the different use case.

rmisio: > we could have a endEventNoError function to separate out the different use case

I dunno if that's the best approach because then you would need a recordEventNoError function and maybe more.

I'm not sure I get the 'no error' use case when I thought the whole point of this was to make sure that the error is standardized. What would be a scenario where `error: none` would not work and it would be critical to not have the error key?

Anyways... we don't have to worry about it now.

PR #1280, opened 2018-03-20
[i18n/es] "Home" -> s/Casa/Inicio

gubatron:
We don't use the word "Casa" to denote "Home" screens, we call them "Inicio" (or Start)

### DIFF
--- a/js/languages/es.json
+++ b/js/languages/es.json
@@ -264,7 +264,7 @@
     "email": "Correo electrónico",
     "noLocation": "No hay ubicación",
     "tabTitles": {
-      "home": "Casa",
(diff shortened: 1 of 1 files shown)


jjeffryes:
I've changed this on https://www.transifex.com/ob1/openbazaar

We pull the language files from there when we update them, which will overwrite any other changes we've made in this repo. 

@gubatron if you could make changes in Transifex, that would help us out. Thank you for spotting this.

gubatron:
Will do next time, wasn't aware. Cheers

jjeffryes:
@gubatron cool, thank you.

PR #1283, opened 2018-03-21
Account for no moderators in listings

jjeffryes:
This fixes an issue where it's possible to create a listing with no value for moderators, which would throw an error due to the matched function expecting an array.


hoffmabc:
Awesome, thanks!

PR #1284, opened 2018-03-22
Verified mods tor

### LINKED ISSUE #1275: Add Tor Endpoint for Verified Moderators
When the client is using Tor, use http://my7nrnmkscxr32zo.onion/verified_moderators for the Verified Moderators endpoint.

jjeffryes:
If the app is in Tor mode, use the new Tor endpoint for the verified moderators.

Closes #1275

### DIFF
--- a/js/collections/VerifiedMods.js
+++ b/js/collections/VerifiedMods.js
@@ -19,7 +19,8 @@ export default class extends Collection {
   }
 
   url() {
-    return app.localSettings.get(`verifiedModsProvider${getCurrentConnection().server.get('useTor') ? 'Tor' : ''}`);
+    const usingTor = getCurrentConnection().server.get('useTor');
+    return app.localSettings.get(`verifiedModsProvider${usingTor ? 'Tor' : ''}`);
   }
 
   get data() {
--- a/js/collections/VerifiedMods.js
+++ b/js/collections/VerifiedMods.js
@@ -18,7 +18,8 @@ export default class extends Collection {
   }
 
   url() {
-    return app.localSettings.get('verifiedModsProvider');
+    const usingTor = app.serverConfig.tor && getCurrentConnection().server.get('useTor');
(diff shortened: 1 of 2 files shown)


js/models/LocalSettings.js
```diff
@@ -51,6 +51,7 @@ export default class extends Model {
       listingsGridViewType: 'grid',
       bitcoinUnit: 'BTC',
       verifiedModsProvider: 'https://search.ob1.io/verified_moderators',
```
rmisio: Is the idea that if the provider doesn't provide a seperate tor url, the standard url will be filled in for both urls?

js/collections/VerifiedMods.js
```diff
@@ -18,7 +19,8 @@ export default class extends Collection {
   }
 
   url() {
-    return app.localSettings.get('verifiedModsProvider');
+    const usingTor = app.serverConfig.tor && getCurrentConnection().server.get('useTor');
```
rmisio: I think you could eliminate the `app.serverConfig.tor` part there since it doesn't really matter if the server is using tor or not because this call wouldn't go through the server. `getCurrentConnection().server.get('useTor')` indicates that the user chose to use tor in their server config and that IMHO should be enough for us to route the verified mods call through a tor endpoint.

js/collections/VerifiedMods.js
```diff
@@ -18,7 +19,7 @@ export default class extends Collection {
   }
 
   url() {
-    return app.localSettings.get('verifiedModsProvider');
+    return app.localSettings.get(`verifiedModsProvider${getCurrentConnection().server.get('useTor') ? 'Tor' : ''}`);
```
rmisio: Is this line too long?

jjeffryes: Yes it is. I switched editors and didn't notice it had linting turned off and a different line length.

PR #1288, opened 2018-03-25
Add support for Zcoin wallet

t3rm1n4l:


t3rm1n4l:
This change depends on https://github.com/OpenBazaar/openbazaar-go/pull/899

rmisio:
We will not be merging the corresponding server PR until it uses an SPV wallet because relying on a full node is something we've found 99% of users won't do and/or don't have the technical skill level to do. If the server PR is updated to support an SPV wallet, please feel free to resubmit this PR.

Thanks for the the interest in contributing to OB!

PR #1294, opened 2018-04-02
Set app window size to 90% of screen size instead of fixed 1200x760

pinheadmz:
I don't know if this bothers anybody else but the first thing I do when I open OB is resize the window! I'm running on OSX, and unlike most other OSX apps, OB doesn't "remember" if its window size was manually adjusted the last time it was run. I find the default 1200x760 too small, it cuts off the bottom of the first row of listings (see screen shot) and I personally just like every app I interact with to use as much screen real estate as possible. I find this 90% setting much more satisfying to interact with, and still leave a margin around the app to click the desktop, etc. 

screen shot 2018-04-02 at 11 48 52 am

### DIFF
--- a/main.js
+++ b/main.js
@@ -1,7 +1,7 @@
 import {
   app, BrowserWindow, ipcMain,
   Menu, Tray, session, crashReporter,
-  autoUpdater, shell, dialog,
+  autoUpdater, shell, dialog, screen,
 } from 'electron';
 import homedir from 'homedir';
 import { argv } from 'yargs';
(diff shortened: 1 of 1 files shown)


pinheadmz:
Here's the "after" with window size at 90% of screen size... NICE!

screen shot 2018-04-02 at 11 52 45 am

jjeffryes:

I don't know if we want to set it to scale for all users, since it was designed for a particular width. 

Maybe we could add in whatever is needed to have the app remember it's last size?

jjeffryes:
It looks like there isn't a built in way to remember the last window stat in Electron. There are a few libraries out there that could help.

https://github.com/mawie81/electron-window-state
https://github.com/Sethorax/electron-window-state-manager

PR #1298, opened 2018-04-04
Lazy Load Verified Mods

jjeffryes:
This will change the verified moderators call to a lazy loading pattern. All the affected views will listen for updates to the verified moderators collection, and refresh any parts that need to. 

This will account for the verified moderators taking a while to load, and for any changes to the list that happen between loads.

Loading the verified moderators is no longer blocking startup of the rest of the app. If the users first view has verified moderator badges, they may see them pop in after it loads.

### DIFF
--- a/js/views/ListingCard.js
+++ b/js/views/ListingCard.js
@@ -1,5 +1,4 @@
 import $ from 'jquery';
-import _ from 'underscore';
 import app from '../app';
 import loadTemplate from '../utils/loadTemplate';
 import { openSimpleMessage } from '../views/modals/SimpleMessage';
--- a/js/views/ListingCard.js
+++ b/js/views/ListingCard.js
@@ -1,4 +1,5 @@
 import $ from 'jquery';
+import _ from 'underscore';
 import app from '../app';
 import loadTemplate from '../utils/loadTemplate';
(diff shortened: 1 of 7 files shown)


js/views/modals/purchase/Purchase.js
```diff
@@ -174,7 +184,14 @@ export default class extends BaseModal {
           buildRefreshAlertMessage(app.polyglot.t('purchase.purchaseDataChangedPopin')),
       });
 
-      this.listenTo(this.dataChangePopIn, 'clickRefresh', () => (this.render()));
+      this.listenTo(this.dataChangePopIn, 'clickRefresh', () => {
```
rmisio: IMHO, it doesn't really work well to pass in a modsChanged flag because multiple data changes may have caused the pop up to show. For example, the localCurrency may have changed and then a minute later, the verified mods might have changed.

I think we could simplify by making refresh re-render both things, e.g:

```
      this.listenTo(this.dataChangePopIn, 'clickRefresh', () => {
          this.render();
          this.moderators.render();
      });
```

js/views/ListingCard.js
```diff
@@ -1,4 +1,5 @@
 import $ from 'jquery';
+import _ from 'underscore';
```
rmisio: is this needed?

PR #1300, opened 2018-04-05
Fix layout, disallow re-selecting invalid mods.

### LINKED ISSUE #1297: Action buttons should be disabled/hidden for an invalid moderator
If a mod is no longer a mod in Store settings, the View and Checkmark buttons should be disabled/hidden.

image

(maybe related to this, but that Loading Moderators text never goes away. Wondering if the invalid mod is messing up some counts?)

### LINKED ISSUE #1299: Don't let mod card name bleed into buttons
image

jjeffryes:
This PR cleans up the layout of the mod card so the name is shown on invalid mods and does not overlap the buttons, removes the view button on invalid mods, and changes the select button so when a mod is invalid, the user can only de-select them (to remove them from their store), and not re-select them.

Closes #1297 
Closes #1299

### DIFF
--- a/js/templates/components/moderatorCard.html
+++ b/js/templates/components/moderatorCard.html
@@ -1,7 +1,9 @@
 <%
-   // check to see if the card was created with at least minimum data, not just a peerID, which would indicate a server error.
+   // Check to see if the card was created with at least minimum data, not just a peerID, which would indicate a server error.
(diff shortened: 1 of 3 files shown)


rmisio:
Looks like this one is incorporated in #1306.

PR #1302, opened 2018-04-06
Catch Socket Data Before REST Returns in Moderators

jjeffryes:
This PR moves the socket listener to before the REST call to get moderators is made, and checked incoming messages for peerId and other data instead of relying on the socket id returned by the REST call.

This prevents issues with the socket returning data before the socket ID is known, which causes the moderator in the socket data not to be added.

### DIFF
--- a/js/views/components/Moderators.js
+++ b/js/views/components/Moderators.js
@@ -126,8 +126,7 @@ export default class extends baseVw {
     });
     this.listenTo(this.moderatorsStatus, 'browseMore', () => this.onBrowseMore());
 
-    // listen to the websocket for moderator data
-    this.serverSocket = getSocket();
+    this.listenForIDs = [];
   }
 
   events() {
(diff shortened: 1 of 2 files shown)


js/views/components/Moderators.js
```diff
@@ -204,34 +226,30 @@ export default class extends baseVw {
...
+        this.listenForIDs.push(...IDs);
+
+        /* Listen to the websocket for moderators, the data may arrive before the POST that triggers
+           it returns. It's possible socket data for profiles will arrive from some other call,
+           that's fine as long as the socket responses have valid data.
+        */
+
+        const serverSocket = getSocket();
+        if (serverSocket) {
+          this.listenTo(serverSocket, 'message', this.onSocketMessage);
```
rmisio: I think how this is done is really error prone. A couple big issue for me are:

- every time you call getModeratorsByID with an op.async option it's binding a new server socket listener. If you call it 10 times, you'll suddenly have 10 handlers doing the same thing, which at the very least is very inefficient.
- If you call it a second time while a first call is happening, the first call essentially gets blown away since an instance wide `listenForIDs` is being set. There's no self containment to this function.

Could these issue be cleared up if we asked the server to refactor the endpoint allow the client to optionally provide a `socketId` that the server would return with the socket data (including errors). So the server would only generate the id if it wasn't provided?

With that in place, the client could go back to only processing data that's for the socketId it's interested in. IMHO, it better suits concurrent and multiple calls.

Another possible optimization with that approach would be to unbind a handler when all the id's it is interested in have returned.

jjeffryes:
I talked to @cpacia and he thinks it should be fairly easy to add an optional socket ID parameter to the POST call, which would be a better way to fix this issue.

https://github.com/OpenBazaar/openbazaar-go/issues/925

Issue #383, opened 2017-03-07
Add support for Fedora, and/or Flatpak for all distros


heyakyra:
Fedora users will not be able to use the "Linux" download provided. It is a bit misleading. https://github.com/OpenBazaar/openbazaar.org/issues/64

There should be .rpm files for them, and Flatpak could be used to support all GNU/Linux distributions.

hoffmabc:
I think we specify that we only officially support Ubuntu, but I can see how this is misleading. I don't know if we will add RPMs but we'll put this in the queue.

hoffmabc:
Oh also I think we're looking at Snap already as well.

thekyriarchy:
Oh cool. Snap is unfortunately an Ubuntu project and they're known to not always play nice with the rest of the distro community (duplicate effortts so that they have full control). Comparing the adoption rate of each speaks for itself i'd say: https://kamikazow.wordpress.com/2017/02/09/adoption-of-flatpak-vs-snap/

hoffmabc:
We have two Canonical team members in our Slack @elopio and @popey who might have some thoughts on this as well.

popey:
🔮 Spell of summoning worked!  ✨ 

I can appreciate that people have different opinions on Canonical and snaps. The fact is we're working hard to get first class snap support in a lot of distros right now. That's not an easy job, especially given the various rules and policies we have to adhere to across them all. Zygmunt Krynicki - our lead developer in this area - wrote some thoughts on it recently. Worth a read for a fuller picture. https://new.zygoon.pl/post/state-of-snapd-support-across-distros/ 

Personally I have snaps running on Ubuntu 14.04, 16.04, 16.10, (32 and 64-bit), Ubuntu Core on ARM (Raspberry Pi), OpenSUSE Leap, KDE Neon, Linux Mint, Zorin OS, Peppermint OS and Antergos (Arch). Some of those are of course Ubuntu derivatives, but in total they all cover quite a huge chunk of the Linux-using market. More so if you consider that Ubuntu alone has the lions share of the cloud and server space.

Of course none of these packaging systems are mutually exclusive. It's entirely possible to have snaps, RPM, Deb and Appimage. It's just a case of where your priorities lie, and how much of your finite engineering effort you're willing to expend to get there. 

With snaps, once you have done the up-front work to get a working snap package, we have the build and store infrastructure to easily distribute them on every update to an audience of millions, which for many is a compelling argument.

elopio:
[...]

PR #1303, opened 2018-04-10
Let's run in docker

arteniioleg:
Just execute `./run-in-docker.sh`

Star this project https://github.com/mviereck/x11docker That guy is doing an awesome job!

_Related to https://github.com/OpenBazaar/openbazaar-desktop/issues/383_

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,19 @@
+FROM ubuntu:17.10
+
+ARG VERSION=2.1.1
+ARG TINI_VERSION=0.17.0
+
+RUN apt-get update \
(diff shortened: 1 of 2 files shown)


rmisio:
@tyler-smith I'm not super familiar with Docker. Do you want to check this PR out and make sure it looks ok?

One of the things I'd want to be sure about is... is this implementation useful to a wide enough net of potential users that it makes sense merging into the repo?

rmisio:
Going to close this for now. At this point, we're not sure what we want our Docker integration to look like, or if we want one. We may circle back to this.

PR #1306, opened 2018-04-17
Settings Store Fixes

### LINKED ISSUE #1297: Action buttons should be disabled/hidden for an invalid moderator

jjeffryes:
This fixes an issue where invalid mods weren't being handled properly, which would cause the loading status to never complete.

This also prevents error messages meant only for the purchase view from being displayed when a moderator isn't found, and fixes an issue with a wrong class name that didn't hide the load by ID div when the ID couldn't be found (this caused a collapsed div with a border to appear when it shouldn't).

NOTE: This incorporates #1300, that should be merged first, or closed after this PR is merged.

Closes #1297

### DIFF
--- a/js/templates/components/moderators.html
+++ b/js/templates/components/moderators.html
@@ -6,27 +6,30 @@
       </div>
     </div>
   <% } %>
-  <% if (ob.noValidModerators) { %>
-  <p class="clrTErr">
-    <%= ob.polyT('moderators.noValidModerators') %>
-    <%= ob.polyT('moderators.noValidModeratorsBody') %>
-  </p>
-  <% } %>
-  <% if (ob.noValidVerifiedModerators && ob.showVerifiedOnly) { %>
-    <div class="moderatorCard moderatorsMessage clrBr">
-      <div class="moderatorCardInner">
-        <div class="flexCent">
-          <div class="flexColRows flexHCent gutterVTn">
-            <h4><%= ob.polyT('moderators.noVerifiedModerators') %></h4>
-            <div class="tx4 clrT2"><%= ob.polyT('moderators.noVerifiedModeratorsBody') %></div>
-            <div class="padTn"><% // just a spacer %></div>
-            <div>
-            <button class="btn clrP clrBr js-showUnverified"><%= ob.polyT('moderators.showUnverified') %></button>
+  <% if (ob.purchase) { %>
+  <% // only show the errors below if this is a purchase. TODO: move this to the purchase template. %>
+    <% if (ob.noValidModerators) { %>
+    <p class="clrTErr">
+      <%= ob.polyT('moderators.noValidModerators') %>
+      <%= ob.polyT('moderators.noValidModeratorsBody') %>
+    </p>
+    <% } %>
+    <% if (ob.noValidVerifiedModerators && ob.showVerifiedOnly) { %>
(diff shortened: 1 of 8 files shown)


js/views/components/Moderators.js
```diff
@@ -150,21 +150,22 @@ export default class extends baseVw {
...
     const buyerCur = app.serverConfig.cryptoCurrency;
-    const modCurs = profile.moderatorInfo && profile.moderatorInfo.acceptedCurrencies || [];
+    const modCurs = data.moderatorInfo && data.moderatorInfo.acceptedCurrencies || [];
     const validCur = modCurs.includes(buyerCur);
 
     if ((!!validMod && validCur || this.options.showInvalid)) {
-      this.moderatorsCol.add(new Moderator(profile, { parse: true }));
-      this.removeNotFetched(profile.peerID);
+      this.moderatorsCol.add(new Moderator(data, { parse: true }));
+      this.removeNotFetched(data.peerID || data.peerId);
```
rmisio: Why are two peerId variables being sent named with slightly different casing?

jjeffryes: @rmisio I refactored things to send both the profile data and the error data to this function, when it's a profile it's peerID, when it's an error it's peerId because the server doesn't capitalize the d.

js/templates/components/moderators.html
```diff
@@ -6,27 +6,30 @@
...
-      <div class="moderatorCardInner">
-        <div class="flexCent">
-          <div class="flexColRows flexHCent gutterVTn">
-            <h4><%= ob.polyT('moderators.noVerifiedModerators') %></h4>
-            <div class="tx4 clrT2"><%= ob.polyT('moderators.noVerifiedModeratorsBody') %></div>
-            <div class="padTn"><% // just a spacer %></div>
-            <div>
-            <button class="btn clrP clrBr js-showUnverified"><%= ob.polyT('moderators.showUnverified') %></button>
+  <% if (ob.purchase) { %>
+  <% // only show the errors below if this is a purchase. TODO: move this to the purchase template. %>
```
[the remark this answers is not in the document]
jjeffryes: I've added an issue here so we don't forget about this. I think it's simpler to make it a future PR than adding more to this one. 

https://github.com/OpenBazaar/openbazaar-desktop/issues/1318

js/views/components/Moderators.js
rmisio: Would it be better to standardize it in getModeratorsByID so the variable names match when they arrive at processMod? Seems really awkward IMHO that processMod needs to be on the lookout for both.

js/views/components/Moderators.js
```diff
@@ -219,6 +218,8 @@ export default class extends baseVw {
                   if (eventData.error) {
                     // errors don't have a message id, check to see if the peerID matches
                     if (IDs.includes(eventData.peerId)) {
+                      // provide the expected capitalization of peerID
+                      eventData.peerID = eventData.peerId;
```
rmisio: I won't hold up merge for this, but it would have been nice to also add in a `delete eventData.peerId` in there.

PR #1307, opened 2018-04-17
Fix no valid mods text

### LINKED ISSUE #1305: No Valid Moderators message in purchase needs some padding
image

jjeffryes:
This fixes the no valid moderators text to use the same formatting as the no verified moderators text, and prevents them from showing at the same time. If no moderators are valid, then the no verified moderators text shouldn't be visible.

This PR is marked WIP because it overlaps with some of the other PRs that haven't been merged yet, and probably needs to be updated once they're merged into master.

Closes #1305


rmisio:
@jjeffryes can this be closed? It looks like it's included in 1306.

rmisio:
@jjeffryes ok #1306 is in. Can you merge master into this one so I could see what changed? Also, if you wouldn't mind fixing the conflict?

PR #1308, opened 2018-04-18
[WIP] Cryptocurrency listings

rmisio:
This is a WIP mainly because it's waiting on the inventory piece to be made available on the server. Currently, it's using a dummy stubbed in inventory module.

Also, since we're also waiting on the search api to start crawling inventory / crypto data, the search results are hard coded to show the crypto list view layout with some dummy crypto related data.

But... outside of the above, the full life-cycle of creating and purchasing a crypto listing is in place and can be tested.

(there is also a "TODO" or two that I need to clean up, which are mainly notes on specific areas I still want to test)


js/data/cryptoListingCurrencies.js
```diff
@@ -0,0 +1,167 @@
...
+  getExchangeRates,
+  getEvents as getCurrencyEvents,
+} from '../utils/currency';
+
+// Since the exchange rate api is dependant on 3rd parties, we can't absolutely rely on
+// consistent results, so we'll start with a hard-coded list of crypto-listing currencies.
+// The base list could be updated from release to release. If there are additional crypto
+// currencies being returned by the exchange rate api, they will be merged into a combined
+// list the UI will use.
+const baseCurrencies = [
```
rmisio: The way this works is that the list of available coin types for crypto currency listings starts with a hard coded list. If the exchange rate api returns additional crypto currencies, they will be added to the list.

If a particular currency is in the base list but is not returned in the exchange rate api, it will still be available as a coin type and the UI will reflect that exchange rate data is not available and the listing won't be purchasable until that time. It may be that the exchange rate for that coin is temporarily unavailable. If it's a more permanent thing, then we can up the base list before the next release.

I thought this was the best way of having the app not be dependent on the exchange rate api for coin types (the api and/or one of the 3rd parties it pulls from may be down), yet be able to adjust to new coin types being available (i.e. if a new coin type is being returned by the exchange rate api, we don't need a new release for it to be available in the app).

js/utils/inventory.js
```diff
@@ -0,0 +1,131 @@
+import $ from 'jquery';
```
rmisio: This is just a moderately elaborate stub until the server exposes the inventory api and a proper client side module can be made.

js/views/components/VerifiedMod.js
```diff
@@ -1,29 +1,47 @@
...
+import BaseVw from '../baseVw';
 
 export default class extends BaseVw {
   constructor(options = {}) {
     const opts = {
-      arrowClass: '',
-      type: {},
-      data: {},
       ...options,
+      initialState: {
```
rmisio: I refactored this guy so you are explicitly sending it the text and classes you want. It gives you a big amount of control as new use cases present themselves. There are helper function below which let you create options for use-cases that are similar or the same.

js/views/search/Results.js
```diff
@@ -22,6 +22,11 @@ export default class extends baseVw {
     this.serverPage = this.options.serverPage || 0;
     this.pageSize = this.options.pageSize || 24;
     this.reportsUrl = this.options.reportsUrl || '';
+    this.viewType = this.options.viewType || 'grid';
+    // temp
+    // temp
+    this.viewType = 'cryptoList';
```
rmisio: Temporarily hard coding the search to show in crypto list view.

styles/components/_containers.scss
```diff
@@ -364,6 +370,45 @@
   }
 }
 
+.arrowBoxTipWrap {
```
rmisio: Create a re-usable tooltip alternative to the standard one we have in place. The main benefit of this one is that it allows you to put full markup (as opposed to plain text) in the tooltip. A usage example is:

```
    <div class="arrowBoxTipWrap">
      <span>Hover over me for fun times!</span> // the tooltip trigger
      <div class="arrowBoxCenteredTop clrBr clrP"> // the content that appears on hover
         <p>It was the best of times. It was the worst of times.</p>
      </div>
    </div>
```

For the content, you currently have two options that the determine the positioning of the tip (arrowBoxCenteredTop & arrowBoxRightTop). We could add more (e.g. options that position the tip on top of the trigger, etc...). Although, I may need to slightly tweak the markup to properly support that (I'll keep you in the loop if I do it).

The main point is that this should be a re-usable component, so we shouldn't have to have a bunch of custom one off tooltips implemented around the app. Currently, you could choose between the simple plain text (`data-tip`) one of this one. Eventually, we can even consider eliminating the simple one, since this one can also do the same thing with only slightly more markup.

morebrownies:
SKU and Adult content should be hidden on a cryptocurrency listing type 

screenshot 2018-04-18 15 25 35

morebrownies:
Initially the current market value is blank. When I switch to a different coin it then displays properly. 

**Current**
screenshot 2018-04-18 15 26 05

**Expected**
screenshot 2018-04-19 10 17 40

rmisio:
> SKU and Adult content should be hidden on a cryptocurrency listing type

@morebrownies Does a SKU not make sense for a cryptocur listing? Wouldn't it be a product just like anything else?

Are we sure we want to hide mature content? What if someone wants to sell FuckCoin?

morebrownies:
I ordered 30 during Vcash during checkout, but the order confirmation shows I ordered 7.05032704 🤔 

screenshot 2018-04-19 10 40 04

morebrownies:
> @morebrownies Does a SKU not make sense for a cryptocur listing? Wouldn't it be a product just like anything else?

I would vote to hide it for now. If people really need it, we can re-enable it. I don't see much of a use case for it, especially for crypto trading.

> Are we sure we want to hide mature content? What if someone wants to sell FuckCoin?

Seems very low risk to me. I would vote to hide it.

morebrownies:
I have some Zcash up for sale here on Testnet if anyone wants to buy from me: ob://QmcbK5VSojKp7AXKo1J1jzbjZSoKB63bZWA4fCBKyXNMW5/store/zec-zcash-for-sale

morebrownies:
Copy needs to be changed here, we now support 4 listing types

screenshot 2018-04-19 10 44 41

morebrownies:
Copy is wrong here @rmisio. I will not receive payment in ARK. I receive payment in whatever currency my store accepts. 

screenshot 2018-04-19 10 44 51

morebrownies:
Hey @rmisio small thing... It would be nice if the quantity value had the currency abbrevation next to the value, e.g. `0.01 ZEC` instead of `0.01`

screenshot 2018-04-19 12 51 59

morebrownies:
Can you make sure to add a shadow to the tooltip @rmisio? 

screenshot 2018-04-19 14 15 26

rmisio:
ok @morebrownies - i think all your requested fixes are in.

morebrownies:
thank you @rmisio. will test again shortly.

morebrownies:
@rmisio found a issue: Clear out the quantity field, hit `Pay` and confirm the tooltip by hitting `Yes` and this happens: 

screenshot 2018-04-19 15 42 51

rmisio:
> found a issue: Clear out the quantity field, hit Pay and confirm the tooltip by hitting Yes and this happens:

What is the problem? You have at least two errors:

1.) You've entered an invalid quantity. It must be a number.
2.) You haven't provided a payment address.

morebrownies:
Loook in the quantity input @rmisio. It filled it with `data-var-t...`

rmisio:
> Loook in the quantity input @rmisio. It filled it with data-var-t...

good catch. fixed.

morebrownies:
> ok @morebrownies - i think all your requested fixes are in.

Looking good @rmisio 👍

jjeffryes:
@morebrownies I noticed when you hover over the items in someone else's store in list mode the block button overlaps the details button.

Where should it appear?

image

rmisio:
> I noticed when you hover over the items in someone else's store in list mode the block button overlaps the details button.

I think that was done to be consistent with where the buttons are in crypto list view, which the designs had on the far right.

FWIW, it was a big game of whack a mole with the positioning of these darn buttons. We now have 3 listing views with slight permutations depending on whether it's your own listing. I'm not sure we'll find absolutely perfect positioning for every single case.

rmisio:
ok @morebrownies, most recent request are in.

> Oh sorry one more thing... If the listing title is truncated, it would be helpful to display the full title on hover (similar to what we do on listing cards)

One little quirk with this is that the tooltip is applied whether the title is truncated or not. CSS, unfortunately, doesn't offer any way to add conditional behavior depending on if truncation is needed.

morebrownies:
> One little quirk with this is that the tooltip is applied whether the title is truncated or not. CSS, unfortunately, doesn't offer any way to add conditional behavior depending on if truncation is needed.

That's fine, thanks.

jjeffryes:
If I enter a value of 0.00001btc in the quantity field, I get this error:

`json: cannot unmarshal number 1000.0000000000001 into Go struct field item.quantity of type int`

Any idea what's happening there?

jjeffryes:
Is there a reasonably easy way to prevent users from putting amounts in that get converted to exponents?

image

rmisio:
> json: cannot unmarshal number 1000.0000000000001 into Go struct field item.quantity of type int
> Any idea what's happening there?

Yeah, I'll fix it. The server is expecting an integer and js and its crappy handling of floating points/ integers is resulting in a minuscule decimal.

rmisio:
> Is there a reasonably easy way to prevent users from putting amounts in that get converted to exponents?

I dunno. I'll look into it.

js/views/modals/moderatorDetails.js
```diff
@@ -55,6 +55,8 @@ export default class extends BaseModal {
         return langData && langData.name || lang;
       });
 
+    const verifiedMod = app.verifiedMods.get(this.model.get('peerID'));
```
jjeffryes: If this.verifiedModModel already has the model, why switch to getting it a second time here?

Or is that a result of a merge?

js/views/modals/orderDetail/OrderDetail.js
```diff
@@ -529,17 +529,21 @@ export default class extends BaseModal {
     let showDisputeOrderButton = false;
 
     if (this.buyerId === app.profile.id) {
+      console.log('moo');
```
jjeffryes: Just a reminder to take the console.logs out.

js/views/modals/orderDetail/summaryTab/Fulfilled.js
```diff
@@ -32,27 +34,24 @@ export default class extends BaseVw {
...
+      'click .js-copyText': 'onClickCopyText',
     };
   }
 
-  onClickCopyTrackingNumber() {
-    clipboard.writeText(this.dataObject.physicalDelivery[0].trackingNumber);
-    this.$trackingCopiedToClipboard
+  onClickCopyText(e) {
+    const $target = $(e.target);
+    console.log(`writing: ${$target.attr('data-content')}`);
```
jjeffryes: A reminder to remove this console.log.

styles/components/_containers.scss
jjeffryes: Good work here.

styles/modules/modals/_listingDetail.scss
```diff
@@ -55,14 +55,24 @@
...
-          width: 100%;
-          box-sizing: border-box;
-          padding: $padMd;
+        .badge {
+          width: 24px;
+          height: 24px;
+          background-size: contain;
         }
+
+        // .confirmBox {
```
jjeffryes: Don't forget to take these commented lines out.

js/models/purchase/Item.js
```diff
@@ -33,23 +35,56 @@ export default class extends BaseModel {
...
+      if (typeof attrs.paymentAddress !== 'undefined') {
+        addError('paymentAddress', 'The payment address should only be provide on ' +
+          'crypto listings');
+      }
+    } else {
+      const inventory = this.inventory();
+
+      if (typeof attrs.quantity !== 'number') {
+        addError('quantity', app.polyglot.t('purchaseItemModelErrors.quantityMustBeNumeric'));
+      } else if (attrs.quantity <= 0) {
```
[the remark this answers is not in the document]
rmisio: They're doing different checks.

Non crypto:
- ensure integer
- must be greater than or equal to 1

Crypto:
- ensure number
- must be greater than 0
- cannot exceed inventory

js/views/modals/orderDetail/OrderDetail.js
rmisio: Are you saying a stray 'moo' isn't useful?

Anyhow, for this one actually, I'd like to leave it in because this code is changed in the escrow timeout branch and I'm concerned the merge might be weird if I tweak it. I'll make sure the consoles are removed in the escrow branch.

Now if we end up releasing without the escrow branch, then I suppose I could remove the consoles and just deal with the likely merge conflicts.

js/templates/modals/purchase/purchase.html
```diff
@@ -40,30 +42,68 @@
...
+              <div class="flexNoShrink">
+                <input
+                    class="clrBr clrP"
+                    type="text"
+                    id="purchaseQuantity"
+                    size="<%= isCrypto ? 8 : 3 %>"
+                    name="quantity"
+                    value="<%= ob.items[0].quantity %>"
+                    data-var-type="number"
+                    placeholder="<%= isCrypto ? '0.0000' : '0' %>">
```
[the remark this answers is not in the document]
rmisio: I fixed it so the placeholder will show now.

rmisio:
ok @jjeffryes, got the requested changes in (aside from a couple which I responded to in comments). Thanks for the review!

js/models/purchase/Item.js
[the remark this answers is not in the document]
jjeffryes: Right. I was noticing that if the number has to be an integer, then < 1 is the same as <= 0, since there are no integers between 1 and 0. So checking <= 0 could be done once for both.

ie: check attrs.quantity <=0 before if (!this.isCrypto) {.

It's not a big deal, it'd be a small simplification of the code.

js/models/listing/Metadata.js
```diff
@@ -17,19 +20,41 @@ export default class extends BaseModel {
...
+        {
+          code: contractType,
+          name: app.polyglot.t(`formats.${contractType}`),
+        }
+      ));
+  }
+
+  get formats() {
+    return [
+      'AUCTION',
```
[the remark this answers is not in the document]
placer14: FYI, AUCTION wasn't being used. I recommended that @tyler remove it and add the new type. I was thinking it would replace 1 since it was never used, but I tried to keep my recommendations implementation-agnostic. OCD aside, I don't have a real problem with this.

jjeffryes:
If a user on master tries to view a store with crypto listings, they get a "Please provide a currency code" error on line 661 of currencies.js. 

Is there a way to avoid that so people that haven't upgraded can still view stores and buy non-crypto listings?

If not, I don't think it's a dealbreaker, we'll just have to be careful about merging it to master close to when we do the full release, and expect a period where some users can't open stores with crypto listings until they upgrade.

rmisio:
> Is there a way to avoid that so people that haven't upgraded can still view stores and buy non-crypto listings?

Not without pushing some sort of patch, but that would require an upgrade... so why not just have them upgrade to the branch that supports crypto listings?

I don't think its such a big deal because I'm not sure what legitimate stores would be creating crypto listings on mainnet before the code is released.

jjeffryes:
I'm hitting some weird errors. After created a crypto listing, I can't open my own store because the listing breaks in store.js onUpdateCollection, it tries to call converCurrency but the price object only has an amount, and is missing the required currencyCode.

That also causes an issue in the listingCard template, it calls currencyMod.formattedCurrency, which expects a currencyCode but doesn't have one.

jjeffryes:
I'm also seeing NaN for the inventory when looking at my own listings. 

image

When looking at someone else's listing the call to inventory fails, but I don't think that's a client issue.

```
"reason": "Could not resolve name."
```

rmisio:
@jjeffryes What server branch are you on? You should be on `TS_cryptolistings`. It used to be 'cryptolistings' but Tyler stages the latest changes in the TS_ version and periodically moves them to the other one. It's safest to be in the TS one.

If the issue still happens, can you please share a link to offending listing?

rmisio:
@jjeffryes this is ready for re-review.

jjeffryes:
@rmisio can you resolve the branch conflicts?

PR #1315, opened 2018-04-25
support back/forward mouse buttons

premek:
This adds support for back and forward mouse buttons.
Tested on linux. 
Could anyone please try this (fix it) on Windows?

On windows, it should be supported natively by https://electronjs.org/docs/all#event-app-command-windows
On linux there is electron/electron/issues/6996 and https://github.com/jostrander/mouse-forward-back

### DIFF
--- a/js/start.js
+++ b/js/start.js
@@ -37,6 +37,7 @@ import { fetchExchangeRates } from './utils/currency';
 import './utils/exchangeRateSyncer';
 import { launchDebugLogModal, launchSettingsModal } from './utils/modalManager';
 import listingDeleteHandler from './startup/listingDelete';
+import initMouseForwardBack from './startup/mouseForwardBack';
 import { fixLinuxZoomIssue, handleServerShutdownRequests } from './startup';
 import ConnectionManagement from './views/modals/connectionManagement/ConnectionManagement';
 import Onboarding from './views/modals/onboarding/Onboarding';
@@ -951,6 +952,8 @@ ipcRenderer.on('close-attempt', (e) => {
 
 // initialize our listing delete handler
 listingDeleteHandler();
+
+initMouseForwardBack();
 
 if (remote.getGlobal('isBundledApp')) {
   console.log(`%c${app.polyglot.t('consoleWarning.heading')}`,
--- a/js/startup/mouseForwardBack.js
+++ b/js/startup/mouseForwardBack.js
@@ -1,12 +1,12 @@
+const mouseForwardBack = require('mouse-forward-back');
 import $ from 'jquery';
 import { remote } from 'electron';
 
 // handles mouse back/forward buttons
 
 export default function () {
-
   // app-command event is triggered on Windows only
-  $(window).on("app-command", (e, cmd) => {
+  $(window).on('app-command', (e, cmd) => {
     if (cmd === 'browser-backward') {
       window.history.back();
     } else if (cmd === 'browser-forward') {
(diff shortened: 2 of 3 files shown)


jjeffryes:
Thanks for this PR, @premek I will test it and get it working on Windows, probably after we wrap up the current release.

js/startup/mouseForwardBack.js
```diff
@@ -0,0 +1,26 @@
+import { remote } from 'electron';
+
+// handles mouse back/forward buttons
+
+export default function () {
+
+  // app-command works on Windows only
+  $(window).on("app-command", (e, cmd) => {
```
rmisio: Can you please import jQuery in this file? It's available and made global because other files in the app are importing it, but we'd like each file to explicitly declare it's dependencies (better for portability and refactoring).

js/start.js
```diff
@@ -37,6 +37,7 @@ import { fetchExchangeRates } from './utils/currency';
 import './utils/exchangeRateSyncer';
```
rmisio: Can you please make sure to lint this PR and fix any lint errors? `npm run lint`

package.json
```diff
@@ -90,5 +90,8 @@
     "url-parse": "^1.1.7",
     "velocity-animate": "^1.2.3",
     "yargs": "^6.6.0"
+  },
+  "optionalDependencies": {
```
rmisio: I don't think this should be an optional dependancy. For users of Linux, the code will bomb if this dependency is not present.

image

If you really wanted it to stay as optional, then the code that imports it would have to handle any failures of it not being there. From the [npm docs](https://docs.npmjs.com/files/package.json#optionaldependencies):

```
It is still your program's responsibility to handle the lack of the dependency. For example, something like this:
```

Could you please either make it a required dependency in package.json -or- handle the failure of it not being there?

js/startup/mouseForwardBack.js
rmisio: If 'app-command' only works on Windows, wouldn't it be better wrap the 'app-command' call in a check to make sure we're on windows, so the code is not unnecessarily executed (and then the linux "if" could be an "else-if")?

Where is OSX handled? Does OSX have this functionality by default?

premek: This seemed ok to me as it just reacts to the "app-command" event which will be fired only on windows. It just registers the handler on startup that is possibly not used later. Also if this is maybe fixed in electron in a way this will be called on other platforms too it can be used here (not sure about this, see the discussion in the electron issue linked above).
I will change the comment to be more accurate.
I haven't found how to do this on OSX.

package.json
premek: Agreed, I just copied this from some other project where this dependency was used. I will make it a normal mandatory dependency.

js/start.js
premek: Fixed, thanks. 
Could Travis be used to run this automatically? Currently it fails for PRs because it tries to build distribution packages but doesn't have the secrets needed to do that when running from a PR. Maybe you could run lint and tests on PRs and full build on tags only.

premek:
I forgot to mention I'm not really good at JS and not very aware of the OB code :) But I think this feature is quite useful to everybody who got used to using those mouse buttons. So thanks a lot @rmisio for your time. I tried to fix the issues

js/startup/mouseForwardBack.js
```diff
@@ -0,0 +1,27 @@
+const mouseForwardBack = require('mouse-forward-back');
```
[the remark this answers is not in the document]
premek: Done :)

jjeffryes:
I looked into this a bit, I don't think we can use this PR.

mouse-forward-back is a Linux only module, which means neither Mac nor Windows can install it. 

I don't think conditional imports are supported yet in ES6, which means the code that depends on that module throws an error on startup on non-Linux machines, which prevents the app from working.

If someone can adjust the code to avoid that issue we can reopen this.

PR #1316, opened 2018-04-30
add docs to connect to remote server via ssh tunnel

MaximilianMeister:


jjeffryes:
@tyler-smith @cpacia what do you think of this?

christroutner:
This is exactly what running an OpenBazaar server on a P2P VPS will do. It's not trival, or easy. Even with step-by-step directions, not many people are going to be able to follow it.

rmisio:
Hmmm... I'm not sure this makes sense in this doc. It seems like a really advanced, unique scenario. I'm not sure where it should go, but questioning if this is the right place.

rmisio:
Going to close this for now as an advanced configuration option which we're not sure exactly where to document. Thanks @MaximilianMeister for the submission. We may very well circle back to it.

PR #1317, opened 2018-04-30
Swap zcashd wallet for zcash-light

paulbellamy:
And got rid of the special zcashd setup text/dialogs.

Depends on integration with the wallet here: https://github.com/OpenBazaar/openbazaar-go/pull/934

### DIFF
--- a/js/languages/ar.json
+++ b/js/languages/ar.json
@@ -53,10 +53,6 @@
     "currencyHelper": "اختر عملة لمحفظتك",
     "currencyHelper2": "قبول وإرسال الدفعات بالعملة المحددة أعلاه. يرجى الاختيار بعناية، حاليًا ليس ممكن تغيير عملتك بمجرد اختيارها.",
     "curInUseTip": "لديك بالفعل خادم مُدمج مع عملة المحفظة المحددة.",
-    "zcashBinaryLabel": "Zcash Binary",
-    "fullNodeLink": "هنا",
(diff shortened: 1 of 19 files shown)


jjeffryes:
How does a user choose to create a node that users zcash light with this PR? It looks like the full node code has been removed, but not replaced with code to select zcash light.

paulbellamy:
With this PR, when the Zcash wallet option is selected, it passes the `--zcash-light` flag to the server startup.

See here: https://github.com/OpenBazaar/openbazaar-desktop/pull/1317/files#diff-9ab149aa6a7417f2a83b42bd21ce3995R37

paulbellamy:
> I tried it with the server branch, but it doesn't seem to work. That's still in progress?

Yes

> I don't know if we want to completely remove the option for a full binary. What would happen to users currently using that?
> 
> (I'm not sure we have many of those, but it may not be zero)

Replacing the existing binary option was suggested by @cpacia. As long as they continue using the same mnemonic, it should seamlessly transition over (but I haven't explicitly tested that). But, it *does* remove the option to continue running your own full node.

jjeffryes:
I'm going to close this as it is not going to be compatible with the changes we're making in 2.3 for the multi-coin wallet.

If you want to submit a new version of this that is compatible with 2.3 after it's out you can, though I don't know if you'll need to, since that release will allow running a Zcash wallet without a full node.

PR #1334, opened 2018-05-07
switching the  block explorer used for urls to one supporting bech32 …

### LINKED ISSUE #1291: blockchain.info/address does not support bech32, causes error when viewing payment details
The `View payment details` link in the moderator view of a dispute opens the contract payment address in blockchain.info, which causes an error if the address is bech32:

https://blockchain.info/address/bc1qfl894qr5f4xs0pczd9vlsrrlulczcc7a069xfsueq6patkdl9pgselad40

I was gunna write you a PR that switched the link to the payment _transaction_ instead but I'm guessing since that txid is not included in the contract, it's not sent to the moderator? 

I'm hope blockchain.info will support bech32 soon for this case it might be worth switching to an explorer that does support bech32 addresses:

https://en.bitcoin.it/wiki/Bech32_adoption

rmisio:
…addresses

closes #1291 

@jjeffryes FWIW, I didn't test the address one. I wiped recently and don't have a moderated transaction handy. Hoping you do and it would be an easy test for you...? If you only have it on tn, you could probably just fudge `isTestnet` inside that addr function.

### DIFF
--- a/js/data/cryptoCurrencies.js
+++ b/js/data/cryptoCurrencies.js
@@ -24,12 +24,12 @@ const currencies = [
     getBlockChainAddressUrl: (address, isTestnet) => (
       isTestnet ?
         `https://www.blocktrail.com/tBTC/address/${address}` :
-        `https://blockchain.info/address/${address}`
+        `https://blockchair.com/bitcoin/address/${address}`
     ),
     getBlockChainTxUrl: (txid, isTestnet) => (
       isTestnet ?
         `https://www.blocktrail.com/tBTC/tx/${txid}` :
-        `https://blockchain.info/tx/${txid}`
+        `https://blockchair.com/bitcoin/transaction/${txid}`
     ),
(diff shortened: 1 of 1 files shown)


jjeffryes:
On testnet, the moderator can't see the payment details.
Buyer (this works): https://www.blocktrail.com/tBTC/tx/a898ca7816666d8b7f6b49e7ba5f8da654e89158178970f1cd200cffe0d8c057

Seller (this works): https://www.blocktrail.com/tBTC/tx/a898ca7816666d8b7f6b49e7ba5f8da654e89158178970f1cd200cffe0d8c057

Moderator (this doesn't work): https://www.blocktrail.com/tBTC/address/tb1qhu9zxyf40hu7mttdkmxx6a3ql4uyerhs0uuec3sxwe7nyqccvrmqutmgtt

Any idea why the moderator gets a different url and tx? At first glance it looks like the summary template is using the payment address instead of the tx if it's a case?

PR #1342, opened 2018-05-09
Inventory unknown text

### LINKED ISSUE #1340: Not sure if the yellow warning is hoverable but it's blocked
The warning indicator on the right gets blocked by the block and report options.

image

jjeffryes:
This switches the warning icon for the world 'unknown' when inventory for an item fails to load.

This also has some minor tweaks to the quantity display template to allow the unknown text to keep the size of the container it's in, and to reduce the space between the text and the retry link.

Closes #1340 

image

### DIFF
--- a/js/templates/components/quantityDisplay.html
+++ b/js/templates/components/quantityDisplay.html
@@ -17,8 +17,8 @@
   print(ob.spinner({ className: `${ob.spinnerClass}` }))
 } else if (ob.fetchFailed) { %>
   <div class="content">
-    <div class="arrowBoxTipWrap <%= ob.tipClass %>">
-      <i class="ion-alert-circled margRSm clrTAlert"></i>
+    <div class="arrowBoxTipWrap">
+      <i class="clrT2"><%= ob.polyT('inventoryDisplay.unknownInventory') %></i>
       <% if (ob.fetchError) { %>
         <%
           const retryLink = `<a class="js-retry">${ob.polyT('inventoryDisplay.retryLink')}</a>`;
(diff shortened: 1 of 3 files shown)


js/templates/components/quantityDisplay.html
```diff
@@ -27,14 +27,13 @@
...
-            message = ob.polyT('inventoryDisplay.fetchErrorWithMsg', {
+            message = `<div class="rowSm">${ob.polyT('inventoryDisplay.fetchErrorWithMsg', {
               msg: ob.fetchError,
-            });
-            message += `<br /> <br /><div class="txCtr">${retryLink}</div>`;
+            })}</div><div class="txCtr">${retryLink}</div>`;
           }
         %>
-        <div class="arrowBoxCenteredTop clrBr clrP"><%= message %></div>
+        <div class="arrowBoxCenteredTop clrBr clrP <%= ob.tipClass %>"><%= message %></div>
```
jjeffryes: The tipClass was moved here so it wouldn't be applied to the unknown text.

morebrownies:
The screenshot looks good @jjeffryes 

I'm having some issues with my local server right now and can't run this at the moment. But I approve based on what I see in the screenshot.

rmisio:
I didn't think we were putting the `unknown` on the inventory in the listing detail view and purchase screens. I thought we were just talking about the listing card crypto list view in search.

I think the yellow icon was fine in the other places. Like I mentioned the yellow warning icon is a very common element we use for such warnings. It's also often hoverable, as it is in the case of the Listing Detail and Purchase screens.

I would leave those as is.

jjeffryes:
After discussion, this PR is on hold until after the 2.2 release. We're expecting there to be feedback that leads to various layout changes which will likely change what we want to do here.

This is on hold rather than closed in case the feedback indicates this would work with few or no changes.

Issue #1032, opened 2017-10-18
List of currencies should display ISO 4217 code


rhcastilhos:
This list of currencies is seen on general settings, moderation settings and on onboarding.
![image](https://user-images.githubusercontent.com/10789865/31743915-9fb45708-b43b-11e7-85ab-38ad50db554d.png)

![image](https://user-images.githubusercontent.com/10789865/31744021-fa0bdbae-b43b-11e7-82f3-3c489e052c09.png)

"Euro" should be "Euro (EUR)"; 
"United States Dollar" should be "United States Dollar (USD)"; 

Alternatively it could be displayed this way:
EUR - Euro
USD - United States Dollar

This is useful because enables user to search for the code.

rmisio:
@morebrownies what do you think of this one?

morebrownies:
I agree it would be helpful for searching the input. Feels low priority though. 

**I prefer this format:**
Euro (EUR)
United States Dollar (USD)
etc...

jjeffryes:
@jashot7 do you want to pick this one up? It was partially completed in #1357, but the submitter didn't update it when we asked for changes. 

This should also incorporate #1354

PR #1357, opened 2018-05-17
add ISO 4217 code in list of currencies

klaha707:
Fix regarding issue #1032

### DIFF
--- a/js/templates/modals/onboarding/onboarding.html
+++ b/js/templates/modals/onboarding/onboarding.html
@@ -93,7 +93,7 @@
           <% if (ob.settingsErrors.currency) print(ob.formErrorTmpl({ errors: ob.settingsErrors.currency })) %>
           <select id="onboardingCurrency" name="localCurrency" class="clrSh2" data-model="settings">
             <% ob.currencyList.forEach((currency) => { %>
-            <option value="<%= currency.code %>" <% if (currency.code == ob.settings.localCurrency) print('selected') %>><%= currency.name %></option>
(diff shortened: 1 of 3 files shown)


morebrownies:
I'm a bit confused on the question @rmisio. Should the code be in the UI? I'm not sure what that means.

rmisio:
@morebrownies Should the currency drop downs include the code as well as the name, e.g. 'United States Dollar (USD)' instead of just 'United States Dollar'?

rmisio:
> I'm not sure if adding the codes in the template is the right place to do it. That could be an issue for languages where the code should not be in parenthesis or should be on the left instead of the right.

@klaha707 I echo these concerns. Would you mind having it driven by a translation string?

```
// en_US.json
"currencyPairing": "%{convertedCurValue} (%{baseCurValue})",
"currencyWithCode": "%{name} (%{code})",
```

I also think it would make sense to centralize this in the currency module rather than every template having to implement it.

```
// data/currencies.js in getTranslatedCurrencies()
  let translated = currencies.map(currency => {
    const name = app.polyglot.t(`currencies.${currency.code}`);
    return {
      ...currency,
      name,
      nameWithCode: app.polyglot.t('currencyWithCode', {
        name,
        code: currency.code,
      }),
    };
  });

// there's also a second section a few lines down in that method
// where the nameWithCode would need to be similarly worked in
```

klaha707:
Sounds like a better solution with a translation string. That the format of the code could be language specific didn't really cross my mind. Thanks for all the feedback.

jjeffryes:
@klaha707 were you planning to make revisions to this PR based on the comments we made, or should I close it?

rmisio:
@klaha707 We're going to close this one for now. If you get back to it and incorporate the suggested changes, feel free to re-submit.

PR #1359, opened 2018-05-17
Refresh inventory on save

### LINKED ISSUE #1347: CryptoListing Inventory Does Not Update After Edit
When a user creates a crypto listing, then edits the inventory, if they view the listing the refresh prompt will appear but the inventory will not change.

The new inventories saved, and will be correct if the listing is edited, and purchases of more than that amount will be rejected by the server.

The inventory updates if the client is reloaded, but not if just the view is re-rendered.

jjeffryes:
This will fetch the inventory with useCache set to true when the listing is updated and the refresh popup link is clicked.

This primarily comes up when a seller updates their own inventory and then views their listing.

Closes #1347

### DIFF
--- a/js/models/listing/Listing.js
+++ b/js/models/listing/Listing.js
@@ -347,6 +347,12 @@ export default class extends BaseModel {
 
       returnSync.done(data => {
         const hasChanged = () => (!_.isEqual(attrsBeforeSync, this.toJSON()));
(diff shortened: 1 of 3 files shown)


rmisio:
Closing in liu of #1370. I wanted to do it in a way that didn't add an extra fetch of data which the client already had.

PR #1369, opened 2018-05-22
Remove path from page tracking

jjeffryes:
This removes the location from the tracking data, which reduces the path information we're collecting from users.

The path is still being collected in the view data, I haven't found a way to remove that yet.


rmisio:
> The path is still being collected in the view data, I haven't found a way to remove that yet.

@jjeffryes Could you elaborate on that? Is that some data being sent to countly?

PR #1370, opened 2018-05-22
updating the inventory cache when changing inventory on own listing

### LINKED ISSUE #1347: CryptoListing Inventory Does Not Update After Edit

rmisio:
closes #1347

### DIFF
--- a/js/utils/inventory.js
+++ b/js/utils/inventory.js
@@ -126,6 +126,11 @@ listingEvents.on('saved', md => {
     };
 
     setInventory(app.profile.id, inventoryData, { slug: flatMd.slug });
+
+    // todo: would be good to also abort any existing xhr, although it's unlikely there
+    // would be a pending one for your own listings since I think that comes from the db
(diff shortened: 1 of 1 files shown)


js/utils/inventory.js
```diff
@@ -1,6 +1,7 @@
 import $ from 'jquery';
```
rmisio: This module has gotten a bit more complex than I would like. One of the complications is that it manages both fetching inventory at the store and slug level.

I'll chew on it a bit and if I get some clarity on how to simplify it, I'll probably circle back to it.

jjeffryes:
I noticed that if you only change the quantity, when you view the listing it's already updated, but you still get the "data is changed, refresh?"popup, which doesn't change anything if you click it.

PR #1384, opened 2018-05-30
Crypto icons

rmisio:
This changeset implements a `getCryptoData` script which pulls down crypto icons from CMC. It also adds an entry for each coin into the translations file as a mapping of code to name.

The script can be run via `npm run getCryptoData`. (It will take a few minutes because it needs to d/l 1500+ icons and that process is slightly throttled otherwise many of the calls where failing if all were kicked off at once)


bin/getCryptoData.js
```diff
@@ -0,0 +1,162 @@
...
+import fetch from 'node-fetch';
+import Promise from 'bluebird';
+import fs from 'fs';
+import jsonformat from 'json-format';
+import draftLog from 'draftlog';
+
+const iconOutputPath = `${__dirname}${path.sep}..${path.sep}imgs${path.sep}cryptoIcons${path.sep}/`;
+const translationFile = `${__dirname}${path.sep}..${path.sep}js${path.sep}` +
+  `languages${path.sep}/en_US.json`;
+
```
rmisio: Below is just an initial whitelist. Once Tyler implements his, I'll beef this one up. For now we could manually manage it, but I suppose if Tyler exposes an endpoint whereby we could obtain the whitelist, then I could work that process right into the script.

In any case, this script should probably be run at least once per release. If it finds any new/updated icons and/or coin names, the appropriate files will be updated and a PR could be made to merge them in.

jjeffryes:
Can you fix the merge conflicts, @rmisio ?

jjeffryes:
@rmisio can you fix the lint error Travis is complaining about?

PR #1385, opened 2018-05-30
Fix: 1023 Misaligned shipping

### LINKED ISSUE #1023: Misaligned Free Shipping tag
This happens due to the price with +1 or -1 character

image

jashot7:
Fixed [1023](https://github.com/OpenBazaar/openbazaar-desktop/issues/1023)

closes #1023 

Fixed this by fixing the price column to 100px, setting a max width on the free shipping column and clamping both to 4 to match the title/description box.

From @rmisio : Another optimization we could do in the future if we see a problem is progressively decrease the font of the price as the number of digits increase. So, just dummy numbers, but… if the price reaches 15 characters, we drop the font a little, if it reaches 20, we drop it some more. I’ve done that in some other places. Not sure it’s really necessary here at this point though. We could keep that card up our sleeve.

### DIFF
--- a/styles/components/_listingCard.scss
+++ b/styles/components/_listingCard.scss
@@ -240,13 +240,11 @@
     // The price column is a set width and the free shipping column is a maximum of 200px and 4 lines.
     // This leaves the title/description/rating to flex to take up the remaining space.
     .priceCol {
-      padding-right: 20px;
       width: 100px;
       text-align: right;
     }
 
     .freeShipCol {
-      padding-right: 20px;
       max-width: 200px;
       word-break: break-word;
     }
--- a/styles/components/_listingCard.scss
+++ b/styles/components/_listingCard.scss
@@ -237,14 +237,18 @@
   }
 
   .listingsGridListView & {
-    // This isn't ideal, but there's inconistent spacing / widths of cols, plus there's variable
-    // content width on the price (e.g. some currencies may have quite large numbers)
+    // The price column is a set width and the free shipping column is a maximum of 200px and 4 lines.
+    // This leaves the title/description/rating to flex to take up the remaining space.
     .priceCol {
       padding-right: 20px;
+      width: 100px;
+      text-align: right;
     }
 
     .freeShipCol {
       padding-right: 20px;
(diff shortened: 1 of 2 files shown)


styles/components/_listingCard.scss
```diff
@@ -237,14 +237,18 @@
...
+    // The price column is a set width and the free shipping column is a maximum of 200px and 4 lines.
+    // This leaves the title/description/rating to flex to take up the remaining space.
     .priceCol {
       padding-right: 20px;
+      width: 100px;
+      text-align: right;
     }
 
     .freeShipCol {
       padding-right: 20px;
```
rmisio: Can we remove the right padding? It doesn't seem necessary.

jjeffryes: Is the right padding still needed if the price gets very large? We don't want the text of the price to touch the free shipping green background.

Maybe change it to 10px?

rmisio: > Is the right padding still needed if the price gets very large?

No. There's a `gutterHSm` class in place which provides a 10px margin. The padding IMHO is superfluous.

image

rmisio: (unless you're thinking 10px is too tight, in which case I suppose we could drop the padding to 10, which would result in 20px of space.)

jashot7: Just waiting on the consensus from @jjeffryes when ya get a moment. Would you still like me to remove the padding-right:20px on priceCol?

jjeffryes:
Don't forget to put 
`Closes #1023`
in the first comment so the issue will be closed when the PR is merged.

styles/components/_listingCard.scss
jjeffryes: The gutterHSm gives it a 5px margin, I tested a variety of numbers for the price, I think it works fine with the padding-right removed. Please go ahead and remove that.

jashot7: Removed. Ready for merge.

jjeffryes:
@rmisio can you recheck your review and approve if the changes you requested look good?

Issue #1338, opened 2018-05-08
Make transaction ID in order details clickable if it's a link


jjeffryes:
Sellers may paste a link to a transaction ID in a crypto listing when fulfilling the order instead of just the tx, it would be nice to make it clickable if it is.

morebrownies:
Something to consider...I think it could be dangerous to encourage people to paste links into the transaction ID input. It's possible users may exploit this and link to block explorers that are non legitimate and carry out fraud attacks that way. Encouraging users to include ONLY the transaction ID may be the safest approach. Might be worth floating this question around to some of the other devs too to see if I'm being overly conservative here.

jashot7:
I'll also add here (from earlier discussion with @morebrownies) that if we decide to accept my fix/enhancement, we update the UI to point out to the users that they may enter either an ID or a URL:

```TLDR: I think this needs to be changed to something like:

Label: *Transaction ID / Link* Placeholder: *Paste the Transaction ID or Link*

This would require a change in our translations, so I’m not sure how that would work.```

rmisio:
> Something to consider...I think it could be dangerous to encourage people to paste links into the transaction ID input.

I totally agree with that. This bug felt a little sketchy to me from the start.

morebrownies:
Maybe the bug should more so be: _Prevent users from entering links into the transaction ID input_

rmisio:
> Maybe the bug should more so be: Prevent users from entering links into the transaction ID input

That totally makes sense, but that really wouldn't work here because we're trying to support thousands of coins with potentially thousands of different tx id formats. It sounds far-fetched, but can we guarantee `HTMLCoin`s tx id format wouldn't be `<a>abcdefg</a>`.

I think to be non prohibitive and future-proof we just need to be quite liberal on what we allow in the field and then truncate the field in the UI as necessary and flatten any html.. just so it doesn't break any layouts. Anyhow, I'll write an issue for that suggestion.

rmisio:
Closing this since making a link could be aiding vendors posting nefarious links.

PR #1386, opened 2018-05-30
Added support for Crypto fulfillment to include urls/links & Fixed Issue with HTML in fulfillment breaking display/copy functionality.

jashot7:
Fixed [1338: Make Crypto TX ID Clickable](https://github.com/OpenBazaar/openbazaar-desktop/issues/1338)

- Changed template to conditionally display either a URL/Link the user has entered or simply the transaction ID.
21b8e32

- Check to determine if the Crypto Transaction ID is a url, if so present it to the template. Either way, we now strip html from the input - which was previously possible and did break the display and functionality of 'copy'.


rmisio:
Closing this since making a link could be aiding vendors posting nefarious links.

Issue #1383, opened 2018-05-30
"Peer ID" title not translatable in User page


rhcastilhos:
v.2.2.0

![image](https://user-images.githubusercontent.com/10789865/40696785-a17cd882-639d-11e8-980c-9d31fe10c092.png)

PR #1392, opened 2018-05-31
Fix: 1383 Peer ID Not Translatable

jashot7:
Fix for [1383](https://github.com/OpenBazaar/openbazaar-desktop/issues/1383)
Added peerID under userPage. Changed static text in home.html to translated text.


rhcastilhos:
closes #1383

Issue #1388, opened 2018-05-30
Encode markup in the transactionId field


rmisio:
In the Fulfillment section on the Order Details modal encode the field so html is not executed. For example a transaction if of `<a href="iwillhackyou.com">abcdefg</a>` would display as `&lt;a href="iwillhackyou.com"&gt;abcdefg&lt;/p&gt`.

Looks like this would accomplish that:
https://stackoverflow.com/a/6221092/632806

PR #1396, opened 2018-05-31
Fix: 1388 Encode Markup in TX ID Field

jashot7:
Fixed [1388](https://github.com/OpenBazaar/openbazaar-desktop/issues/1388)
Added function to convert characters to escaped versions to reveal attempts at users including markup in the transaction id fulfillment field.

### DIFF
--- a/js/templates/modals/orderDetail/summaryTab/fulfilled.html
+++ b/js/templates/modals/orderDetail/summaryTab/fulfilled.html
@@ -70,7 +70,7 @@
         }) %></div>
 
         <div class="row">
-          <span><%= ob.polyT('orderDetail.summaryTab.fulfilled.transactionIdLabel') %></span> <span class="clamp3 inline"><%= ob.transactionID %></span>
+          <span><%= ob.polyT('orderDetail.summaryTab.fulfilled.transactionIdLabel') %></span> <span class="clamp3 inline"><%= ob.encodedTxId %></span>
           <a class="clrTEm js-copyText flexNoShrink" data-content="<%= ob.transactionID %>" data-status-indicator=".js-transactionIdCopiedToClipboard"><%= ob.polyT('orderDetail.summaryTab.fulfilled.copyLink') %></a>
           <a class="hide js-transactionIdCopiedToClipboard"><%= ob.polyT('copiedToClipboard') %></a>
         </div>
--- a/js/templates/modals/orderDetail/summaryTab/fulfilled.html
+++ b/js/templates/modals/orderDetail/summaryTab/fulfilled.html
@@ -53,7 +53,6 @@
       </div>
     </div>
   <% } else if (ob.contractType === 'CRYPTOCURRENCY') { %>
-    <% const cryptocurrencyDelivery = ob.cryptocurrencyDelivery && ob.cryptocurrencyDelivery[0] || {}; %>
     <div class="flex gutterH clrT">
       <div class="statusIconCol"><span class="clrBr ion-cash"></span></div>
       <div class="flexExpand tx5 posR">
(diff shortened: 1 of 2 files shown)


js/views/modals/orderDetail/summaryTab/Fulfilled.js
```diff
@@ -72,11 +72,31 @@ export default class extends BaseVw {
...
+      '\n': '<br />',
+    }[x]
+    ));
+
+    return output;
+  }
+
   render() {
+    const cryptocurrencyDelivery = this.dataObject.cryptocurrencyDelivery;
+    let transactionID = cryptocurrencyDelivery && cryptocurrencyDelivery[0] || {};
```
rmisio: Would a one-line accomplish the same thing here?

```
const transactionID = cryptocurrencyDelivery && cryptocurrencyDelivery[0] &&
  cryptocurrencyDelivery[0].transactionID || '';
```

I think the encoding could also be done when you pass the variable to the template, although I'd probably rename it to a clearer name, since the encoded version is distinct from what I'd consider the actual transaction id.

```
         ...this.dataObject || {},
         encodedTxId: this.revealEscapeChars(transactionID),
```

js/templates/modals/orderDetail/summaryTab/fulfilled.html
```diff
@@ -71,8 +70,8 @@ <h2 class="tx4 margRTn"><%= ob.polyT('orderDetail.summaryTab.fulfilled.heading')
         }) %></div>
 
         <div class="row">
-          <span><%= ob.polyT('orderDetail.summaryTab.fulfilled.transactionIdLabel') %></span> <span class="clamp3 inline"><%= cryptocurrencyDelivery.transactionID %></span>
-          <a class="clrTEm js-copyText flexNoShrink" data-content="<%= cryptocurrencyDelivery.transactionID %>" data-status-indicator=".js-transactionIdCopiedToClipboard"><%= ob.polyT('orderDetail.summaryTab.fulfilled.copyLink') %></a>
+          <span><%= ob.polyT('orderDetail.summaryTab.fulfilled.transactionIdLabel') %></span> <span class="clamp3 inline"><%= ob.transactionID %></span>
+          <a class="clrTEm js-copyText flexNoShrink" data-content="<%= ob.transactionID %>" data-status-indicator=".js-transactionIdCopiedToClipboard"><%= ob.polyT('orderDetail.summaryTab.fulfilled.copyLink') %></a>
```
rmisio: I think josh and I decided it would be better to have the unencoded form be in the copy text since the user may have some need to get the pure text.

I'm not sure I have strong feelings on this one. @jjeffryes do you have some clarity on which way to go?

jjeffryes: I think it would be good to have the unencoded text be copied, as a failsafe in case the user really does want the unaltered text.

js/views/modals/orderDetail/summaryTab/Fulfilled.js
```diff
@@ -72,11 +72,31 @@ export default class extends BaseVw {
     return this;
   }
 
+  revealEscapeChars(input) {
+    const output = input.replace(/[<>&\n"]/g, x => ({
+      '<': '&amp;lt;',
```
jjeffryes: Would it be simpler to use encodeURIComponent here?

rmisio: encodeURIComponent is for encoding urls. I'm not sure it's going to convert html to an escaped format like this.

jjeffryes: It will turn `<a href="http://hackedyou.com">ClickMe</a>` into `%3Ca%20href%3D%22http%3A%2F%2Fhackedyou.com%22%3EClickMe%3C%2Fa%3E`, that'd do what we need wouldn't it?

rmisio: I guess it would but we're using a screw driver where a hammer is needed. It works for that specific case, but I'm wondering if it wouldn't for others we need it to.

Is there some source that suggested this approach?

jjeffryes: We can leave it as-is.

jashot7: IMO using HTML character entities is significantly more readable than URI encoding. 
```
&lt;a href=&quot;https://www.blocktrail.com/tBTC/tx/7336e170834672538a79ecb293c28957c1d1888ecab8182c3fc7138973e260f2&quot; rel=&quot;nofollow&quot;&gt;the tx test&lt;/a&gt;
```

<a href='https://www.blocktrail.com/tBTC/tx/7336e170834672538a79ecb293c28957c1d1888ecab8182c3fc7138973e260f2' rel='nofollow'>the tx test</a>

I thoroughly explored the approach before deciding to use this, but we can always change it later.

js/views/modals/orderDetail/summaryTab/Fulfilled.js
jashot7: - Made these changes to accompany the team's/Rob's style suggestions.

jashot7:
Ready for review/merge @rmisio @jjeffryes.

jjeffryes:
Looks good to me.

rmisio:
This is pending a couple of changes I ping J So Hot directly with.

jashot7:
@rmisio fixed.

rmisio:
ok @jjeffryes - if latest changes are ok with you, this is ready to merge.

PR #1397, opened 2018-06-01
Fix: 1389 Add length validation to TxID in cryptodelivery

### LINKED ISSUE #1389: Add length validation to the transactionId in the CryptoDelivery model
Validate the field length doesn't exceed 512 characters.

Additionally, the markup in the UI can be given a maxlength attribute so the user is prevented from the exceeding the length.

(I would still leave the model validation in case someone is programmatically setting the model)

jashot7:
closes #1389 

- Added validation to model for 512 maxLength.
- Added translated error message entry.
- Added maxLength attribute to input tag.


js/models/order/orderFulfillment/CryptoDelivery.js
```diff
@@ -8,6 +8,12 @@ export default class extends BaseModel {
     };
   }
 
+  get max() {
```
rmisio: I used to use the name `max` as well and I found `constraints` works better because it would support something like this:

```
  get constraints() {
    return {
      maxTransactionIDLength: 512,
      minTransactionIDLength: 25,
    };
  }
```

js/views/modals/orderDetail/FulfillOrder.js
```diff
@@ -98,13 +98,19 @@ export default class extends BaseVw {
...
+    const transactionIDLength = cryptoDelivery && cryptoDelivery.max.transactionIDLength || 0;
+
     loadTemplate('modals/orderDetail/fulfillOrder.html', (t) => {
       this.$el.html(t({
         contractType: this.contractType,
         isLocalPickup: this.isLocalPickup,
         ...this.model.toJSON(),
         errors: this.model.validationError || {},
         fulfillingOrder: fulfillingOrder(this.model.id),
+        max: {
```
rmisio: Could we just future proof it and pass the whole object in? for example:

```
constraints: cryptoDelivery && cryptoDelivery.constraints || {},
```

js/models/order/orderFulfillment/CryptoDelivery.js
```diff
@@ -20,6 +26,12 @@ export default class extends BaseModel {
       addError('transactionID', app.polyglot.t('orderFulfillmentModelErrors.provideTransactionId'));
     }
 
+    if (attrs.transactionID.length > this.max.transactionIDLength) {
```
rmisio: This should be an else to the block above it because it doesn't make sense to check the length if the attr isn't provided in the first place.

js/models/order/orderFulfillment/CryptoDelivery.js
jashot7: Sounds reasonable. Fixed.

js/models/order/orderFulfillment/CryptoDelivery.js
jashot7: Ah, yeah. Makes sense.

js/views/modals/orderDetail/FulfillOrder.js
jashot7: Makes sense to me. Done.

jashot7:
Requested changes complete.

js/templates/modals/orderDetail/fulfillOrder.html
```diff
@@ -51,7 +51,7 @@
     </div>
     <div class="col7">
       <% if (ob.errors['cryptocurrencyDelivery.transactionID']) print(ob.formErrorTmpl({ errors: ob.errors['cryptocurrencyDelivery.transactionID'] })) %>
-      <input type="text" class="clrBr clrSh2" name="cryptocurrencyDelivery.transactionID" id="fulfillOrderTransactionId" value="<%= ob.cryptocurrencyDelivery.transactionID %>" placeholder="<%= ob.polyT(`orderDetail.fulfillOrderTab.transactionIdPlaceholder`) %>" />
+      <input type="text" class="clrBr clrSh2" name="cryptocurrencyDelivery.transactionID" id="fulfillOrderTransactionId" value="<%= ob.cryptocurrencyDelivery.transactionID %>" placeholder="<%= ob.polyT(`orderDetail.fulfillOrderTab.transactionIdPlaceholder`) %>" maxLength="<%= ob.constraints.transactionIDLength %>" />
```
rmisio: @jashot7 Have you tested this maxlength attr? The doc I see have it as not camel case, so I'm not sure if the camel case will work.

jashot7:
@rmisio Yeah, it was tested. It works, but I pushed a change for the sake of consistency. Here's some background info:

[source](https://www.w3.org/TR/html5/dom.html#embedding-custom-non-visible-data-with-the-data-*-attributes)

```
Attribute names for HTML elements may be written with any mix of lowercase and uppercase letters that are a case-insensitive match for the names of the attributes given in the HTML elements section of this document; that is, attribute names are case-insensitive.
```

```
All attribute names on HTML elements in HTML documents get ASCII-lowercased automatically, so the restriction on ASCII uppercase letters doesn't affect such documents.
```

This in mind, I did push a change from maxLength to maxlength. I found another spot with a global search. Even though it doesn't matter, I'm a fan of consistency as well as I know @rmisio and @jjeffryes are.

rmisio:
@jashot7 sorry - forget to mention this earlier... can you please fix the merge conflict?

jashot7:
@rmisio I merged master in and resolved the conflicting en_US file.

js/languages/en_US.json
```diff
@@ -4261,4 +4262,5 @@
         "zu": "Zulu",
         "zu-ZA": "Zulu (South Africa)"
     }
+    "EUROPEAN_UNION": "European Union",
```
rmisio: @jashot7 Something got borked here. This translation is in a random place. You sure the master you merged from is correct (it seemed to have an odd name)?

jashot7:
@rmisio: Not sure what happened there with that extra errant translation line... I removed it and I'm pretty sure its good now. (Note: The name of master was _weird_ because I merged from a fresh checkout of master that I renamed locally with a temporary name.)

PR #1401, opened 2018-06-04
Mod Currency Warning

jjeffryes:
This will show a warning in settings/store if a mod doesn't support your crypto currency, and prevent users from adding them. This is mostly an issue if the user pastes a peerID into the find by ID input.

If they've already added the mod (if, for instance, the mod has changed their crypto currency through their config file, or they added them before this PR), they can remove the mod.

The easiest way to test this PR is to have two nodes with different currencies, and paste the peerID of one of them into the find by ID input in settings/store for the other one.


js/templates/components/moderatorCard.html
```diff
@@ -30,33 +30,37 @@
...
-          <div class="flex snipKids gutterH tx5 detailsRow">
-            <div class="flexNoShrink TODO">
-              <%= ob.parseEmojis('👍') %> XX<% // placeholder for reputation %>
-            </div>
-            <div class="flexNoShrink">
-              <% var amount = ob.currencyMod.convertAndFormatCurrency(ob.moderatorInfo.fee.fixedFee.amount, ob.moderatorInfo.fee.fixedFee.currencyCode, ob.displayCurrency) %>
-              <%= ob.polyT(`moderatorCard.${ob.moderatorInfo.fee.feeType}`, { amount: amount, percentage: ob.moderatorInfo.fee.percentage }) %>
+            <div class="flex snipKids gutterH tx5 detailsRow">
+              <% if (ob.hasValidCurrency) { %>
+                <div class="flexNoShrink TODO">
```
jashot7: Can you comment on your usage of TODO? There's a comment for it in the stylesheet, but I'd like to know more clearly how you and @rmisio intend to use it.

jjeffryes: Sometimes we have features that were in the design but we can't complete because they're waiting on other work, but we want to make sure they're in the template so adding them won't require a lot of template and CSS  refactoring. 

There's some point at which things have been "TODO" for long enough they can be deleted, that would be a team discussion for each of them.

rmisio:
@jjeffryes Could you fix the merge conflicts on this one?

bin/getCryptoData.js
```diff
@@ -0,0 +1,157 @@
+import path from 'path';
```
rmisio: Hmm... why is this file showing up? This file was merged into master a couple week ago.

build.sh
```diff
@@ -177,7 +177,7 @@ case "$TRAVIS_OS_NAME" in
     echo 'Building Installer...'
```
rmisio: Something seems off. This PR shouldn't be changing these files, right?

bin/getCryptoData.js
jjeffryes: Crud. Updating from master must have screwed something up.

jjeffryes:
This branch seems to have been messed up when I updated it from master.

PR #1403, opened 2018-06-04
View listings link template

rmisio:
Separates the View Listing link from the Edit Listing modal into its own template since it's being re-used twice already and will be a third time in an upcoming feature.


rmisio:
@jjeffryes @jashot7 This is the preparatory piece I mentioned in the View Listing link req call. The change-set looks 10 times larger than it should because I need to wrap a significant piece of code which cause an indentation change.

Anyhow, @jashot7 this is the branch you want to base your branch off of.

PR #1406, opened 2018-06-05
Enable Lint Pre-commit Hook

jashot7:
If linting fails, you will now be unable to commit.


jashot7:
I am pushing a change to make this a pre-push hook instead. I just confirmed that it works after wrangling some local file system issues that were unrelated.

PR #1418, opened 2018-06-13
Ensure BCH prefix exists when using qrCodeText

jashot7:
Adds the prefix to the QR Code text if it is missing.

### DIFF
--- a/js/data/cryptoCurrencies.js
+++ b/js/data/cryptoCurrencies.js
@@ -64,7 +64,7 @@ const currencies = [
 
       const prefix = app.serverConfig.testnet ? 'bchtest' : 'bitcoincash';
       prefixedAddress = address.startsWith(prefix) ?
-      prefixedAddress : `${prefix}:${address}`;
+        prefixedAddress : `${prefix}:${address}`;
 
       return prefixedAddress;
     },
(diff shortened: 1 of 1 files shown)


js/data/cryptoCurrencies.js
```diff
@@ -59,7 +59,16 @@ const currencies = [
...
+    qrCodeText: address => {
+      let prefixedAddress = address;
+
+      if (address.split(':').length === 1) {
+        const prefix = app.serverConfig.testnet ? 'bchtest' : 'bitcoincash';
+        prefixedAddress = `${prefix}:${address}`;
+      }
+
+      return prefixedAddress;
+    },
```
rmisio: would using startsWith() be simpler here?

```
const prefix = app.serverConfig.testnet ? 'bchtest' : 'bitcoincash';
const prefixedAddress = address.startsWith(prefix) ?
  prefixedAddress : `${prefix}:${address}`;
```

Not a huge deal if you prefer the other way, assuming BCH addresses will never have a colon in them later on (which I think is a correct assumption).

FWIW, it's pretty standard in JS from what I've seen to do:

`if (address.split(':').length) {`
-instead of-
`if (address.split(':').length === 1) {`

-and-

`if (!address.split(':').length) {`
-instead of-
`if (address.split(':').length === 0) {`

jashot7: Agreed. Fixed and ready for merge.

PR #1419, opened 2018-06-14
Fix extra spaces in language file

jjeffryes:
Somehow extra spaces were merged into the language file, this puts them back to normal.

The diff looks big, if you set it to ignore white spaces you'll see spacing is the only difference.


rmisio:
The spacing needs to be adjusted in the crypto icons script.

PR #1436, opened 2018-07-02
Trumbowyg translations

### LINKED ISSUE #1395: Text editor (Trumbowyg) is not translated
Listing text editor (Trumbowyg) is not being translated. However the editor has language parameters that could be configured (https://alex-d.github.io/Trumbowyg/documentation/#localization)

image

jashot7:
closes #1395 

See commits below for details.


js/start.js
```diff
@@ -29,7 +29,7 @@ import StartupConnectMessaging from './views/StartupConnectMessaging';
 import { openSimpleMessage } from './views/modals/SimpleMessage';
 import Dialog from './views/modals/Dialog';
 import StatusBar from './views/StatusBar';
-import { getTranslationLangByCode } from './data/languages';
+import { getTranslationLangByCode, getTrumboLangFileNameByCode } from './data/languages';
```
rmisio: Everything done in this file is completely unnecessary because there's no point of integrating the trumbo translation files into polyglot because they are incompatible. Furthermore, the trumbo component is driven by the lang option when it's constructed and what is or isn't in polyglot will have no bearing on it.

I would just revert all the changes in this file. Most of the changes in the other files are relevant and should stay. I'll give more comments in those files.

js/utils/lib/trumbowyg.js
```diff
@@ -12,26 +13,28 @@ export const editorOptions = {
...
-    ...options,
+  const language = app.localSettings.get('language');
+
+  options.editorOptions = {
+    ...defaultEditorOptions,
+    lang: language,
+    ...options.editorOptions || {},
   };
 
+  console.log(`trumboOptions: ${JSON.stringify(options.editorOptions)}`);
```
rmisio: Can we please delete the debugging output?

js/utils/lib/trumbowyg.js
```diff
@@ -12,26 +13,28 @@ export const editorOptions = {
...
 };
 
 export function installRichEditor(attachPoint, options = {}) {
-  const opts = {
-    editorOptions: {
-      ...editorOptions,
-      ...options.editorOptions || {},
-    },
-    ...options,
+  const language = app.localSettings.get('language');
```
rmisio: Since this isn't used anywhere else, can we just set this inline when setting the `options.editorOptions`?

js/data/languages.js
```diff
@@ -1187,3 +1200,13 @@ export function getTranslationLangByCode(code) {
 
   return getIndexedTranslationLang()[code];
 }
+
+export function getTrumboLangFileNameByCode(code) {
```
rmisio: Instead of having to loop through that list everytime, this could take advantage of the `getIndexedTranslationLang` function which will index that list by code and cache it. The list doesn't change, so caching it will ensure it only has to be looped through once instead of on every call.

You could basically change:

```
const lang = translationLangs.find((x) => x.code === code);
return lang.trumbowyg || null;
```

to

```
return getIndexedTranslationLang()[trumbowyg];
```

js/utils/lib/trumbowyg.js
rmisio: This should be run through `getTrumboLangFileNameByCode` to ensure you are mapping the lang code from settings to the trumbo one since they differ on some codes, for example:

```
options.editorOptions = {
  ...defaultEditorOptions,
  lang: getTrumboLangFileNameByCode(app.localSettings.get('language')),
```

js/data/languages.js
rmisio: Just updated that example code... there was a mistake in there.

js/start.js
jashot7: > Everything done in this file is completely unnecessary because there's no point of integrating the trumbo translation files into polyglot because they are incompatible. 

@rmisio @jjeffryes The point of what I'm doing in start.js is including the appropriate trumbowyg translation file. That is 100% necessary to do here. Should I being using a different method to programmatically include the necessary translation files?

Trumbowyg doesn't work without including the specific translation file as I'm doing here.

js/utils/lib/trumbowyg.js
jashot7: Can we get this into the linter? There's no way we should have listing allow console logs ;D

js/data/languages.js
jashot7: @rmisio This makes sense - but I can't reuse the getIndexedTranslationLang as is since it doesn't include the trumbowyg property. Would you prefer that I modify getIndexedTranslation so that the trumbowyg property is included _or_ make a separate cache?

js/start.js
jashot7: @rmisio @jjeffryes Ah, I see. I didn't intend to use polyglot to include trumbowyg's translation files. BUT, I did intend to include the file - so what I'll plan to do is programmatically inject the file without using polyglot. 

Do we have a preferred method for injecting the appropriate file? Everything about what I'm doing here is still necessary, just not importing it with polyglot.

js/start.js
```diff
@@ -62,27 +62,25 @@ function getValidLanguage(lang) {
...
 const initialLang = getValidLanguage(app.localSettings.get('language'));
 app.localSettings.set('language', initialLang);
 moment.locale(initialLang);
 app.polyglot = new Polyglot();
-app.polyglot.extend(require(`./languages/${initialLang}.json`));
-
-let trumboLang = getTrumboLangFileNameByCode(initialLang);
-if (trumboLang) {
-  // eslint-disable-next-line global-require
-  app.polyglot.extend(require(`../node_modules/trumbowyg/dist/langs/${trumboLang}.min.js`));
```
jashot7: @rmisio @jjeffryes: So, this is the specific line I need to change from using polyglot to simply including this specific language file. 

Am I able to programmatically simply change this to:

``` 
require(`../node_modules/trumbowyg/dist/langs/${trumboLang}.min.js`)
```

or is there some other/better way to programmatically inject this trumbo lang file?

rmisio: Well, those language files just add those translations to a `jquery` namespaced variable in the global scope, so in `installRichEditor`, I would do something like:

```
const lang = getTrumboLangFileNameByCode(app.localSettings.get('language'));

if (lang && !window.jQuery.trumbowyg.langs[lang]) {
  require(`../node_modules/trumbowyg/dist/langs/${lang}.min.js`);
}
```

js/utils/lib/trumbowyg.js
jjeffryes: I don't think we need a linting rule against console logs. We do use them in production in a few places.

jashot7: I prefer a no console rule where we override, but if everyone's fine just manually catching them if they slide... ;D

jashot7:
@jjeffryes @rmisio Ready for review/merge.

js/data/languages.js
jashot7: sure does. I overlooked that re-mapping before. Thanks!

js/start.js
```diff
@@ -626,6 +626,7 @@ function start() {
                 const parsedData = app.walletBalance.parse({
                   confirmed: e.jsonData.walletUpdate.confirmed,
                   unconfirmed: e.jsonData.walletUpdate.unconfirmed,
+                  height: e.jsonData.walletUpdate.height,
```
jjeffryes: Should this change be in this PR? It doesn't seem to be related to trumbowyg.

js/utils/lib/trumbowyg.js
```diff
@@ -12,26 +14,33 @@ export const editorOptions = {
...
 
 export function installRichEditor(attachPoint, options = {}) {
-  const opts = {
-    editorOptions: {
-      ...editorOptions,
-      ...options.editorOptions || {},
-    },
-    ...options,
+  const thing = app.localSettings.get('language');
+  const lang = getTrumboLangFileNameByCode(thing);
```
jjeffryes: Can we use a descriptive variable like "currentLang" instead of "thing"?

Or just do `const lang = getTrumboLangFileNameByCode(app.localSettings.get('language');`

jashot7: Whoops. That was debug code. Thanks for catching. ;D

jashot7: Fixed.

js/start.js
jashot7: Ah, that got merged in from master when I reverted start.js. I'll remove that...

rmisio: Please make sure it matches master. It may just be a situation where you need to merge master into your branch.

jashot7: Fixed.

rmisio: Please make sure it matches master. It may just be a situation where you need to merge master into your branch.

jashot7:
Merged master.

PR #1438, opened 2018-07-05
Price modifier

### LINKED ISSUE #1340: Not sure if the yellow warning is hoverable but it's blocked

### LINKED ISSUE #1371: Add Markup to Crypto Listings
## User Story
As a seller, I want to be able to price my crypto listing above or below the market price.

### Edit Listing
- [ ] Show accepted crypto currency to currency being sold where the title would normally be.
-[ ] The title is not shown, and can not be edited, but should still be saved. It should have the format of `${payment currency code} -> ${currency code for sale}`. For example "BTC -> ZEC'.
- [ ] If the title is changed outside of this client, when it is saved in this client it will be overwritten with the format above.
- [ ] If the coin for sale changes after the listing was saved, and is saved again, the title should be updated. The slug will still be based on the old title, like it is with other listings.
- [ ] The for sale coin select list should be renamed ’Trade’.
- [ ] The for sale coin select list should have coin icons at the start of each coin name option label.
- [ ] A new price modifier input should be added. It is a percentage number field, which can have up to 2 decimal places, and can be negative.
    - [ ] We need to determine how to handle the decimal points when communicating with the server to avoid floating point issues. This is already done with coupon discounts, we can probably use the same pattern that is used there.
    - [ ] The price modifier input value is negative if a minus sign is manually entered in front of the number. It should accept a space or no space between the minus sign and the number. It should also accept a plus sign as positive.
    - [ ] The price modifier cannot be less than -99.99%, and cannot be more than + 1000%.
    - ie: "-0.01," "- 99.99," "10," "+5," and "+ 1000" are all valid values.

### Discovery
- [ ] The design of cryptocurrencies when displayed in a list should be updated (see design).
- [ ] Each listing would show if it is over, at, or under market price and by what percentage.

### Listing Details
- [ ] The title treatment from the edit listing should be shown here.
- [ ] The conversion rate between the buying and selling coins should be shown.
- [ ] The buy button should say Trade Now.
- [ ] The inventory should have a coin icon.
- [ ] This line should gracefully shift to two or more lines if it is too big when it is translated into other languages.

### Purchase
- [ ] The title treatment from the edit listing should be shown here.
- [ ] The amount of the user's coin being spent for the current quantity being bought should be shown under the title.
- [ ] The quantity would have an option to purchase a quantity in fiat, if the user has a fiat currency. For example, purchase $30 of ETH.
    - This is considered a nice to have, and is optional.
- [ ] REVISION: The total inventory available will be shown.
- [ ] The price per coin should be shown with the percentage above or below the market, and color and an icon indicating if it is at, above, or below.
- [ ] The receipt should show the title treatment.

### Transactions
- [ ] Show the title treatment.
- [ ] The purchases, sales and cases APIs will need to be updated to support the data needed.

### Order Details
- [ ] Add the coin icon when fulfilled.
- [ ] Add a coin icon to the payment.
- [ ] REVISION: The icons next to the payment will remain unchanged. The crypto coin icon could be added to the payment amount next to it, but that is considered optional, since that could impact many parts of the app.
- [ ] REVISION: The conversion rate will be removed from the design, due to not having data for the relative prices at the time of the payment.
- [ ] Show the percentage above or below the market, and color code the amount.
- [ ] Show the icon on the payment address.

## Notes
- When multicoin wallets are supported, we will need to decide how to handle that in crypto listings.

## Data
Server changes are documented here:
https://github.com/OpenBazaar/openbazaar-go/issues/962

rmisio:
closes #1371 
closes #1340

Currently there is a server bug preventing viewing Cases in the Cases transaction table:
https://github.com/OpenBazaar/openbazaar-go/issues/1003

In theory, when that's fixed, no client action is necessary. But, for the time being, it will prevent you from seeing any cases with this PR.

*NOTE:* this must be tested in conjunction with the TS_crypto-markup server branch.


rmisio:
@jjeffryes As an FYI, one divergence from the Zeppelin designs is that we won't be showing "neutral" prices in green. Those will be in black. We will only use the green/red in the places where we show the market relation (e.g. 3% below market value) text.

@morebrownies is fine with punting on the green colored neutral text and he'll review this PR and let us know if it doesn't feel right.

Also, we aren't showing this price on the Order Detail:
image

Instead we're sticking with the standard price that is in place now:
image

morebrownies:
Hey @rmisio I had a chance to review the PR and I found a few small things that can be imporved. 

1) **Discover** should display green/red/black price based on the price modifier. The green checkmark and/or arrows should display next to the prices too. 
--
**Expected design** 
screenshot 2018-07-09 12 27 05
--
**Current**
screenshot 2018-07-09 12 28 31

2) All inventory is rending as _Unknown_ for me. 

3) Similar to 1 above, the color pricing should also carry over to the listing detail view. It looks like the checkmark is displaying correctly and the colors are working, but we should also output `market price` under the price when price = market.
screenshot 2018-07-09 12 32 49 

4) The listing card should also inherit the pricing color. See the example I posted in Zeplin
screenshot 2018-07-09 12 37 44

5) Small thing I noticed... The price is bold on the listing detail view but not the checkout view. Let's bold in both spots.
screenshot 2018-07-09 12 42 31
--
screenshot 2018-07-09 12 42 35

6) I haven't tested the transaction detail views yet. If someone has a dollar to spend, can you please place an order for the following listing? ob://QmYTRdwd4nhBRJX5QLNCbKxzPNrNaGCUVnvhFMK1tm3fhz/store/BTC-ARK

styles/_theme.scss
```diff
@@ -130,6 +130,14 @@ input[type=range][class~="clrS"] {
   @include textColor($emphasis2);
 }
 
+.clrTPriceAboveMarket {
+  @include textColor(#e4072b);
```
jjeffryes: Should this be hard coded, or use the error color? 

We may want it to be hard coded so it never changes if we later add theming. If so, the below market color should probably be hard coded below.

rmisio: I didn't use the error color because the error color is a slightly different color. We use `red` for that whereas the designs use `#e4072b` for the above market price. @morebrownies Should those colors be the same?

> We may want it to be hard coded so it never changes if we later add theming.

I was thinking the colors would have to be theme-able, otherwise if we hard code them and someone would theme up a background not conducive to the ones we hard code, it could completely wash them out.

morebrownies: > @morebrownies Should those colors be the same?

Yeah sorry about that. Let's use the same red.

jjeffryes:
@morebrownies should the price in the store also show if it is below or above market price?

image

PR #1440, opened 2018-07-09
Analytics enhancements

### LINKED ISSUE #1407: Analytics Enhancements
Add the following enhancements to the metrics system.

- [x] Add a version tracking system so the opt in can be shown again if the current version of the analytics is updated. For instance, if the user has opted into version 1, and the analytics is now version 2, the opt in should be shown again with changes highlighted.
- [x] Refactor common segmentation like error: none so it doesn't have to be repeated.

### Opt In Modal
- [ ] Add changes to the description, with the release version they were changed in.

### Specific Events
- [x] Discover_SearchError: add segmentation for the provider.
- [x] Purchase_PayFromWallet: can this record if it was done from the purchase or order details screen?
- [x] Purchase_ConfirmBoxInsufficientFunds/Purchase_ConfirmBoxEstimatedFeeFailed: consolidate these into a single start/end event.

### New Events
Edit: Add the following events. Each one will be recorded as an event with no other data, unless segmentation is noted, as standard for our analytics events.
- [x] Searches from the address bar.
- [x] Track events like fulfillment in the order details view.
- [x] Create listing.
- [x] Clone listing.
- [x] Delete listing.
- [x] Edit listing.
- [x] Save listing.
    - type of listing.
    - new or edited.
- [x] Open settings.
    - tab
- [x] Significant buttons in settings/advanced
- [x] Clicking the wallet receive money button.
- [x] Clicking the wallet send money button.
- [x] Opening the feedback modal.
- [x] Clicking on a notification.
- [x] Chat, message sent, message received.
- [ ] If possible, record payment events in a way that shows how common paying with an external wallet vs. the internal wallet is.

### Session Issues
- [ ] Fix bug where the session sometimes ends while the user is still using the app.
- [ ] See if there is a way to record the currently loaded page when the session ends.

jjeffryes:
This adds tracking to many events, and makes two changes to the analytics code:

1. A default errors: none is added to the ajax event tracking, so it doesn't need to be added to calls that had no issues.

2. A version system was added so when a significant update is made to what is tracked users will be asked to re-opt in.

Closes #1407


js/views/modals/purchase/Payment.js
```diff
@@ -42,6 +43,7 @@ export default class extends BaseVw {
     this.paymentAddress = options.paymentAddress;
     this.orderId = options.orderId;
     this.isModerated = options.isModerated;
+    this.page = Backbone.history.getFragment().split(/[//?]/)[1];
```
rmisio: What is the purpose of this? Isn't it alway going to be `store`?

js/views/modals/purchase/Payment.js
```diff
@@ -142,6 +147,7 @@ export default class extends BaseVw {
           this.showSpendError(err);
           recordEvent('Purchase_SpendFromWallet', {
             currency,
+            page: this.page,
             errors: err || 'unknown error',
```
rmisio: Can we be consistent with `none` vs. `unknown error`?

js/views/modals/wallet/SpendConfirmBox.js
```diff
@@ -21,6 +21,7 @@ export default class extends baseVw {
     };
 
     this.lastFetchFeeEstimateArgs = {};
+    this.metricsOrigin = options.metricsOrigin;
```
rmisio: Can we please handle the possibility of this not being provided? In the current implementation it will probably end up sending something like `undefined_ConfirmBoxSend`.

I'm not sure we should make it required because could there reasonably be a case we'd want a spend confirm box that we don't track via analytics?

Perhaps a less rigid approach would be to make an internal recordEvent helper:

```
recordEvent(..args) {
  if (!this.metricsOrigin) return;
  recordEvent(...args);
}
```

js/views/modals/wallet/SpendConfirmBox.js
```diff
@@ -110,8 +117,9 @@ export default class extends baseVw {
           fetchFailed: true,
           fetchError,
         });
-        recordEvent('Purchase_ConfirmBoxEstimateFeeFailed', {
-          errror: fetchError || 'unknown error',
+
+        endAjaxEvent(`${this.metricsOrigin}_ConfirmBoxEstimateFee`, {
+          errors: fetchError || 'unknown error',
```
rmisio: Same comment as above about standard no error state.

js/start.js
```diff
@@ -584,14 +582,18 @@ function start() {
 
           localStorage.serverIdAtLastStart = curConn && curConn.server && curConn.server.id;
 
-          if (app.localSettings.get('shareMetrics') === undefined) {
+          const metricsOn = app.localSettings.get('shareMetrics');
+          const metricsNewV = app.localSettings.get('mVersion') !== mVersion();
```
rmisio: This is optional, but would it be better to check if the version is smaller rather than not equal? This wouldn't come into play in prod, but would for the devs going between branches.

(there's another place this check it made as well)

js/utils/metrics.js
```diff
@@ -8,6 +8,15 @@ import { remote } from 'electron';
 
 let metricsRestartNeeded = false;
 
+/** Set the returned string to a higher number any time there are changes to the analytics that
+ * require a new opt in. This will cause the opt in modal to appear again to users that have
+ * perviously opted in. It will not show it to users that have opted out.
+ * @returns {string}
+ */
+export function mVersion() {
```
rmisio: What is the purpose of this being a function as opposed to a constant?

js/views/modals/purchase/Payment.js
jjeffryes: It's to tell if the payment was made on the purchase screen or the transactions screen.

It returns "store" or "purchases".

js/views/modals/purchase/Payment.js
jjeffryes: It is. This is in the fail handler, so it's not 'none', 'none' means there were no errors.

I noticed sometimes err is undefined, I'd rather record that as 'unknown error' in the metrics so it's more clear to Jenn and Sam.

js/views/modals/purchase/Payment.js
rmisio: Can we just capture the whole url instead of piece mealing it? It seems fragile to me to integrate into a common component that we're only going to grab the third section of the url which may be who know's what in a different context.

js/utils/metrics.js
```diff
@@ -126,6 +126,11 @@ export function showMetricsModal(opts) {
 }
 
 export function recordEvent(key, segmentation) {
+  if (!key) throw new Error('Please provide a key');
+  if (segmentation && !_.isObject(segmentation)) {
```
rmisio: What does _.isObject give us that the native `typeof segmentation === 'object'` wouldn't?

js/utils/metrics.js
```diff
@@ -152,3 +162,49 @@ export function endAjaxEvent(key, segmentation) {
...
 }
+
+/** Use when a component needs to pass a different prefix for a recorded event, usually passed in by
+ * the parent. If no prefix is passed in, the event will be named with just the eventName.
+ * @param opts.prefix(string)
+ * @param opts.eventName(string)
+ * @paren opts.segmentation(object)
+ */
+export function recordPrefixedEvent(opts) {
+  if (!opts.eventName) throw new Error('Please provide an eventName');
```
rmisio: Generally, you want to explicitly declare required arguments that don't have any default, with the options hash at the end, for example:

`recordPrefixedEvent(eventName, segmentation, opts = {})`

I suppose a possible exception to that is if you have a whole lot of required ones, but even then you'd probably want to semantically seperate required and options ones, e.g:

`recordPrefixedEvent(manyManyReq = {}, opts = {})`

(required args are fine within the opts hash, but if they are given defaults. It's a bit odd, IMHO to put ones that the user is forced to provide in the options hash)

js/utils/metrics.js
```diff
@@ -152,3 +162,49 @@ export function endAjaxEvent(key, segmentation) {
...
+  const prefix = opts.prefix ? `${opts.prefix}_` : '';
+  startAjaxEvent(`${prefix}${opts.eventName}`);
+}
+
+/** Use when a component needs to pass a different prefix for the end of an ajax event, usually
+ * passed in by the parent. If no prefix is passed in, the event will be named with just the
+ * eventName. The start ajax event can be prefixed manually.
+ * @param opts.prefix(string)
+ * @param opts.eventName(string)
+ * @paren opts.segmentation(object)
```
rmisio: did you mean `@param` instead of `@paren`?

js/utils/metrics.js
jjeffryes: It detects null, typeof will say null is an object.

It's not too likely null would get passed in, but it seemed worth checking. I could also do typeof and !==null

rmisio: but in the first part you're checking for a `segmentation` truthy, which would be false if it was null.

Anyways, it's not a huge deal since we're using underscore anyway. In the web-buy, I imagine we'd be a bit more discriminating since page size is a bit bigger concern.

jjeffryes: Oh, duh. Your're right.

PR #1441, opened 2018-07-10
Reformed Markup & CSS to suit larger/smaller translations

### LINKED ISSUE #1437: View in web link in address bar should handle potentially long translations.
image

The View On Web link in the address needs to handle potentially longer translations. While 80px works for English, it might be insufficient for other langs. Can it maybe have a default of 80px, but grow up to a reasonable amount (I dunno, maybe 120px) if necessary? If it is longer than that, can it please be truncated with ellipses... since breaking to a second line doesn't look right?

jashot7:
closes #1437 

Let this be a lesson to me to take into account translatable text from the get go! ;D

Let me know if you guys have any feedback or change requests.

### DIFF
--- a/styles/modules/_pageNav.scss
+++ b/styles/modules/_pageNav.scss
@@ -119,7 +119,6 @@
         content: '';
         width: 15px;
         height: 100%;
-        background: linear-gradient(to right, transparent, $primary 50%);
       }
 
       &.hidePointer {
@@ -130,7 +129,6 @@
         display: none;
         height: 100%;
         pointer-events: all;
-        background: $primary;
 
           .torIndicator {
             background-image: url(../../imgs/icon-tor.png);
@@ -147,7 +145,6 @@
         height: 100%;
         display: flex;
         align-items: center;
-        background: $primary;
 
         a {
           display: inline-flex;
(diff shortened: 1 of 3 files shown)


jashot7:
@jjeffryes Tested and ready for review/merge.

styles/modules/_pageNav.scss
```diff
@@ -106,37 +106,75 @@
...
+      display: flex;
+      align-items: center;
+      padding: 1px 5px 1px 0;
+      box-sizing: border-box;
+
+      &:before {
+        content: '';
+        width: 15px;
+        height: 100%;
+        background: linear-gradient(to right, transparent, $primary 50%);
```
rmisio: Any styling that deals with color classes should be made in _themes.scss. If we ever implement user driven theming, it will all be driven by that file.

Just to clarify, it's fine if the vast majority of the css lives here, just any rules that change color should be moved to that file.

styles/modules/_pageNav.scss
```diff
@@ -106,37 +106,75 @@
...
-        background: linear-gradient(to right, rgba(255,0,0,0), white 25%);
-        position: absolute;
-        top: 1px;
-        right: 7px;
-        width: 15px;
-        height: 22px;
-        padding-left: 5px;
+        height: 100%;
         pointer-events: all;
+        background: $primary;
```
rmisio: See color comment above.

jashot7:
@jjeffryes Should be ready for merge.

PR #1446, opened 2018-07-18
Add Legal Language to Analytics

jjeffryes:
This adds text requested by our legal advisors to the analytics modal.


jjeffryes:
@SamPatt I'm adding you to the review, to make sure I have all the language correct. I had to add an extra line about listing data, due to new events I added, I followed the pattern for the other lines.

js/utils/metrics.js
```diff
@@ -171,7 +171,7 @@ export function endAjaxEvent(key, segmentation) {
  */
 export function recordPrefixedEvent(eventName, prefix = '', segmentation = {}) {
   if (!eventName) throw new Error('Please provide an eventName');
-  if (segmentation && !(typeof segmentation === 'object')) {
+  if (segmentation && typeof segmentation !== 'object') {
```
rmisio: Actually, I don't even know why the first part is needed, this should be fine:

`if (typeof segmentation !== 'object') {`

and if null would bomb anything (e.g. if countly doesn't handle it), then you were right that you would need the null check.

`if (typeof segmentation !== 'object' || segmentation === null) {`

PR #1453, opened 2018-07-19
1452

### LINKED ISSUE #1452: Possible race condition on order details
Brief Description: When opening the order via the notifications tab for the very first time the name and avatar of the party in the upper left corner does not display. All subsequent times you open the page it does load properly. 

Operating System (OS and version): Ubuntu

image

rmisio:
closes #1452 

This should be tested and merged in conjunction with this server PR:
https://github.com/OpenBazaar/openbazaar-go/pull/1027

(This utilizes the new `asyncID` provided in the above server PR, but it only applies it the "async" profile fetches. We should still add it to the other "async" api's)

### DIFF
--- a/js/models/profile/Profile.js
+++ b/js/models/profile/Profile.js
@@ -1,5 +1,6 @@
 import $ from 'jquery';
 import app from '../../app';
+import { guid } from '../../utils/';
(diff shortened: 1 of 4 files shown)


cpacia:
Server PR is merged

PR #1463, opened 2018-07-23
Pmt finalized messaging

rmisio:


js/views/modals/orderDetail/summaryTab/Summary.js
```diff
@@ -422,9 +422,9 @@ export default class extends BaseVw {
       awaitingBlockHeight: false,
       isFundingConfirmed: false,
       blockTime: cryptoCur.blockTime,
-      isCompletable: orderState === 'FULFILLED',
       isDisputed: orderState === 'DISPUTED',
       hasDisputeEscrowExpired: false,
+      isCompletable: orderState === 'FULFILLED' && this.buyer.id === app.profile.id,
```
jjeffryes: Wouldn't this need to check for the state being PAYMENT_FINALIZED?

jjeffryes:
I tried changing the check in the review to PAYMENT_FINALIZED, that shows the right message, but the buyer doesn't get the UI elements needed to complete the order. 

image

rmisio:
@jjeffryes This has the changes we discussed at the meeting earlier.

js/views/modals/orderDetail/summaryTab/Summary.js
```diff
@@ -675,6 +679,11 @@ export default class extends BaseVw {
     this.$subSections.prepend(this.refunded.render().el);
   }
 
+  shouldShowCompleteOrderForm() {
+    return this.buyer.is === app.profile.id &&
```
jjeffryes: You want buyer.id here, not buyer.is, I think.

js/views/modals/orderDetail/summaryTab/Summary.js
```diff
@@ -936,6 +935,10 @@ export default class extends BaseVw {
       });
     }
 
+    if (this.shouldShowCompleteOrderForm()) {
+      this.renderCompleteOrderForm();
+    }
```
jjeffryes: It might be better to move this after the sort. so it shows up at the top. Currently is gets added in the middle of the other sections, and you have to scroll down to see it.

image

image

rmisio: Yeah, good catch. I couldn't test properly because my buyer just wasn't getting the PAYMENT_FINALIZED message.

PR #1476, opened 2018-07-25
Fix disabled verified only

jjeffryes:
This fixes two small bugs:
- it's possible to end up in a state where an unverified moderator is selected in the purchase, but the show verified moderators only checkbox is checked, so the selected moderator is hidden and the checkbox is disabled so it can't be unhidden.
- when an unverified moderator is selected, the show verified moderators chekcbox is disabled, but not re-enabled when the direct payment option is selected.

### DIFF
--- a/js/templates/modals/purchase/purchase.html
+++ b/js/templates/modals/purchase/purchase.html
@@ -144,7 +144,7 @@
             <div class="flexVCentClearMarg">
               <h2 class="h4 flexExpand required"><%= ob.polyT('purchase.paymentTypeTitle') %></h2>
               <% if (ob.showModerators) { %>
-                <input type="checkbox" id="purchaseVerifiedOnly" class="js-purchaseVerifiedOnly"<% if(ob.showOnlyVerified) print('checked') %>>
+                <input type="checkbox" id="purchaseVerifiedOnly" class="js-purchaseVerifiedOnly"<% if(ob.showOnlyVerified && !ob.unverifedSelected) print('checked') %>>
                 <label class="tx5b <% if (ob.unverifedSelected) print('disabled') %>" for="purchaseVerifiedOnly"><%= ob.polyT('settings.storeTab.verifiedOnly') %></label>
               <% } %>
             </div>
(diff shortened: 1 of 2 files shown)


jjeffryes:
Closely temporarily for some tweaks.

jjeffryes:
Reopening, tweaks not needed.

jjeffryes:
This fix doesn't have to be in 2.2.2, it's pretty minor.

js/templates/modals/purchase/purchase.html
```diff
@@ -144,7 +144,7 @@
             <div class="flexVCentClearMarg">
               <h2 class="h4 flexExpand required"><%= ob.polyT('purchase.paymentTypeTitle') %></h2>
               <% if (ob.showModerators) { %>
-                <input type="checkbox" id="purchaseVerifiedOnly" class="js-purchaseVerifiedOnly"<% if(ob.showOnlyVerified) print('checked') %>>
+                <input type="checkbox" id="purchaseVerifiedOnly" class="js-purchaseVerifiedOnly"<% if(ob.showOnlyVerified && !ob.unverifedSelected) print('checked') %>>
```
rmisio: What is the difference between `showOnlyVerified`  and `unverifedSelected`?

It seems like we may have 2 sources of truth on that state or maybe I'm misunderstanding the distinction.

jjeffryes: unverifiedSelected mean you have selected an unverified moderator, so the ability to show only verified moderators gets disabled, otherwise you end up hiding the selected moderator, and the UX gets weird.

It'd probably be better to do something in the view to set things up so showOnlyVerified can be true or false as long as unverifiedSelected is false, but showOnlyVerified can only be false if unverifiedSelected is true.

I was being lazy with this fix.

rmisio:
I'll circle back to this one. I can't probably review this now, because I can't get listings on mainnet to show up.

jjeffryes:
This is ready for re-review.

jashot7:
- Spent quite a bit of effort trying to repro the original issue, but had no success. 
- Checked a multitude of listings with this fix and found no issues.

Logic looks good to me, so I approve. (Talked to @jjeffryes about this and we're now discussing other potential issues unrelated to this.)

PR #1511, opened 2018-08-09
Add ISO4217 Currency Code to Currency Names

### LINKED ISSUE #1032: List of currencies should display ISO 4217 code

### LINKED ISSUE #1354: Setup selecting base currency does not have GBP ... and should also show currency codes for all
Brief Description: 

Operating System (OS and version): Linux Mint 18.3 (Linux 4.13.0)
OpenBazaar version (shown on `About OpenBazaar` page in menu): 2.2.0 - but can not get the About box at this piont of the install
Hardware: HP Omen 17

Reproducible (Always / Almost Always / Sometimes / Rarely / Couldn't Reproduce): 
Always

Steps to reproduce:
1. Install
2. On the 'set your information' dialogue, select location as United Kingdom Great Britain
3. Try to find base currency GBP under any name or description in drop down

Observed Behavior:
GBP not present

Expected Behavior:
GBP present (and it would help if all the other entries included ISO currency codes)

temp5

jashot7:
closes #1357 
closes #1032 
closes #1354

### DIFF
--- a/js/data/currencies.js
+++ b/js/data/currencies.js
@@ -683,10 +683,17 @@ function getTranslatedCurrencies(lang = app.localSettings.standardizedTranslated
       ' should be returned in.');
   }
 
-  let translated = currencies.map((currency) => ({
-    ...currency,
-    name: app.polyglot.t(`currencies.${currency.code}`),
-  }));
+  let translated = currencies.map(currency => {
+    const name = app.polyglot.t(`currencies.${currency.code}`);
+    return {
+      ...currency,
+      name,
+      nameWithCode: app.polyglot.t('currencyWithCode', {
+        name,
+        code: currency.code,
+      }),
+    };
+  });
 
   if (opts.includeServerCur && app && app.serverConfig && app.serverConfig.cryptoCurrency) {
     const serverCur = getCryptoCurByCode(app.serverConfig.cryptoCurrency);
--- a/js/templates/modals/onboarding/onboarding.html
+++ b/js/templates/modals/onboarding/onboarding.html
@@ -93,7 +93,7 @@
           <% if (ob.settingsErrors.currency) print(ob.formErrorTmpl({ errors: ob.settingsErrors.currency })) %>
           <select id="onboardingCurrency" name="localCurrency" class="clrSh2" data-model="settings">
             <% ob.currencyList.forEach((currency) => { %>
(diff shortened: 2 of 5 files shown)


js/data/currencies.js
```diff
@@ -683,10 +683,17 @@ function getTranslatedCurrencies(lang = app.localSettings.standardizedTranslated
...
+    const name = app.polyglot.t(`currencies.${currency.code}`);
+    return {
+      ...currency,
+      name,
+      nameWithCode: app.polyglot.t('currencyWithCode', {
+        name,
+        code: currency.code,
+      }),
+    };
+  });
```
jjeffryes: Wouldn't it be simpler to leave currencies.js unchanged, and just call the currencyWithCode in the template, like this:
`<%= ob.polyT('currencyWithCode', { name: currency.name, code: currency.code }) %>`?

js/templates/modals/onboarding/onboarding.html
```diff
@@ -93,7 +93,7 @@
           <% if (ob.settingsErrors.currency) print(ob.formErrorTmpl({ errors: ob.settingsErrors.currency })) %>
           <select id="onboardingCurrency" name="localCurrency" class="clrSh2" data-model="settings">
             <% ob.currencyList.forEach((currency) => { %>
-            <option value="<%= currency.code %>" <% if (currency.code == ob.settings.localCurrency) print('selected') %>><%= currency.name %></option>
+            <option value="<%= currency.code %>" <% if (currency.code == ob.settings.localCurrency) print('selected') %>><%= currency.nameWithCode %></option>
```
jashot7:
```
currency.nameWithCode
```

is nicer to use across multiple templates and easier to maintain if we need changes than:

```
<%= ob.polyT('currencyWithCode', { name: currency.name, code: currency.code }) %>
```

I definitely prefer the 'logic' being in one place, but I see your point for simplicity.

jashot7: @jjeffryes ^

js/data/currencies.js
jjeffryes: It's just a question/suggestion, not a requirement.

PR #1515, opened 2018-08-10
No testnet analytics

### LINKED ISSUE #1514: Don't record analytics at all if Testnet.
If the user is on testnet, don't record any analytics events.

jjeffryes:
This will prevent metrics from being recorded if the user is on testnet.

It also adds segmentation for whether a saved listing is new or not for listing_save, and for whether a listing belongs to the user on purchase_start.

Closes #1514

### DIFF
--- a/js/views/modals/editListing/EditListing.js
+++ b/js/views/modals/editListing/EditListing.js
@@ -52,7 +52,7 @@ export default class extends BaseModal {
     super(opts);
     this.options = opts;
 
-    // So the passed in modal does not get any un-saved data,
+    // So the passed in model does not get any un-saved data,
     // we'll clone and update it on sync
     this._origModel = this.model;
     this.model = this._origModel.clone();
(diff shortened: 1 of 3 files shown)


js/views/modals/editListing/EditListing.js
```diff
@@ -750,6 +750,7 @@ export default class extends BaseModal {
       type: serverData.metadata.contractType,
       currency: serverData.metadata.pricingCurrency,
       moderated: serverData.moderators && !!serverData.moderators.length,
+      isNew: this.model.isNew(),
```
rmisio: So this will be true the first time the listing is saved, if they continue to tweak things after that, even without closing and re-opening the modal, it would not be new anymore.

That may not change anything here, but just wanted to make that clear.

jjeffryes: That's exactly what we want here, the intent is to differentiate between the first saves and any further edits.

PR #1521, opened 2018-08-14
Use Smart Counts in Found Phrases

### LINKED ISSUE #1510: Some translations require plural for word "found"
v2.2.3

Some languages require two different translations (singular and plural) to word "found" in messages:
"1 listing found" and "2 listings found".

In spanish it should be:
"1 listado encontrado" and "2 listados encontrados".

This is needed in Store page and in Orders page:
image

image

jjeffryes:
There are several places in the app where the phrase "X found" is used, where a smart count is not used for X. 

This creates an issue where translations that change based on the number can't be used.

This PR replaces the variables for the number with smart_count.

Closes #1510

### DIFF
--- a/js/languages/en_US.json
+++ b/js/languages/en_US.json
@@ -124,7 +124,7 @@
   },
   "connectedPeersPage": {
     "heading": "Connected Peers",
-    "totalPeers": "%{totalPeers} Found",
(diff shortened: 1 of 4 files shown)


rhcastilhos:
shouldn't it be "%{smart_count} found ||| "%{smart_count} found" in en_US.json?

PR #1531, opened 2018-08-15
Update modals

### LINKED ISSUE #1501: Improve unsaved changes modal
There's a few issues with the unsaved changes modal that we should clean up when we have a chance.

1. Simply providing `Yes` `No` buttons isn't clear enough. We should provide more context in the button label, e.g. `Yes, close` or `No, don't close`.
2. The `Yes, close` button should be on the right side as it implies forward progress, while `No, don't close` takes you back.
3. I cleaned up the styling of the modal a bit (buttons are larger and headline text is smaller). It feels more balanced this way.

**Current**
screenshot 2018-07-31 21 23 41

**Ideal**
screenshot 2018-08-02 11 24 23

jashot7:
closes #1501 

~*This is a Work in Progress - looking for some early feedback from @jjeffryes as I continue.*~

### DIFF
--- a/js/views/modals/Dialog.js
+++ b/js/views/modals/Dialog.js
@@ -46,7 +46,7 @@ export default class extends BaseModal {
       buttons: [],
       removeOnClose: true,
       // defaultBtnClass: 'btn clrP clrBr clrSh2',
-      defaultBtnClass: 'flexColRows btnFlx clrP',
+      defaultBtnClass: 'flexExpand btnFlx clrP',
       ...options,
     };
 
--- a/js/views/modals/Dialog.js
+++ b/js/views/modals/Dialog.js
@@ -45,7 +45,6 @@ export default class extends BaseModal {
       messageClass: '',
       buttons: [],
       removeOnClose: true,
-      // defaultBtnClass: 'btn clrP clrBr clrSh2',
       defaultBtnClass: 'flexExpand btnFlx clrP',
       ...options,
     };
--- a/js/views/modals/Dialog.js
+++ b/js/views/modals/Dialog.js
@@ -45,7 +45,8 @@ export default class extends BaseModal {
       messageClass: '',
       buttons: [],
       removeOnClose: true,
-      defaultBtnClass: 'btn clrP clrBr clrSh2',
+      // defaultBtnClass: 'btn clrP clrBr clrSh2',
+      defaultBtnClass: 'flexColRows btnFlx clrP',
       ...options,
     };
 
--- a/js/views/modals/editListing/EditListing.js
+++ b/js/views/modals/editListing/EditListing.js
@@ -611,11 +611,11 @@ export default class extends BaseModal {
         title: app.polyglot.t('editListing.confirmCloseDialog.title'),
         message: app.polyglot.t(`editListing.confirmCloseDialog.${messageKey}`),
         buttons: [{
-          text: app.polyglot.t('editListing.confirmCloseDialog.btnYes'),
-          fragment: 'yes',
-        }, {
           text: app.polyglot.t('editListing.confirmCloseDialog.btnNo'),
           fragment: 'no',
+        }, {
+          text: app.polyglot.t('editListing.confirmCloseDialog.btnYes'),
+          fragment: 'yes',
         }],
       })
         .on('click-yes', () => {
--- a/styles/_theme.scss
+++ b/styles/_theme.scss
@@ -541,6 +541,6 @@ input[type=range][class~="clrS"] {
 
   &:hover, &:focus {
     @include borderColor($border, $border3);
-    z-index: 0;
+    z-index: 1;
   }
 }
--- a/styles/_theme.scss
+++ b/styles/_theme.scss
@@ -344,7 +344,7 @@ input[type=range][class~="clrS"] {
       .selectize-input {
         @extend .#{$baseClass}#{$number};
       }
-  }  
+  }
 }
 
 #appFrame {
@@ -389,7 +389,7 @@ input[type=range][class~="clrS"] {
 
 .arrowBoxTipWrap {
   .arrowBox {
-    box-shadow: 0 1px 2px rgba(0,0,0,0.4);  
+    box-shadow: 0 1px 2px rgba(0,0,0,0.4);
   }
 }
 
@@ -536,12 +536,11 @@ input[type=range][class~="clrS"] {
   }
 }
 
-.btnFlx {
-  // @include borderColor(transparent, transparent);
-  border: none;
-  border-top: 1px solid $border;
+.flexBtnWrapper .btnFlx {
+  border-color: $border;
 
-  &:hover, &::focus {
+  &:hover, &:focus {
     @include borderColor($border, $border3);
+    z-index: 0;
   }
 }
--- a/styles/_theme.scss
+++ b/styles/_theme.scss
@@ -535,3 +535,13 @@ input[type=range][class~="clrS"] {
     }
   }
 }
+
+.btnFlx {
+  // @include borderColor(transparent, transparent);
+  border: none;
+  border-top: 1px solid $border;
+
+  &:hover, &::focus {
+    @include borderColor($border, $border3);
+  }
+}
--- a/styles/components/_buttons.scss
+++ b/styles/components/_buttons.scss
@@ -325,27 +325,30 @@
   }
 }
 
-.btnFlx {
-  // see styleSheet.html for example usage.
-  // use for buttons in the new 'flex' style with more padding and use css flex.
-  @extend .btn;
-  padding: 25px;
+.flexBtnWrapper {
+  .btnFlx {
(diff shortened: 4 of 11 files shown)


js/views/modals/editListing/EditListing.js
```diff
@@ -611,11 +611,11 @@ export default class extends BaseModal {
...
         message: app.polyglot.t(`editListing.confirmCloseDialog.${messageKey}`),
         buttons: [{
-          text: app.polyglot.t('editListing.confirmCloseDialog.btnYes'),
-          fragment: 'yes',
-        }, {
           text: app.polyglot.t('editListing.confirmCloseDialog.btnNo'),
           fragment: 'no',
+        }, {
+          text: app.polyglot.t('editListing.confirmCloseDialog.btnYes'),
+          fragment: 'yes',
```
jashot7: just flipping the order of yes/no, which are now more verbose in the translation file.

styles/modules/modals/_message.scss
```diff
@@ -26,6 +26,10 @@
     .buttonWrap {
       text-align: center;
     }
+
```
jashot7: This seemed like the place to make h1s a tad smaller to match the new design.

styles/components/_buttons.scss
```diff
@@ -325,6 +325,30 @@
   }
 }
 
+.btnFlx {
```
jashot7: This new button type seems different enough from our existing row buttons to warrant a separate name/style, but let me know if you think otherwise.

styles/components/_buttons.scss
```diff
@@ -325,6 +325,30 @@
...
+  }
+
+  + a:hover {
+    margin-right: -1px;
+  }
+
+  &:hover, &:focus {
+    z-index: 0;
+    border: 1px solid;
+    margin: 0px -2px -1px -1px;
```
jashot7: Playing with margins here to get the hover/focus borders to line up well with the existing container. There's still a little content shifting going on, so I'm open to thoughts on how to better do this.

jashot7:
The Dialog's container has rounded borders.

screen shot 2018-08-15 at 5 37 53 pm

The new flat design doesn't have any border radius, therefore displays straight edged borders over the rounded ones. 🤔 

screen shot 2018-08-15 at 5 38 00 pm

styles/components/_buttons.scss
```diff
@@ -325,6 +325,30 @@
...
+.btnFlx {
+  // see styleSheet.html for example usage.
+  // use for buttons in the new 'flex' style with more padding and use css flex.
+  @extend .btn;
+  padding: 25px;
+
+  border-radius: 0px;
+  border-top: 1px solid;
+
+  & + a {
```
jjeffryes: I think this should be `& + .btnFlx`, in case the sibling isn't an anchor tag. And to protect against siblings
that for some reason are anchor tags but not btnFlx.

styles/components/_buttons.scss
```diff
@@ -325,6 +325,30 @@
...
+    border-left: 1px solid #dbdbdb; // Move this color into themes.
+  }
+
+  + a:hover {
```
jjeffryes: This is selecting a sibling to the right when the sibling is hovered, isn't it?

I think you want 
```
&:hover + .btnFlx
```

styles/components/_buttons.scss
```diff
@@ -325,6 +325,30 @@
...
+    margin-right: -1px;
+  }
+
+  &:hover, &:focus {
+    z-index: 0;
```
jjeffryes: What does the z-index of zero do here?

styles/components/_buttons.scss
jjeffryes: I found a better way to prevent the shifting.

Set the borders to transparent, then set the top to be a color.

That way the border is always there, it's just not visible unless it's being hovered over.

js/views/modals/Dialog.js
```diff
@@ -45,7 +45,8 @@ export default class extends BaseModal {
       messageClass: '',
       buttons: [],
       removeOnClose: true,
-      defaultBtnClass: 'btn clrP clrBr clrSh2',
+      // defaultBtnClass: 'btn clrP clrBr clrSh2',
+      defaultBtnClass: 'flexColRows btnFlx clrP',
```
jjeffryes: You want flexExpand here instead of flexColRows.

jjeffryes: Or just add flex-grow: 1 to btnFlx, I think that's cleaner.

styles/_theme.scss
```diff
@@ -534,4 +534,14 @@ input[type=range][class~="clrS"] {
       background: linear-gradient(to right, transparent, $primary 50%);
     }
   }
+}
+
+.btnFlx {
+  // @include borderColor(transparent, transparent);
+  border: none;
+  border-top: 1px solid $border;
```
jjeffryes: This should be border-top-color so it only sets the color.

styles/components/_buttons.scss
```diff
@@ -325,6 +325,30 @@
...
 }
 
+.btnFlx {
+  // see styleSheet.html for example usage.
+  // use for buttons in the new 'flex' style with more padding and use css flex.
+  @extend .btn;
+  padding: 25px;
+
+  border-radius: 0px;
+  border-top: 1px solid;
```
jjeffryes: This has to be border-top-width and border-top-style separately, or you're defaulting the color to black.

styles/_theme.scss
```diff
@@ -534,4 +534,14 @@ input[type=range][class~="clrS"] {
       background: linear-gradient(to right, transparent, $primary 50%);
     }
   }
+}
+
+.btnFlx {
+  // @include borderColor(transparent, transparent);
+  border: none;
```
jjeffryes: Take this out, you don't want to override the width and style from the buttons stylesheet. Instead there should be a border reset there.

styles/components/_buttons.scss
```diff
@@ -325,6 +325,30 @@
   }
 }
 
+.btnFlx {
+  // see styleSheet.html for example usage.
+  // use for buttons in the new 'flex' style with more padding and use css flex.
+  @extend .btn;
+  padding: 25px;
+
+  border-radius: 0px;
```
jjeffryes: Add a border: none; here, to reset the default button borders.

jashot7:
@jjeffryes Ready for Review. I'm adding some comments for @morebrownies in the issue (https://github.com/OpenBazaar/openbazaar-desktop/issues/1501) for design consideration as well.

js/views/modals/Dialog.js
```diff
@@ -45,7 +45,8 @@ export default class extends BaseModal {
       messageClass: '',
       buttons: [],
       removeOnClose: true,
-      defaultBtnClass: 'btn clrP clrBr clrSh2',
+      // defaultBtnClass: 'btn clrP clrBr clrSh2',
```
rmisio: Can we remove the commented out line?

js/views/modals/Dialog.js
```diff
@@ -45,7 +45,8 @@ export default class extends BaseModal {
...
+      defaultBtnClass: 'flexExpand btnFlx clrP',
```
rmisio: I think we should still keep the border and shadow color classes: `clrBr clrSh2`

styles/modules/modals/_message.scss
rmisio: Should we move this style up a level so it also gets applied to simpleMessage modals?

js/views/modals/Dialog.js
jjeffryes: We don't want a shadow class here or we'd get the button shadow overlapping the dialog shadow.

jjeffryes:
@jashot7 If you use a button instead of an anchor, the text stops aligning right. I'm investigating why, but this should work with both buttons and a tags.

jjeffryes:
Well, nevermind. It turns out button elements can not have a display flex property in Chrome.

So we'll have to remember this CSS has to always use anchors, and can't have button elements.

https://github.com/philipwalton/flexbugs#9-some-html-elements-cant-be-flex-containers

jashot7:
Yeah, I found that early on. I'll add that to our documentation, @jjeffryes.

js/views/modals/Dialog.js
jashot7: Yeah, that's no problem. I kept that in there because it was a WIP.

styles/modules/modals/_message.scss
jashot7: Good point. I'll give that a go and see if it causes any unforeseen issues.

jjeffryes:
@jashot7 is this still considered in progress? If not, can you take that out of the title?

jashot7:
@rmisio @jjeffryes This should be able to be merged into your PRs and/or master now.

rmisio:
@jjeffryes this one LGTM.

PR #1536, opened 2018-08-21
fix large number display on user page

### LINKED ISSUE #747: Thousands separators in numbers [several places]
I've noticed thousands separators missing on some numbers in the UI.

User Page Header
image

Search Results
image

jashot7:
closes #747 

All of the tab headers show the abbreviated numbers correctly now. 

I spent quite a bit of time exploring a solution to combine abbrNum and polyglot smart_count/pluralization. It looks like if we wanted to be able to show the abbreviated numbers like we are combined with polyglot, we will have to modify polyglot to be able to have it accept two parameters - the actual number and the formatted number.

On that note, the abbreviation should really be built into polyglot instead.

I'll let you guys discuss this and determine if that's something we really wish to pursue. There may be other options you've previously considered.

Therefore, this fix will not affect the localized text for listings found and number of listings.


jashot7:
screen shot 2018-08-21 at 2 53 36 pm

jjeffryes:
@jashot7 only the tabs need to be abbreviated, to close #747 but we also want to localize the non-abbreviated numbers. Can we add that to this PR?

jashot7:
@jjeffryes Ah, great. If we don't need to combine abbreviation and polyglot pluralization, then this should be ready to go. When you added smart_count, it took care of localization for the listing count here as well. 

Therefore, this should be ready for review.

rmisio:
@jashot7 There already is an integration between abbrNum and our polyglot because the former is piped through both the following translations as well as the native JS number localization functionality (Intl.NumberFormat):

```
  "abbreviatedNumbers": {
    "thousand": "%{number}k",
    "million": "%{number}m",
    "billion": "%{number}b",
    "trillion": "%{number}t"
  },
```

Given that, we probably could do this:

> Therefore, this fix will not affect the localized text for listings found and number of listings.

But I don't know if it's super necessary to shorten those numbers - I think it works best to show the pure number there. I do agree though they should be localized.

> When you added smart_count, it took care of localization for the listing count here as well.

I don't think that's the case. Before the number is passed into the translation as a smart_count variable it would need to be run through something like:

`new Intl.NumberFormat(lang).format(99999)`

You would get the lang from app.localSettings.

If you're feeling really froggy I suppose you could add a helper function to the `util/number`, e.g:

```
const defaultLang = app && app.localSettings && app.localSettings.standardizedTranslatedLang() || 'en-US'

function localizeNumber(num, lang = defaultLang) {
  return new Intl.NumberFormat(lang).format(num);
}
```

jjeffryes:
> I don't think that's the case. Before the number is passed into the translation as a smart_count variable it would need to be run through something like:

Will a smart_count variable work with an internationalized string, or does it need to be a number?

rmisio:
> Will a smart_count variable work with an internationalized string, or does it need to be a number?

oh, you're right. I think what you'd need to do is pass both the smart_count and the internationalized number string and have two translations using the latter as a placeholder.

jjeffryes:
I think you're ok if you just pass the smart count, you don't have to actually have the %{smart_count} anywhere in the translated phrase.

So you could do:
`ob.polyT('somePhrase', { smart_count: ob.numberCount, display_count: ob.stringCount })`

`somePhrase: "%{display_count} things found`

rmisio:
Yup, that's exactly what I meant.

to be specific, it would be:

`somePhrase: "%{display_count} thing not found|%{display_count} things not found`

The smart_count just determines which portion of that translation is used.

js/utils/number.js
```diff
@@ -50,3 +51,26 @@ export function upToFixed(number, decimalPlaces) {
...
+*/
+export function localizeNumber(number, optionalLang) {
+  if (typeof number !== 'number') {
+    throw new Error('Please provide a number.');
+  }
+
+  let lang = optionalLang;
+
+  if (optionalLang === undefined) {
+    lang = app && app.localSettings && app.localSettings.standardizedTranslatedLang() || 'en-US';
```
jjeffryes: Do we need to check app and app.localSettings here? We should be able to rely on those existing.

This could be done with a default in the parameters, like this:
`export function localizeNumber(number, lang = app.localSettings.standardizedTranslatedLang() || 'en-US')`

That would optimize things a bit.

jashot7:
I definitely wouldn't ever want a default param that long.

I agree with all suggestions and this should be ready for merge now, @jjeffryes unless you'd like to retest.

jashot7:
Now, finally. The merge may occur. @jjeffryes

PR #1538, opened 2018-08-22
Check server compatibility

### LINKED ISSUE #1487: Client/Server Mismatch Alert
Add a visible cue in the client if client & server versions are mismatched

jjeffryes:
This will check if the current server matches the server version in package.json, and show a warning modal on startup if it doesn't.

Evidently we have users that have remote servers that have been only updating their clients, and running into issues as a result.

This could also help us avoid issues like we saw in recent releases where the RC was built with the wrong server version and we didn't notice it at first.

Closes #1487


js/start.js
```diff
@@ -634,6 +635,20 @@ function start() {
...
+            const cLink = `<a href="https://github.com/OpenBazaar/openbazaar-desktop/releases">${app.polyglot.t('serverVersionWarning.clientLink')}</a>`;
+            const sLink = `<a href="https://github.com/OpenBazaar/openbazaar-go/releases">${app.polyglot.t('serverVersionWarning.serverLink')}</a>`;
+            const message = app.polyglot.t('serverVersionWarning.message', {
+              serverVersion: app.settings.prettyServerVer,
+              expectedVersion: serverVersionRequired,
+            });
+            const body = `<p>${message}</p><p>${cLink}</p><p>${sLink}</p>`;
+            openSimpleMessage(
+              app.polyglot.t('serverVersionWarning.title'), body
+            ).$el.css('z-index', '9999999');
```
rmisio: I guess this incredibly ugly z-index approach is ok if used extremely judiciously. Normally, you'd just want to open the modal in the right sequence for it to be on top. But... I understand some pages end up asynchronously loading a modal, so that could be a can of worms. The alternative would be to delay the router until they close this dialog, but I suppose it's not super critical here.

PR #1539, opened 2018-08-23
More listings

### LINKED ISSUE #344: Display "More listings by..." section at the bottom of the listing detail screen
It would be nice to display additional listings sold by the vendor at the bottom of the listing detail screen. This is a good opportunity to help people discover more and keep them engaged in the shopping process. 

Not urgent, but if it's relatively low effort, it would be worth adding. 

screenshot 2017-01-31 15 46 22

jjeffryes:
This adds a more listings section to the bottom of each listings details modal. 

Clicking on a listing card in the more listings section will open a new modal and close the parent modal.

Up to 8 listing cards are shown. If none have loaded, or the user doesn't have other listings, the more listings section isn't shown.

Closes #344

### DIFF
--- a/js/languages/en_US.json
+++ b/js/languages/en_US.json
@@ -994,6 +994,7 @@
     "noRefundPolicy": "No return policy entered",
     "termsAndConditions": "Terms Of Service",
     "noTermsAndConditions": "No terms and conditions entered",
+    "moreBy": "More by %{name}",
     "shipping": "Shipping",
     "shipsFrom": "Ships from %{country}",
     "shipTo": "Ship to:",
--- a/js/templates/modals/listingDetail/listing.html
+++ b/js/templates/modals/listingDetail/listing.html
@@ -279,4 +279,6 @@
     <% if (!ob.termsAndConditions) { print(`<i class="clrT2">${ob.polyT('listingDetail.noTermsAndConditions')}</i>`) } %>
   </div>
 
+  <div class="js-moreListings"></div>
+
 </div>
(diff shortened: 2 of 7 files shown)


morebrownies:
Thank you for adding this @jjeffryes

morebrownies:
Looks good to me @jjeffryes

morebrownies:
There seems to be an issue with the back button. I have to hit back twice after clicking a listing within "More by [Name]".

morebrownies:
It would be better to show 8 random listings @rmisio.

rmisio:
> It would be better to show 8 random listings @rmisio.

Cool, I'll work that in. Eventually, if we're feeling really ambitious, we could work it into the randomizer where more weight is given to listing with the same category or tags... but, for now, I'll leave that as a very stretchy goal at best.

morebrownies:
> Cool, I'll work that in. Eventually, if we're feeling really ambitious, we could work it into the randomizer where more weight is given to listings with the same category or tags... but, for now, I'll leave that as a very stretchy goal at best.

Sounds good!

jashot7:
@rmisio @jjeffryes I'm getting the same issue @morebrownies is getting with the back button needing to be pressed twice.

When pressing on one of the results in 'more by', the user navigates to the listing, but the URL displays the store URL rather than that for the specific listing. 

screen shot 2018-08-27 at 4 38 16 pm

Then, when the user hits the back button - the page reloads that same listing, but with the correct URL.

screen shot 2018-08-27 at 4 38 36 pm

rmisio:
closing in liu of #1544

PR #1542, opened 2018-08-24
Add store filter by Listing Contract Type & Update Disabled Logic

jashot7:


js/views/userPage/TypeFilter.js
```diff
@@ -0,0 +1,91 @@
...
+  onChangeType(e) {
+    this._state.selected = e.target.value;
+    this.trigger('type-change', { value: $(e.target).val() });
+  }
+
+  get selectedType() {
+    return this._selected;
+  }
+
+  getState() {
```
rmisio: @jashot7 I'll give this a full review next week, but one quick mention is that setState, getState and replaceState don't need to be declared here. They are already available via the baseVw which all views inherit from. The one caveat is that there is no explicit replaceState, but you can all setState with a `replace` option, e.g. `setState(newState, { replace: true })`.

(I know there are a lot of views in the codebase that implement these methods. That's because for the longest time we were duplicating the code everywhere. When we finally put them in the baseVw, we didn't remove them from everywhere because we weren't sure if any had custom logic. So... we took the stance that going forward we'd use the baseVw ones (unless we needed to override with custom functionality) and for old ones we'd refactor as we went along.)

js/views/userPage/TypeFilter.js
```diff
@@ -0,0 +1,91 @@
...
+      types: [],
+      selected: 'all',
+      expanded: false,
+      maxInitiallyVisibleTypes: 6,
+      ...(options.initialState || {}),
+    };
+
+    super(opts);
+    this.options = opts;
+    this._state = opts.initialState;
```
rmisio: This is not needed. Since the initial state is being passed into `super`, the `baseVw` constructor will take care of this.

js/views/userPage/CategoryFilter.js
```diff
@@ -70,6 +74,10 @@ export default class extends BaseVw {
   }
 
   render() {
+    if (this.getState().categories.length === 1) {
```
jjeffryes: This should be checking if the length is zero, because "all" is added in the template. If the length is 1, you actually have 2 choices, "all" and the 1 category, and it shouldn't be disabled.

jjeffryes: I'm not sure disabling it is the right thing to do in any case. If the only option is All, it's automatically selected and the user can't change it, so it doesn't hurt anything.

js/views/userPage/Store.js
```diff
@@ -329,11 +331,21 @@ export default class extends BaseVw {
       (this._$listingsContainer = this.$('.js-listingsContainer'));
   }
 
+  get $shippingFilterContainer() {
```
jjeffryes: Should this, and the other gets, be using getCachedEl instead?

I assume you're just replicating what's in use, it's ok to update this to use getCachedEl. If you want to replace the other gets in this file with getCachedEl that would be nice.

js/views/userPage/TypeFilter.js
```diff
@@ -0,0 +1,91 @@
...
+      'change input[type="radio"]': 'onChangeType',
+    };
+  }
+
+  onClickShowMoreLess() {
+    this.setState({ expanded: !this.getState().expanded });
+  }
+
+  onChangeType(e) {
+    this._state.selected = e.target.value;
```
jjeffryes: You're setting state.selected here, but retrieving this._selected in get selectedType. I don't think selectedType is used anywhere?

js/collections/Listings.js
```diff
@@ -41,4 +41,19 @@ export default class extends Collection {
...
+   */
+  get types() {
+    const types = [];
+
+    this.models.forEach(listing => {
+      const type = listing.get('contractType');
+      if (types.indexOf(type) === -1) types.push(type);
+    });
+
+    return types.sort();
```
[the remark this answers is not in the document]
rmisio: That's pretty nifty Josh.

One minor thing though is I think it could do without the concat:

`return [...new Set(this.pluck('categories'))].sort()`

js/templates/userPage/typeFilter.html
```diff
@@ -0,0 +1,50 @@
...
+           ob.types.slice(ob.maxInitiallyVisibleTypes - 1).forEach((type, index) => {
+             var flatType = type.replace(/\s/g, '-'); // remove spaces
+         %>
+         <div class="btnRadio">
+           <input type="radio"
+                  name="filterListingType"
+                  value="<%= type %>"
+                  id="<%= `filterListingType${flatType}` %>"
+                  <% if (ob.selected === type) print('checked') %>>
+           <!-- <label for="<%= `filterListingType${flatType}` %>"><%= type %></label> -->
```
rmisio: Do we need the commented line?

js/views/userPage/TypeFilter.js
rmisio: ah @jashot7, I see what you did. You were just copying the CategoryFilter which was made before all this stuff was moved into the baseVw. If you have some time, I'd just delete the overridden methods (and replaceState) and see if it works with the baseVw. It probably will and would make the code-base cleaner.

js/views/userPage/CategoryFilter.js
rmisio: > This should be checking if the length is zero, because "all" is added in the template. If the length is 1, you actually have 2 choices, "all" and the 1 category, and it shouldn't be disabled.

I think J just followed what I did in the Category filter. My logic was probably that even though you have two choices (`All` and `Something`), they both evaluate to the same result set, so it didn't make sense to even allow the toggle.

> I'm not sure disabling it is the right thing to do in any case. If the only option is All, it's automatically selected and the user can't change it, so it doesn't hurt anything.

That's true. We could remove these checks if you want.

js/views/userPage/TypeFilter.js
rmisio: Yeah, that's a copied over oversight from `CategoryFilter.js`. We could probably also remove `get selectedCat() {` from that view.

rmisio: Another thing we've changed going forward is generally we don't modify the `_state` object directly. We do it via `setState`. Up to you if you want to update things in that regard. (I know you were just using a file as a template that was probably written a year and a half ago)

rmisio:
@jashot7 I see exactly what you mean about the pain of not abstracting this one. TBH, now that I look at it, I think I would have abstracted this into a RadioFilter common view or maybe even a RadioCheckboxFilter view since I'm not seeing how they differ so much, aside from the checkbox allowing multiple selections.

But... I don't think it's worth it now. Like I mentioned, this is the first filter addition in nearly a year. Now, if we get another one, then we might want to abstract.

js/collections/Listings.js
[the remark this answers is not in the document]
jjeffryes: I think you need the concat because each categories is an array, without it you get a set of arrays.

js/views/userPage/CategoryFilter.js
jjeffryes: You could have 1 listing with the "something" category, and 1 with no category on it. In that situation "all" gets both listings, and "something" gets only 1 listing.

So you need to keep both all and something as choices if there is only 1 category. Unless we also check to see if all the listings have the one category?

js/collections/Listings.js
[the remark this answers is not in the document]
rmisio: oh right, didn't realize those were arrays.

But you still need another spread operator on the Set which converts the return value to an array. Also, the position of the sort needs to slightly move so it operates on the result of the concat.

This should work:

`return [...new Set([].concat(...this.pluck('categories')).sort())]`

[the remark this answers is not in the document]
rmisio: either way, it's a pretty clever one-liner!

js/views/userPage/TypeFilter.js
```diff
@@ -0,0 +1,56 @@
...
+import loadTemplate from '../../utils/loadTemplate';
+import BaseVw from '../baseVw';
+
+export default class extends BaseVw {
+  constructor(options = {}) {
+    const opts = {
+      ...options,
+    };
+
+    opts.initialState = {
```
rmisio: Just as an FYI, you could have included the `initialState` key inside the `opts` declaration above (after the `...options`).

js/views/userPage/TypeFilter.js
```diff
@@ -0,0 +1,56 @@
...
+      types: [],
+      selected: 'all',
+      expanded: false,
+      maxInitiallyVisibleTypes: 6,
+      ...(options.initialState || {}),
```
rmisio: I used to do this as well, but I've found the spread operator to be quite smart where if it's not an object, it no-ops, which is what we want, so this could have just been `...options.initialState`.

js/views/userPage/TypeFilter.js
jashot7: That makes sense. Do you think it might be less clean that way though? 🤔

js/views/userPage/TypeFilter.js
jashot7: Gotchya. Yeah, I took this from existing code, but I didn't actually realize it works that way - that spread operator is kickass.

js/views/userPage/TypeFilter.js
rmisio: I think it's cleaner, but i imagine it's a matter of taste... so totally up to you.

PR #1563, opened 2018-09-19
Multi wallet

### LINKED ISSUE #1425: Multicoin Integration - Wallet
## User Story
As a multicoin wallet user, I want to be able to see all the coins I use in my wallet, and see the balance and transactions for each one separately.

It should be easy for me to send and receive funds for any coin the multicoin wallet supports manually. My node should be able to handle transactions in any coin the multicoin wallet supports without needing action from me.

As a user, I should not need to worry about adding or removing coins, my experience should be focused on interacting with the coins in my wallet that are active. My wallet should always be able to handle new transactions in coins it supports.

## Functional Requirements
The multicoin wallet will always support all coins it can support. The user won't ever turn coins on or off. If they don't want to use a particular coin they can simply not create listings that accept that coin.

Note: we will eventually support wallet plugins, but there would never be two wallets with the same coin running at the same time. If a plugin wallet handles a coin the multicoin wallet handles, the plugin would be used instead.

- [x] The wallet screen will need a way to select the current coin it is displaying.
- [x] The send and receive buttons should apply to the current coin being viewed.
- [x] There should be some way to see all your wallet balances at once, so you don't have to check them one by one.
    - This could be an "all coins" view, or a sidebar with just the coins and their balances, or the balance could be shown next to each coin name in the control to select a coin screen.
- [x] Add a coin parameter to the API calls detailed below.
- QUESTION: should there be a Shapeshift feature to move funds from one of your coin wallets to another easily, without having to manually get a receive address and enter it into a shapeshift dialog (ie: a quick send X of coin Y to wallet Z)?
    - an example is the Jax wallet, or the Exodus wallet.
- [x] The resync button should be moved to the wallet screen, and apply to one wallet at a time. It should have the same restrictions on once per hour it does now.
- [x] The client config for coins should be updated if there are new coins.

## API Changes
The config will need to return a list of coins/wallets that are available.

The following calls will need a coin parameter in data or the path.
ex: wallet/BTC/address or coin: BTC
- GET wallet/address
- GET wallet/balance
- GET wallet/transactions
- GET wallet/status
- GET wallet/fees
- POST wallet/spend
- POST wallet/bumpfee
- POST wallet/resyncblockchain
- the wallet socket for new transactions will need to send a coin type.
- the wallet socket for first confirmation on a transaction will need to send a coin type.
- the walletUpdate socket will need to return an object of data keyed to coin types or separate socket data would need to be delivered for each wallet.

We might want the following calls to accept an array of coins, so a single call can get the data for multiple coins.
- GET wallet/balance
    - this could be an 'all' parameter or just no coin parameter

rmisio:
This is the base branch for the multi-wallet client integration. It should be run with the `multiwallet` server branch.

closes #1585
closes #1425


js/start.js
```diff
@@ -496,38 +497,16 @@ function ensureValidSettingsCurrency() {
...
-
-      unsupportedCryptoCurDialog.$('.js-connect')
-        .on('click', () => app.connectionManagmentModal.open());
-
-      serverConnectEvents.once('connected', () => unsupportedCryptoCurDialog.remove());
-
-      return;
-    }
-
+    app.serverConfig.walletsHelp = 'CAUTION! - Rather than reading the wallet curs directly ' +
```
rmisio: Is this helpful at all? If you were exploring `app.serverConfig` in the JS console, would you notice this warning?

js/templates/modals/listingDetail/listing.html
```diff
@@ -106,23 +106,34 @@ <h2 class="flexExpand js-cryptoTitle cryptoTitle"></h2>
...
+                  ob.crypto.anySupportedByWallet([priceCryptoCode.code, priceCryptoCode.testnetCode])
+                ) &&
                 ob.currencyMod.getCurrencyValidity(ob.price.currencyCode) !== 'VALID') {
+                // Ensure we have exchange rate data based on the currency the listing is listed
+                // in. Otherwise, it is not purchaseable since the price cannot be determined.
+                // If the listing price code is one of the accepted currencies AND the user pays in
+                // that currency, then exchange rate data is not needed.
+                // TODO:
+                // TODO:
+                // TODO: ugh - how to handle that case?
```
rmisio: This is a pretty interesting scenario, albeit quite the edge case, so we should probably take a lower scope path.

Anyhow, the idea is that if a listing is priced in a certain currency we don't have exchange rate data for then the Buy Now button is disabled. But, if the listing happens to be priced in one of the accepted currencies (and the accepted currency is supported by the user's wallet), then as long as the user pays in that currency, the purchase would work. So maybe in this case, we leave the button enabled:

image

And then in the Checkout, maybe we just disable all options other than the one in the pricing currency? Maybe with a little 'exchange rate unavailable` text, similar to how `Not available` is used when your wallet doesn't support one of the accepted currencies?

image

(just want to stress to not put too much effort into this because most listings are probably priced in fiat AND the exchange rate API shouldn't be down too often (hopefully))

js/start.js
jjeffryes: When would you see this in the console?

js/templates/modals/listingDetail/listing.html
jjeffryes: I think it would be nice to allow like to like crypto purchases (ie if the listing is priced in a crypto coin, I can buy it with that crypto coin), since that would allow the network to keep working if the ticker was down, but not if it adds a lot of complexity.

Your idea seems right to me, we'd just disable all the currencies that won't work with a note next to each explaining why.

js/start.js
rmisio: if a developer is examining `app.serverConfig` in the console. Rather than them write code that directly references `app.serverConfig.wallets`, the preference is for them to get the curs from a utility function

js/data/walletCurrencies.js
```diff
@@ -0,0 +1,372 @@
...
+// TODO: unit test this bad boy
+// TODO: unit test this bad boy
+// TODO: unit test this bad boy
+// TODO: unit test this bad boy
+// TODO: unit test this bad boy
+// TODO: unit test this bad boy
+/**
+ * Returns a list of the crypto currencies supported by the wallet.
+ *
+ * @param {object} [options={}] - Function options
```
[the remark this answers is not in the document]
rmisio: @jjeffryes The bracket notation on the variable name signifies that it's optional:
http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values

The type is driven by the preceding fragment (`{object}` in this case).

js/views/modals/wallet/CoinNavItem.js
```diff
@@ -0,0 +1,63 @@
...
+
+  onClick() {
+    if (!this.getState().active) {
+      this.trigger('selected', { code: this.getState().code });
+    }
+  }
+
+  render() {
+    const state = this.getState();
+    this.$el.toggleClass('active', state.active);
```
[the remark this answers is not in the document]
rmisio: Maybe, but that then gives you an extra wrapping element which, not only is kind of ugly, but also doesn't play nice with stuff like `padMdKids borderStacked`.

I'll keep this in mind though and maybe circle back to it when I make some other optimizations.

PR #1564, opened 2018-09-20
menu.popup options object should be optional (fix #1560)

### LINKED ISSUE #1560: Error on address bar right click event
Brief Description: 
When you do a right click on the address bar it appears the follow error
Operating System (OS and version): 
Arch Linux updated to 12/09/2018
Electron 2.0.8
OpenBazaar version 2.2.5: 

Steps to reproduce:
1. Open openbazaar and server part
2. right click on address bar

schermata del 2018-09-12 13-26-13

M0Rf30:
As in title, fix an issue with latest electron versions as described here

[menu.popup options object should be optional](https://github.com/electron/electron/issues/12915)

### DIFF
--- a/main.js
+++ b/main.js
@@ -142,7 +142,7 @@ crashReporter.start({
   productName: 'OpenBazaar 2',
   companyName: 'OpenBazaar',
   submitURL: 'http://104.131.17.128:1127/post',
-  autoSubmit: true,
+  uploadToServer: true,
   extra: {
     bundled: isBundledApp,
   },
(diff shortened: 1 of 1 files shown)


jjeffryes:
We haven't upgraded to 2.0 yet, but this change is compatible with the current code, and it doesn't hurt to preemptively fix it.

jjeffryes:
Thank you for submitting this, @M0Rf30

placer14:
@M0Rf30 , would you mind reaching out to us in slack or via email? We'd like to send a small token of appreciation but aren't sure where to send it. You can find my [email here](https://gist.github.com/placer14/b7bb0c3c0510af1cfda3c474cc923a3a).

PR #1565, opened 2018-09-20
Update page controls

### LINKED ISSUE #1541: Change pagination buttons in Discover
**Current**
screenshot 2018-08-23 22 53 04

**Ideal (matches OpenBazaar.com)** 
screenshot 2018-08-23 22 51 40

jjeffryes:
This updates the pageControls view to take a pair of parameters to show/not show the X of Y count and to use a style like discover has on openbazaar.com.

The transaction pages will continue to use the current style, we could update those to this style if we want. Any thoughts there @morebrownies ?

This also adds the localizeNumber function to the template helper, so it can be used in the results view and the pageControl view.

Closes #1541


js/templates/components/pageControls.html
```diff
@@ -1,10 +1,12 @@
 <%
-  var countsAvailable = false;
```
rmisio: Can we make a seperate component for this style of page controls? I think the data required is quite different than the other style and I just am finding it confusing to blend the two.

Really, the only state the new style needs is `curPage` and `morePages: true/false`. I think it's important that the view is dumb about the curPage rather than implementing logic that forces a set page size. I could think of examples where a list gets new data added to the top of it where you'd want to let the first page grow without affecting the subsequent pages.

js/views/components/PageControls.js
```diff
@@ -37,32 +33,11 @@ export default class extends baseVw {
     this.trigger('clickPrev');
   }
 
-  getState() {
```
rmisio: thanks for this!

js/views/search/Results.js
```diff
@@ -89,7 +89,7 @@ export default class extends baseVw {
     this.$resultsGrid.html(resultsFrag);
     // update the page controls
     // this.$displayText.html(app.polyglot.t('search.displaying', { start, end, total }));
```
rmisio: do we need the commented line?

js/templates/components/pageControlsSearch.html
```diff
@@ -0,0 +1,35 @@
...
+    if (ob.start > 1) disabledPrev = false;
+    if (ob.end < ob.total) disabledNext = false;
+  }
+%>
+
+<div class="floR">
+  <div class="pageControls flexVCent gutterH tx5">
+    <div class="btnStrip">
+      <button class="btn clrP clrBr pagePrev js-pagePrev <% if (disabledPrev) print('disabled') %>">
+          <span class="txUnb">< <%= ob.polyT('pageControls.previous') %></span>
```
rmisio: Can this view get it's own translation key rather than piggy-backing onto the other `pageControls` component?

js/views/components/PageControls.js
```diff
@@ -34,7 +34,9 @@ export default class extends baseVw {
   }
 
   render() {
-    loadTemplate('components/pageControls.html', (t) => {
+    const path = this.getState().type === 'search' ?
+      'components/pageControlsSearch.html' : 'components/pageControls.html';
+    loadTemplate(path, (t) => {
```
rmisio: Can we actually make a seperate view for the other page controls? I'm not sure why we'd want them to share one view.

Also, I'm not sure if `pageControlsSearch` is the best name for the second component since we may want to use it in other areas than search. Unless by that name, you mean it's "search style" page controls. I guess I could buy that if that's your intent.

jjeffryes: I couldn't think of a good abstract name for it. Maybe pageControlsTextStyle?

Re: view, the view functionality is literally identical other than the template, having 2 seems like unnecessary duplication to me.

PR #1566, opened 2018-09-21
Remove shapeshift multi

jjeffryes:
This removes the ShapeShift code, since ShapeShift has ended support for the ShapeShift button we use.

There are still references to ShapeShift in the coin data that should be removed.


rmisio:
@jjeffryes Can you resolve the conflict. It's probably because I moved the shapeshift icon class, so I imagine it could be deleted from the file I moved it from.

If that's the only issue, please feel free to merge this guy once it's resolved.

PR #1567, opened 2018-09-21
Add Linux troubleshooting

jjeffryes:
This add some instructions to the readme for fixing an issue that can occur on Linux when trying to start the client.


jjeffryes:
@tyler-smith I'm asking you for a review too, just to make sure I have the Linux instructions down correctly. They should be the same as what we talked about in Slack.

tyler-smith:
LGTM.

I haven't test the Arch stuff but the other commands should work.

PR #1568, opened 2018-09-24
Multi Donations

### LINKED ISSUE #1559: Add BCH Tab to Donations in About
The donation address can be gotten from Sam.

image

jjeffryes:
This adds inner tabs to the donation tab of the About modal, one for BTC donations, one for BCH donations, and updates everything to match the current designs.

It also adds some classes for the new style of flex buttons that can be used at the top of a container.

Closes #1559


js/views/modals/about/Donations.js
```diff
@@ -1,69 +1,107 @@
...
 let hiderTimer;
 
 export default class extends baseVw {
   constructor(options = {}) {
+    const opts = {
+      ...options,
+      initialState: {
+        showCoin: 'BTC',
+      },
+      ...options.initialState,
```
rmisio: you want this two lines up... right?

js/views/modals/about/Donations.js
```diff
@@ -1,69 +1,107 @@
...
+    this.dCoins = {
+      BTC: {
+        obDonationAddress: btcAddress,
+        qrCodeDataURI: qr(`bitcoin:${btcAddress}`,
+          { type: 6, size: 6, level: 'Q' }),
+        supported: isSupportedWalletCur('BTC'),
+      },
+      BCH: {
+        obDonationAddress: bchAddress,
+        qrCodeDataURI: qr(`bitcoin:${bchAddress}`,
```
rmisio: Does bch also have a `bitcoin:` preface in the QR code?

js/views/modals/about/Donations.js
```diff
@@ -1,69 +1,107 @@
...
+          { type: 6, size: 6, level: 'Q' }),
+        supported: isSupportedWalletCur('BCH'),
```
rmisio: Can this be `walletSupported` or something less generic?

rmisio: Also, I think you'd want to conditionally pass in the testnet code if we're on testnet, otherwise the buttons will always be false on testnet.

You could hard-code the mainnet code with a `T` preface if you want. But, that breaks the whole motif where the testnet code is driven by thee config. So, to keep that in play, you could make a `ensureTestnetCode` as an inverse of `ensureMainnetCode`. But, completely up to you - it may be overkill - as I'm pretty sure the server just mashes a `T` onto a mainnet code to get testnet.

js/views/modals/about/Donations.js
```diff
@@ -1,69 +1,107 @@
...
       clearTimeout(hiderTimer);
     }
     hiderTimer = setTimeout(
-      () => this.$copyNotification.removeClass('active'), 3000);
-  }
-
-  get $copyNotification() {
-    return this._$copyNotification ||
-      (this._$copyNotification = this.$('.js-copyNotification'));
+      () => this.getCachedEl('.js-copyNotification')
```
rmisio: Could you store a reference to the `this.getCachedEl('.js-copyNotification')` you add the active class to above and then remove class only on that reference? Otherwise, it's a major edge case, but it looks like if the user switches tabs in-between the animation, the class removal would be on the element in the new tab.

js/languages/en_US.json
```diff
@@ -453,13 +453,11 @@
...
-      "bchWarning": "Do not send Bitcoin Cash to this address.",
-      "unableToOpenInWallet": {
-        "title": "Unable to open in wallet",
-        "body": "Unable to open the address in the wallet because a payment is in the process of being sent. Please wait until the send completes and try again."
-      }
+      "bannerTitleBTC": "Scan or send Bitcoin (BTC) to this address",
+      "bannerTitleBCH":"Scan or send Bitcoin Cash (BCH) to this address",
+      "BCHWarning": "Do not send Bitcoin (BTC) to this address.",
+      "BTCWarning": "Do not send Bitcoin Cash (BCH) to this address.",
+      "coinWarning": "You are using a %{walletCoin} wallet, this address can only accept %{donationCoin}."
```
rmisio: Just a reminder to remove this.

js/languages/en_US.json
```diff
@@ -453,13 +453,11 @@
       "sectionHeader": "Donations",
```
rmisio: The way the underline class is being applied on the hover of the button is making the button text rise a couple pixels.

js/views/modals/about/Donations.js
```diff
@@ -1,69 +1,107 @@
-import app from '../../../app';
+import { clipboard } from 'electron';
```
rmisio: nice!

js/languages/en_US.json
rmisio: FWIW, that has nothing to do with that line of code.

js/views/modals/about/Donations.js
jjeffryes: Whoops, yes.

js/views/modals/about/Donations.js
jjeffryes: We want it to always be false on testnet. You can't send donations from a testnet wallet.

js/views/modals/about/Donations.js
jjeffryes: Good catch.

js/languages/en_US.json
jjeffryes: :)

I'll take a look at it.

rmisio: Still getting this. It happens on the in-active coin nav.

jjeffryes:
@rmisio I haven't looked at the hover issue yet.

js/templates/modals/about/about.html
```diff
@@ -15,7 +15,7 @@ <h1 class="h3 txUp clrT2"><%= ob.polyT('tabMenuHeading') %></h1>
       <div class="contentBox padMd clrP clrBr clrSh3">
         <div class="flexVCent">
           <div><%= ob.brandingBoxT() %></div>
-          <h2 id=tabTitle class="h3 clrT txCtr flexExpand txUnl"></h2>
+          <h2 id="tabTitle" class="h3 clrT txCtr flexExpand txUnl"></h2>
```
jjeffryes: I just noticed this didn't have quotes around the ID value, and wanted to fix it. Didn't seem like it needed a separate PR.

PR #1580, opened 2018-10-19
Update year in the license file

StrikerRUS:


jjeffryes:
Thank you for catching that, @StrikerRUS

PR #1582, opened 2018-10-19
Crypto cur selector tweaks

rmisio:
@jjeffryes I tweaked the setState a little bit of the crypto cur selector component. The main reason is that I discovered in the other PR there were cases when the processedCurrencies needed to either be recreated or updated and they weren't.

Anyhow, the change-set looks bigger than it seems, since I really just moved a lot of stuff around.

### DIFF
--- a/js/views/components/CryptoCurSelector.js
+++ b/js/views/components/CryptoCurSelector.js
@@ -1,35 +1,11 @@
-import _ from 'underscore';
 import $ from 'jquery';
 import app from '../../app';
 import loadTemplate from '../../utils/loadTemplate';
-import { polyTFallback } from '../../utils/templateHelpers';
 import { isSupportedWalletCur } from '../../data/walletCurrencies';
 import baseVw from '../baseVw';
 
 export default class extends baseVw {
   constructor(options = {}) {
-    const controlTypes = ['checkbox', 'radio'];
-
-    const controlType = options.initialState.controlType;
-    if (controlType && !controlTypes.includes(controlType)) {
-      throw new Error('If provided the controlType must be a valid value.');
-    }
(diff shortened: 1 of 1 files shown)


js/views/components/CryptoCurSelector.js
```diff
@@ -1,35 +1,11 @@
...
 import $ from 'jquery';
 import app from '../../app';
 import loadTemplate from '../../utils/loadTemplate';
-import { polyTFallback } from '../../utils/templateHelpers';
 import { isSupportedWalletCur } from '../../data/walletCurrencies';
 import baseVw from '../baseVw';
 
 export default class extends baseVw {
   constructor(options = {}) {
-    const controlTypes = ['checkbox', 'radio'];
```
rmisio: I moved these state data integrity checks to set state, so basically the rest of set state could assume good data instead of always checking.

js/views/components/CryptoCurSelector.js
```diff
@@ -82,68 +58,79 @@ export default class extends baseVw {
...
-        curState.processedCurs : [],
+      currencies: Array.isArray(state.currencies) ?
+        [...new Set(state.currencies)] : curState.currencies,
     };
 
-    // De-dupe any passed in currencies.
-    if (state.currencies && Array.isArray(state.currencies)) {
-      processedState.currencies = [...new Set(state.currencies)];
+    // Remove any disabled currencies from the active list.
+    if (state.activeCurs || state.disabledCurs) {
```
rmisio: I started using the pattern a lot here where, rather than using `isEqual` to test if things changed, I just assume if they are passing something in, it's changed.

PR #1585, opened 2018-10-23
Multi wallet edit listing

### LINKED ISSUE #1426: Multicoin Integration - Listing Details
## User Story
As a multicoin wallet buyer, it should be obvious to me which coins a listing accepts, and I should be able to purchase it using any of those coins.+

As a multicoin wallet seller, I should be able to set each of my listings to accept all or a subset of the coins supported by my wallet. I should be able to have different accepted coins for each listing.

### Listing Details
- [ ] Make sure the current display of the accepted currency handles multiple currencies at once.
- [ ] Make sure the current code that blocks clicking the buy button if the buyer has the wrong currency works with multicoin wallets seller listings.
    - The buyer could have a single currency SPV wallet, or an older version of the app that doesn't have a matching currency. ie: if v2.8 adds BCG, the listing accepts only BCG, and the buyer is using v2.7.

### Edit Listing
- [x] When editing a listing, the seller should be able to select one or more of the coins their own wallet supports as the listing's accepted currency/currencies.
    - The control should default to using the user's saved list of preferred coins, or all coins supported by their wallet if they don't have a list.
    - The minimum currency is one, the maximum is the full set of currencies supported by the multicoin wallet.
    - The user can't add coins to be accepted that they can't actually accept.
    - Cryptocurrency listings can only accept one purchase currency.

## API Changes
- [x] ob/listing
    - An array of accepted coins should be sent when creating or editing the order.
    - The list should be validated so it's not possible to add coins that can't be accepted.
    - The server should validate that cryptocurrency listings have exactly one accepted currency.

### LINKED ISSUE #1571: The Edit Listing modal takes too long to load the first time.
The first time the edit listing modal loads it takes a few seconds to appear. It appears that the bottleneck is probably that it's going to the translation file for the display names of hundreds of crypto currencies.

Many of them are missing translations. Not sure if missing ones slow things down more than present ones..?

In either case, here are some potential improvements:
- Upon start-up, get the translations in a Web Worker (i.e. a seperate thread), so they will likely already be there by the time the Edit Listing modal (or any other interested party) needs them. A seperate thread should prevent the process from slowing anything else down.
- Reconfigure the Edit Listing modal so they are not retrieved  immediately, but only once you start to make a crypto listing. Non-crypto listings should not be slowed down due to data required solely for crypto listings.

rmisio:
closes #1571 
closes #1426


js/views/modals/editListing/CryptoCurrencyType.js
```diff
@@ -1,18 +1,107 @@
...
+    const receiveCur = this.options.getReceiveCur();
+
+    if (receiveCur && !this.receiveCurs.includes(receiveCur)) {
+      // if the model has the receiving currency set to an unsupported cur,
+      // we'll manually add that to the list of available options. Upon a
+      // a save attempt, the user will be presented with an error prompting them
+      // to select a valid currency.
+      this.receiveCurs.push(receiveCur);
+    }
+    this.receiveCurs.push('FAT-BILLY');
```
jjeffryes: Is this test code?

js/views/modals/editListing/EditListing.js
```diff
@@ -1437,10 +1482,35 @@ export default class extends BaseModal {
...
+            ...supportedWalletCurs(),
+          ],
+          activeCurs,
+          sort: true,
+        };
+        if (this.cryptoCurSelector) {
+          this.cryptoCurSelector.remove();
+          curSelectorInitialState = {
+            ...curSelectorInitialState,
+            ...this.cryptoCurSelector.getState(),
```
jjeffryes: Does this work after it's removed?

rmisio: It should. Removal is just dom cleanup mainly and I'm not aware of any of our code clearing state on remove. But, I could re-arrange to make it less confusing.

PR #1587, opened 2018-10-26
Fix typos and incorrect grammar

agarwalrounak:


jjeffryes:
Thanks @agarwalrounak !

PR #1588, opened 2018-10-29
Multi purchase

### LINKED ISSUE #1492: Multicoin Integration - Purchase
## User Story
- As a buyer, when I make a purchase I want to be able to choose the compatible coin from one of my wallets I use to pay for the purchase.
- As a buyer, it would be useful to know what my balance is when choosing a coin, so I don't pick a coin that doesn't have enough funds.
- As a buyer, I only want to see moderators that want to moderate orders using the coin I chose to use to pay for the purchase.

## Functional Requirements
- [ ] There should be a control that allows the buyer to pick which of their coins they want to use when purchasing this listing.
    - They should only be able to choose a coin that is compatible with the list of coins the listing can accept.
        - The control element can either only show compatible coins, or incompatible coins can be disabled.
    - It would be nice to indicate how much funds each coin has when selecting it, so the buyer doesn't select a coin they can't pay with. This gets tricky because they may not know the total price yet at this point, and don't have any way to know what the fee will be.
    - Choosing the coin affects which moderators can be selected (some of the moderators on the order may not support the chosen coin. ex: an old BTC SPV only moderator on a listing that accepts BTC and BCH).
    - We may want to encourage the user to choose the coin before they choose the moderator (if any) to avoid confusing interactions.

- [ ] Only moderators compatible with the chosen payment coin should be shown.
    - Compatible means the coin is on their accepted coin list. This should only be an issue in the near future for mods that haven't upgraded to multi wallet nodes yet.
    - QUESTION: Should incompatible moderators be hidden, or disabled? It might be useful for a buyer to know if they pick a different coin which moderators will be selectable.
    - [ ] The user may pick a moderator, then choose a coin that moderator can't accept. Doing that should de-select the moderator and revert the order to unmoderated. The user should probably be warned if this happens so they don't just click pay assuming the moderator is still selected.
        - alternately, we could have a "moderated payment, but no moderator selected" state, and show an error if the buyer has no moderator selected and tries to pay.

## API Changes
- [ ] ob/purchase
    - Each item will need a coinType (coinCode?) field.
    - When the purchase is made, the moderator (if any) should be validated to confirm they are capable of accepting the coin used for payment.
- [ ] ob/estimatetotal
   - send a coin parameter
- [ ] wallet/estimatefee see #1425
    - send a coin parameter

jjeffryes:
This adds currency selection to the purchase flow.

There is also some refactoring here, such as cleaning up how turning the show verified moderators only toggle works and removing the purchase model, which wasn't really used for anything.

Closes #1492


js/models/purchase/Order.js
```diff
@@ -12,16 +13,17 @@ export default class extends BaseModel {
...
       countryCode: '',
-      addressNotes: '',
-      moderator: '',
       items: new Items(),
-      alternateContactInfo: '',
+      moderator: '',
+      paymentCoin: '',
+      postalCode: '',
+      shipTo: '',
+      state: '',
```
jjeffryes: I just alphabetized these defaults.

js/models/purchase/Order.js
```diff
@@ -57,8 +59,19 @@ export default class extends BaseModel {
...
+    }
+
+    if (this.shippable) {
+      if (!attrs.shipTo || !(typeof attrs.shipTo === 'string')) {
+        addError('shipping', app.polyglot.t('orderModelErrors.missingAddress'));
+      }
+
+      if (!attrs.countryCode || !(typeof attrs.countryCode === 'string')) {
+        addError('shipping', app.polyglot.t('orderModelErrors.missingAddress'));
+      }
```
jjeffryes: There was a bug in the shipping validation where it only failed validation if both the shipTo and countryCode were false.

js/templates/components/moderators.html
```diff
@@ -1,11 +1,21 @@
-<div class="moderatorsWrapper clrBr <% print(ob.wrapperClasses) %> <% if (ob.showVerifiedOnly) print('showVerifiedOnly') %> js-moderatorsWrapper">
```
jjeffryes: The showing or not showing of moderator cards is handled in the render now.

js/templates/modals/purchase/payment.html
```diff
@@ -30,15 +27,15 @@
         <div class="flexRow gutterH rowLg">
           <div class="col6">
             <%= ob.processingButton({
-              className: 'btn btnThin width100 clrP clrBr clrSh2 js-payFromWallet',
-              textClassName: 'flexCent',
-              btnText: `<i class="icon">${ob.walletIconTmpl()}</i>${ob.polyT('purchase.pendingSection.payFromWallet')}`,
+            className: 'btn btnThin width100 clrP clrBr clrSh2 js-payFromWallet',
+            textClassName: 'flexCent',
+            btnText: `<i class="icon">${ob.walletIconTmpl()}</i>${ob.polyT('purchase.pendingSection.payFromWallet')}`,
```
jjeffryes: All the changes here were copied from the multi-wallet-order branch.

js/views/components/Moderators.js
```diff
@@ -331,6 +317,27 @@ export default class extends baseVw {
...
 
+  modShouldRender(model) {
+    const hideOnUnverified = this.getState().showVerifiedOnly && !model.isVerified;
+    const showCur = this.getState().showOnlyCur;
+    const hideOnCur = showCur && !model.hasModCurrency(showCur);
+    return !(hideOnUnverified || hideOnCur);
+  }
+
+  batchCardRender(model) {
+    if (!model || !(model instanceof Moderator)) {
```
jjeffryes: This solves an issue where huge amounts of renders, mostly in the settings/store view, were locking things up while mods were arriving over the socket. It will batch the renders just enough that very large sets arriving together are rendered as a group, but stragglers aren't delayed too much.

js/views/components/Moderators.js
```diff
@@ -331,6 +317,27 @@ export default class extends baseVw {
...
   }
 
+  modShouldRender(model) {
+    const hideOnUnverified = this.getState().showVerifiedOnly && !model.isVerified;
+    const showCur = this.getState().showOnlyCur;
+    const hideOnCur = showCur && !model.hasModCurrency(showCur);
+    return !(hideOnUnverified || hideOnCur);
+  }
+
+  batchCardRender(model) {
```
jjeffryes: This solves an issue where huge amounts of renders, mostly in the settings/store view, were locking things up while mods were arriving over the socket. It will batch the renders just enough that very large sets arriving together are rendered as a group, but stragglers aren't delayed too much.

(edit, moved the comment to the right line)

js/views/modals/purchase/Payment.js
```diff
@@ -1,12 +1,12 @@
 /*
-  This view is also used by the Order Detail overlay. If you make any changes, please
-  ensure they are compatible with both the Purchase and Order Detail flows.
-*/
+ This view is also used by the Order Detail overlay. If you make any changes, please
+ ensure they are compatible with both the Purchase and Order Detail flows.
+ */
```
jjeffryes: The changes in this view were copied from the multi-wallet order details branch.

js/views/modals/purchase/Purchase.js
```diff
@@ -446,19 +464,18 @@ export default class extends BaseModal {
...
           .done((data) => {
             this.setState({ phase: 'pending' });
-            this.purchase.set(this.purchase.parse(data));
             this.payment = this.createChild(Payment, {
-              balanceRemaining: this.purchase.get('amount'),
-              paymentAddress: this.purchase.get('paymentAddress'),
-              orderId: this.purchase.get('orderId'),
+              balanceRemaining: integerToDecimal(data.amount, paymentCoin),
+              paymentAddress: data.paymentAddress,
+              orderId: data.orderId,
```
jjeffryes: The only thing the purchase model was used for was to get the balanceRemaining value, so I refactored it out.

test/models/purchase/Item.js
```diff
@@ -0,0 +1,95 @@
...
+      t: (str) => str,
+    };
+  });
+
+  it('fails validation if it a non-crypto item and has a non-number quantity.', () => {
+    const item = new Item();
+    item.set({ quantity: 'a' }, { validate: true });
+    const valErr = item.validationError;
+
+    expect(valErr && valErr.quantity && !!valErr.quantity.length || false).to.equal(true);
```
rmisio: Why is the `|| false` needed?

rmisio:
This is minor, but I've seen it before... we have a double border on top:

image

Thought I had a bug for it, but can't seem to find it now.

rmisio:
Also seeing a situation where both the Mod and the Direct Order radio are selected:

image

All I did was click to show Unverified Mods and then selected the first and only mod.

rmisio:
The currency selector does not get the hand when I hover over an option. I think it probably should, but if it shouldn't, then there's an inconsistency because when you hover over a Mod or Direct Payment radio, you do get the hand.

js/models/purchase/Order.js
```diff
@@ -57,8 +59,19 @@ export default class extends BaseModel {
       addError('items', 'At least one item is required.');
     }
 
-    if (this.shippable && !attrs.shipTo && !attrs.countryCode) {
-      addError('shipping', app.polyglot.t('orderModelErrors.missingAddress'));
+    const c = attrs.paymentCoin;
```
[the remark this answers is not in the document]
jjeffryes: I took the advice of some sharp guy that told me if the code is obvious enough it doesn't need extra explanation. ;)

js/views/components/Moderators.js
```diff
@@ -381,42 +388,64 @@ export default class extends baseVw {
...
+        }
+
+        mod.setState({
+          preferredCurs: this.getState().preferredCurs,
+          ...newState,
+        }, { renderOnChange: false });
+
+        if (shouldRender) {
+          mod.delegateEvents();
+          this.getCachedEl('.js-moderatorsWrapper').append(mod.render().$el);
```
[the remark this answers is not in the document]
jjeffryes: Yes, good call.

js/views/modals/purchase/Payment.js
```diff
@@ -40,13 +40,28 @@ export default class extends BaseVw {
...
+      // pass
+    }
+
+    if (!paymentCoinData) {
+      throw new Error(`Unable to obtain wallet currency data for "${options.paymentCoin}"`);
+    }
+
     super(options);
     this.options = options;
+    console.log(`the balance remaining is ${options.balanceRemaining}`);
```
rmisio: Can we remove this before merging?

rmisio:
After I click "Pay from Wallet", I get the following exception:

image

If you copy in the latest version of Payment.js from the multi-wallet-order branch, I believe I've fixed this issue.

js/views/components/CryptoCurSelector.js
```diff
@@ -17,6 +17,7 @@ export default class extends baseVw {
     }
 
     const opts = {
+      disabledMsg: '',
```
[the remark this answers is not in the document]
jjeffryes: It's not something that should ever change, shouldn't state just handle things that track state?

js/templates/components/moderators.html
```diff
@@ -1,34 +1,36 @@
...
+            <% const showOnlyVer = ob.showVerifiedOnly && ob.unVerCount ? 'Verified' : '';
+               const showOnlyCur = ob.showOnlyCur ? 'Matching' : '';
+               const opts = ob.showOnlyCur ? { coin: ob.showOnlyCur } : {};
+               const msgPath = `moderators.noModsMsg.no${showOnlyCur}${showOnlyVer}Moderators`;
+            %>
+            <h4><%= ob.polyT(`${msgPath}.title`, opts) %></h4>
+            <% // The section below is only relevant if the moderators are loaded in a purchasing context. %>
+            <% if (ob.purchase) { %>
+              <div class="tx4 clrT2"><%= ob.polyT(`${msgPath}.body`) %></div>
+              <% if (showOnlyVer) { %>
```
jjeffryes: This was refactored to handle the multiple slightly different situations where no moderators might be visible. It handles whether only verified moderators are shown or not, whether moderators that match a currency are shown or not, and if no valid moderators at all were loaded (ie: the listing had moderators, but they have turned moderation off).

js/templates/modals/purchase/directPayment.html
```diff
@@ -0,0 +1,19 @@
+<% // duplicate the moderator card html to make sure everything aligns %>
```
jjeffryes: The direct payment button was broken out into it's own view and template so it's state could be updated without re-rendering the entire view. For example, it starts selected, but then switches to unselected when the first moderator loads.

js/views/components/Moderators.js
```diff
@@ -268,35 +260,26 @@ export default class extends baseVw {
...
-      // check if there are mods that loaded but none were valid
-      if (!this.moderatorsCol.length && this.fetchingMods.length) {
-        this.trigger('noValidModerators');
-        this.setState({ noValidModerators: true });
-      }
-      // check if no valid verified mods where loaded
-      if (this.verifiedMods.length && !this.modCount) {
-        this.trigger('noValidVerifiedModerators');
-        this.setState({ noValidVerifiedModerators: true });
-      }
```
jjeffryes: The logic above was moved to the render.

jjeffryes:
@rmisio this is ready for re-review now.

js/views/modals/purchase/Purchase.js
```diff
@@ -116,24 +115,46 @@ export default class extends BaseModal {
...
       (hashes, codes) => this.changeCoupons(hashes, codes));
 
+    const currencies = this.listing.get('metadata').get('acceptedCurrencies') || [];
+    const disabledCurs = currencies.filter(c => !isSupportedWalletCur(c));
+    this.cryptoCurSelector = this.createChild(CryptoCurSelector, {
+      disabledMsg: app.polyglot.t('purchase.cryptoCurrencyInvalid'),
+      initialState: {
+        controlType: 'radio',
+        currencies,
+        // Kludge to set it to the first alphabetically, may not always match view sorting
```
jjeffryes: I'm not actually sure of the best way to handle this, the goal is to set the initial active currency to the first one in the selector, but sometimes the order is different after the selector's set state alphabetizes things.

In this example, sometimes the currencies were ['LTC', 'BTC'], and just setting it to the first one resulted in LTC being selected, which was the second after it was alphabetized.

For now I think this kludge is ok, the ideal would be to not set activeCurs, and have the crytoCurSelector automatically select the first one after alphabetizing if no value was passed in and the controlType is radio.

js/templates/modals/purchase/directPayment.html
rmisio: That’s such a crazy UX for me. You’re telling me the purchase screen loads with Direct selected and then some amount of time later, unbeknownst to the user, it changes?

I think that’s crazy. I might glance and see direct and  check out a couple other things, suddenly by the time I go to click purchase, thinking I’m making a direct order, I got me a moderator.

@morebrownies What do you think about this?

morebrownies: Yeah I wouldn't recommending switching it on the user after more content loads. We can't guarantee they're going to see the selection change.

Either keep it as `direct` by default or don't select either option and let the user decide.

rmisio: or you select nothing and if nothing is still selected when the mod comes back, then select the mod.

If they click purchase while nothing is selected, we show an error forcing them to explicitly choose Direct or wait for the mods.

jjeffryes: I think it might be best to just simplify things and have either:
a. direct purchase is selected at first, and stays selected if the user doesn't change it once moderators have loaded. This is the simplest way to do it.

b. neither direct purchase nor any moderator is automatically selected, the user has to purposely select one before they can continue with their order. If no moderators that match the current criteria (coin, show only verified) have loaded, the user has to click on the direct payment option.

rmisio: We think alike @morebrownies - we were typing at the same time.

morebrownies: Looks like we all basically said the same thing at the same time 🥇

js/models/purchase/Order.js
```diff
@@ -57,8 +59,19 @@ export default class extends BaseModel {
...
+    if (!(c && typeof c === 'string' && isSupportedWalletCur(c))) {
+      addError('paymentCoin', app.polyglot.t('orderModelErrors.paymentCoinInvalid'));
+    }
+
+    if (this.shippable) {
+      if (!attrs.shipTo || !(typeof attrs.shipTo === 'string')) {
+        addError('shipping', app.polyglot.t('orderModelErrors.missingAddress'));
+      }
+
+      if (!attrs.countryCode || !(typeof attrs.countryCode === 'string')) {
```
rmisio: Can we also do the !== here for the typeof?

js/views/components/CryptoCurSelector.js
```diff
@@ -134,6 +139,7 @@ export default class extends baseVw {
       // If active or disabled lists are passed in, we'll assume they're
       // different and ensure the processedCurrencies list reflects them.
       processedState.processedCurs = processedState.processedCurs.map(cur => ({
+        ...cur,
```
jjeffryes: I caught a bug here, if the currencies weren't recreated the values of each one were being set to just active and disabled.

jjeffryes:
@rmisio patched up and ready for a (hopefully) final look.

js/views/components/CryptoCurSelector.js
rmisio: yeah, good catch.

PR #1589, opened 2018-10-30
Fix double negation in Donations.js

gy741:
Hello,

I think, In contexts such as an if statement's test where the result of the expression will already be coerced to a Boolean, casting to a Boolean via double negation(!!) unnecessary.

What do you think of it?

Thanks.

### DIFF
--- a/js/views/modals/about/Donations.js
+++ b/js/views/modals/about/Donations.js
@@ -30,9 +30,7 @@ export default class extends baseVw {
     clipboard.writeText(obDonationAddress);
 
     this.$copyNotification.addClass('active');
-    if (!!hiderTimer) {
-      clearTimeout(hiderTimer);
-    }
+    clearTimeout(hiderTimer);
     hiderTimer = setTimeout(
       () => this.$copyNotification.removeClass('active'), 3000);
   }


rmisio:
Actually, that whole `if` could be removed and just the `clearTimeout` left since it gracefully handles anything being passed to it.

gy741:
@rmisio Thank you for your review.
If so, remove the "if statement" and the code below would be fine?
```diff
--- a/js/views/modals/about/Donations.js
+++ b/js/views/modals/about/Donations.js
@@ -30,9 +30,7 @@ export default class extends baseVw {
     clipboard.writeText(obDonationAddress);
 
     this.$copyNotification.addClass('active');
-    if (!!hiderTimer) {
-      clearTimeout(hiderTimer);
-    }
+    clearTimeout(hiderTimer);
     hiderTimer = setTimeout(
       () => this.$copyNotification.removeClass('active'), 3000);
   }
```

rmisio:
@gy741

> If so, remove the "if statement" and the code below would be fine?

Yes, correct. And, thanks for the help!

gy741:
Due to a "rebase" error problem, a new PR was created.
Please check. 
And thanks for the kind answers. 👍

PR #1591, opened 2018-10-30
De-duplicate active currencies.

jjeffryes:
Minor tweak to de-duplicate active currencies in the crypto currency selector.

This doesn't remove the need for the server to fix the bug that allows duplicate copies of currencies to be saved, but I think it's beneficial if the control makes sure it's own data is correct regardless of where it comes from.


jashot7:
👍

PR #1592, opened 2018-10-30
Dependency updates

jjeffryes:
This updates some dependencies with vulnerabilities. The test code required some tweaking and the addition of jsdom-global, since the newest version of js-dom no longer supports putting the window, etc. in the global scope like we were doing, and I didn't think it was worth it to rewrite all the tests. 

jsdom-global makes window, document, etc. all available in the global scope for mocha to use.

https://github.com/jsdom/jsdom/wiki/Don't-stuff-jsdom-globals-onto-the-Node-global


jashot7:
> jsdom-global makes window, document, etc. all available in the global scope for mocha to use.

Nice.

test/mocha.opts
```diff
@@ -0,0 +1 @@
+-r jsdom-global/register
```
rmisio: I don't think you need this if you are setting things up in the `_setup.js` file.

package.json
```diff
@@ -53,11 +53,12 @@
     "eslint-plugin-react": "^5.2.2",
```
rmisio: So when I run `npm test` it goes through the tests, but just hangs:

image

If you try in master (you'll need to `npm install` again), it immediately exits after displaying the results.

(btw, ignore the 5 failing tests. I'm planning to fix those soonish.)

PR #1595, opened 2018-10-30
Fix Z-Index Issue In Wallet

jjeffryes:
This fixes an issue where the resync button was on top of the spend tooltip.

### DIFF
--- a/js/templates/modals/wallet/wallet.html
+++ b/js/templates/modals/wallet/wallet.html
@@ -26,8 +26,8 @@
           </div>
         </div>
         <div class="clrP clrSh3 posR">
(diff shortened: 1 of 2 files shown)


js/templates/modals/wallet/wallet.html
```diff
@@ -26,10 +26,10 @@
           </div>
         </div>
         <div class="clrP clrSh3 posR">
-          <div class="js-reloadTransactionsContainer reloadTransactions"></div>
           <div class="js-transactionsContainer"></div>
+          <div class="js-reloadTransactionsContainer reloadTransactions"></div>
```
jjeffryes: Since this div is absolutely positioned, it needs to be after the transactionsContainer div, so it will be layered on top of it.

PR #1602, opened 2018-11-09
Fix mod disclaimer link

### LINKED ISSUE #1601: Dispute guidelines link in settings not working
There is no hover state, and nothing happens when you click:

image

jjeffryes:
This removes the fauxchecked class from the moderation view in settings.

If I remember right, the original intent was to make it so that if you had turned on moderation, you couldn't turn off the checkboxes until after also turning off moderation, since there's no situation in which you'd turn on moderation, check the boxes, save, then uncheck one of the boxes and try to save again.

But that's way more complexity than is needed for a very minor UX issue. 

Closes #1601


jjeffryes:
Set to merge with wrong branch.

PR #1607, opened 2018-11-15
Enhance error reporting

jjeffryes:
This does a few things:
1. Does not collect metrics unless the user is using a bundled app, that way we are less likely to pollute the metrics when testing. We'll lose a few users that are installing from source, which is fine.

The metrics controls are hidden if the user is running from source.

2. Improves some errors in purchasing.

3. Adds more metrics around listing loading and purchasing, so we can better diagnose any issues we see.

4. Changes the collection of errors in purchasing to client-side errors are all considered different segments. This is a bit of a kludge, but better than concatenating errors into the errors segment. Concatenating made it difficult to tell how many times an error was happening, since you might have 40 records of "error1" and 23 records of "error1, error2" and 9 records of "error1, error5, error 12", all as different values of the same segment.

This makes it so each category of client-side error concatenates the errors in it, so UserError-shipping is a single segment (for example).

5. Removes the details of the listing that failed to load from the "no link named" server errors. It's not ideal to need to have to manipulate the data, but what we want with that error is how many times it happened, not dozens of individual errors with different names.

### DIFF
--- a/js/views/modals/purchase/Purchase.js
+++ b/js/views/modals/purchase/Purchase.js
@@ -56,11 +56,10 @@ export default class extends BaseModal {
     this.variants = opts.variants;
     this.vendor = opts.vendor;
     const shippingOptions = this.listing.get('shippingOptions');
-    const shippable = !!(shippingOptions && shippingOptions.length);
-
     const moderatorIDs = this.listing.get('moderators') || [];
     const disallowedIDs = [app.profile.id, this.listing.get('vendorID').peerID];
     this.moderatorIDs = _.without(moderatorIDs, ...disallowedIDs);
+
     this.setState({
       showModerators: this.moderatorIDs.length,
       showVerifiedOnly: true,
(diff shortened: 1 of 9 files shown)


js/views/modals/purchase/Purchase.js
```diff
@@ -503,23 +507,27 @@ export default class extends BaseModal {
...
-      const purchaseErrs = [];
+      const purchaseErrs = {};
       Object.keys(this.order.validationError).forEach(errKey => {
         const domKey = errKey.replace(/\[[^\[\]]*\]/g, '').replace('.', '-');
-        purchaseErrs.push(domKey);
         let container = this.$(`.js-${domKey}-errors`);
         // if no container exists, use the generic container
         container = container.length ? container : this.getCachedEl('.js-errors');
-        this.insertErrors(container, this.order.validationError[errKey]);
+        const err = this.order.validationError[errKey];
```
rmisio: Yeah, I think this is much better than just capturing the failed key.

PR #1613, opened 2018-11-19
Add a Copy Contract Link

### LINKED ISSUE #1545: Contract copying in the UI should not first require unfolding every key
Brief Description: When copying a contract for support purposes, I must expand every section first. Either expand them by default or offer a way to expand/collapse them all easily. Or allow a "click to copy contract" button to do the heavy lifting. Bonus points for making the copy of the contract on the clipboard NOT include the (+) and (-) nodes which the user clicks to expand/collapse the JSON key.

Motivation: As a developer, I would prefer the contract copy always included every key's contents so I wouldn't have to ask the user to resubmit the contract because they forgot to expand an important section before copying.

Bonus Motivation: As a developer, I would love to format/lint/have-valid-syntax in my JSON locally without having to clean it of the (+) and (-) icons in the body.

jjeffryes:
This adds a copy link to the contract in the order details.

@morebrownies I used the same style we use everywhere else for copying things, and put it below the contract. This seemed simple and obvious enough we didn't need a design for it, but we can change it if you have any suggestions.

Closes #1545

image

### DIFF
--- a/js/views/modals/orderDetail/contractTab/Contract.js
+++ b/js/views/modals/orderDetail/contractTab/Contract.js
@@ -1,5 +1,8 @@
+import $ from 'jquery';
+import { clipboard } from 'electron';
 import renderjson from '../../../../lib/renderjson';
 import BaseVw from '../../../baseVw';
+import '../../../../utils/lib/velocity';
 import loadTemplate from '../../../../utils/loadTemplate';
 
 export default class extends BaseVw {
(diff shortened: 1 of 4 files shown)


js/views/modals/orderDetail/contractTab/Contract.js
```diff
@@ -27,6 +30,31 @@ export default class extends BaseVw {
...
+          $(els[0]).removeClass('unclickable');
+        },
+      });
+    this.getCachedEl('.js-copyContractDone')
+      .velocity('stop')
+      .velocity({ opacity: 1 })
+      .velocity({ opacity: 0 }, { delay: 5000 });
+  }
+
   render() {
```
rmisio: Can we get the `super.render()` in here?

js/views/modals/orderDetail/contractTab/Contract.js
```diff
@@ -1,5 +1,8 @@
+import $ from 'jquery';
+import { clipboard } from 'electron';
 import renderjson from '../../../../lib/renderjson';
 import BaseVw from '../../../baseVw';
```
rmisio: Could we also put the BaseVw below the util mods?

I know you just followed the already out-of-order approach in place.

PR #1616, opened 2018-11-20
Add Search Tooltip

### LINKED ISSUE #1575: Add Search Tooltip
A tooltip should be added to the discover screen to match the one being added to the .com site.

https://github.com/OB1Company/Design/issues/42

jjeffryes:
This adds a tooltip to search explaining why some listings may not appear. 

Closes #1575

### DIFF
--- a/js/languages/en_US.json
+++ b/js/languages/en_US.json
@@ -1285,6 +1285,7 @@
     "suggestions": "Suggestions",
     "resultsFound": "<b>%{smart_count} listing</b> found for <b>\"%{term}\"</b> |||| <b>%{smart_count} listings</b> found for <b>\"%{term}\"</b>",
     "resultsTotal": "%{smart_count} listing |||| %{smart_count} listings",
+    "resultsHelper":"Search providers may not display all results on the network. Spam, offensive, or unreachable listings may be omitted.",
     "sortBy": "Sort by",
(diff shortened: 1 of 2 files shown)


js/languages/en_US.json
```diff
@@ -1285,6 +1285,7 @@
     "suggestions": "Suggestions",
     "resultsFound": "<b>%{smart_count} listing</b> found for <b>\"%{term}\"</b> |||| <b>%{smart_count} listings</b> found for <b>\"%{term}\"</b>",
     "resultsTotal": "%{smart_count} listing |||| %{smart_count} listings",
+    "resultsHelper":"Search providers may not display all results on the network. Spam, offensive, or unreachable listings may be omitted.",
```
rmisio: > Search providers may not display all results on the network. Spam, offensive, or unreachable listings may be omitted.

@morebrownies @SamPatt What do you think of this phrasing?

It seems a bit odd to me because it seems to kind of suggest that a search provider will display everything it sees outside of things it deems offensive... whereas some dude who's really into shoes might make a ShoeBazaar search endpoint and it feels like that use-case doesn't really jive with the phrasing of the tip.

rmisio: Well, if they had a huge objection, I would think they would have voiced there views in less than 6 days. So gonna merge.

PR #1618, opened 2018-11-20
Move cur warning

### LINKED ISSUE #1615: The disabled text in the cryptoCur component should not be disabled when the element is
image

jjeffryes:
Moves the disabled currency text outside of the label, so it's not disabled when the label is disabled.

Closes #1615


jjeffryes:
@rmisio the cursor is fixed, along with some tweaks so the padding at the left and right side is clickable now.

PR #1621, opened 2018-11-21
CSS Fix and No Auto Radio Select

jjeffryes:
This PR fixes some issues with the currency selector CSS and HTML (like no checkbox appearing), and removes the logic to always select at least one active currency if the control is set to radio behavior.

The purchase view will no longer set an active currency, requiring the user to pick one. If they don't, they'll get an error if they try to move past that screen.


rmisio:
The `<icon> <display-name>` no longer get the hand on hover whereas the rest of the row does. You could still click to select, but it's a bit off there's no hand there.

rmisio:
Can we get a `flexNoShrink` on the cryptoIcon, otherwise long text causes it to shrink?

jjeffryes:
@rmisio this is ready for re-review.

PR #1634, opened 2018-12-10
Order spend url

rmisio:
Updating the Spend model to support the new order spend url and calling the Order Spend functionality from a purchase.

### DIFF
--- a/js/models/wallet/Spend.js
+++ b/js/models/wallet/Spend.js
@@ -8,9 +8,17 @@ import { polyTFallback } from '../../utils/templateHelpers';
 import app from '../../app';
 import BaseModel from '../BaseModel';
 
+const getWalletSpendUrl = () => app.getServerUrl('wallet/spend/');
+const getOrderSpendUrl = () => app.getServerUrl('ob/orderspend/');
+
 class Spend extends BaseModel {
+  constructor(attrs, options = {}) {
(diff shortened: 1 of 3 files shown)


js/models/wallet/Spend.js
```diff
@@ -106,6 +114,14 @@ class Spend extends BaseModel {
         if (attrs.memo && typeof attrs.memo !== 'string') {
           addError('memo', 'If provided, the memo should be a string.');
         }
+
+        if (
+          this.url === getOrderSpendUrl() &&
+          typeof attrs.orderId !== 'string' ||
+          !attrs.orderId
```
jjeffryes: This error is firing when trying to send funds directly from the wallet. 

I think you want the second two conditions in parenthesis:
```
this.url === getOrderSpendUrl() &&
          (typeof attrs.orderId !== 'string' ||
          !attrs.orderId)
```

jjeffryes:
I noticed the package.lock was included with the PR, we probably want to leave those out if none of the dependencies have changed.

rmisio:
> I noticed the package.lock was included with the PR, we probably want to leave those out if none of the dependencies have changed.

I think the issue is that it's not in the repo, so after this initial commit it shouldn't be changing unless we update a package.

PR #1637, opened 2018-12-11
Use all currencies in search

### LINKED ISSUE #1636: Update Search to Send All Supported Currencies
When the search view is loaded, the default query should send all of the node's supported currencies.

This is dependent on the search endpoints being updated to accept multiple currencies.

jjeffryes:
This updates search to use all the currencies supported in the wallet. 

It also fixes the formData method so it collects checkbox data in a usable way.

This isn't actually usable until the search server has been updated to support checkboxes and multiple currencies.

To test it, you can add some fake filter data to line 492:

```
    if (data.options) {
      data.options.acceptedCurrencies = {
        label: 'test multi-currency',
        type: 'checkbox',
        options: [
          {
            checked: false,
            default: true,
            label: 'Bitcoin',
            value: 'BTC',
          },
          {
            checked: false,
            default: true,
            label: 'Bitcoin Cash',
            value: 'BCH',
          },
          {
            checked: false,
            default: true,
            label: 'Litecoin',
            value: 'LTC',
          },
          {
            checked: false,
            default: true,
            label: 'ZCash',
            value: 'ZEC',
          },
        ],
      };
    }
```

Closes #1636


js/views/search/Search.js
```diff
@@ -451,6 +456,17 @@ export default class extends baseVw {
     recordEvent('Discover_ChangeFilter');
   }
 
+  changeFilterAll(e) {
+    const targ = $(e.target);
+    this.formOverrides[targ.prop('name')] = targ.data('val').split(',') || '';
```
rmisio: Why default to an empty string here instead of an empty array? The latter keeps the types consistent.

Also, if there are no vals (edge case), then the data attribute will come in as `undefined` and this will bomb on the `split` call. So you should probably check that `targ.data('val')` is a string and only split if so.

js/views/search/Search.js
```diff
@@ -451,6 +456,17 @@ export default class extends baseVw {
...
+    this.changeFilter();
+  }
+
+  changeFilterNone(e) {
+    this.formOverrides[$(e.target).prop('name')] = '';
```
rmisio: Why not default to an empty array?

jjeffryes:
@morebrownies I think there was a design for the all and none links a long time ago, but I couldn't find it. If you want to review this, run this branch and add the staging search server as a search provider.

https://staging.search.ob1.io/listings/search

rmisio:
This LGTM, just giving @morebrownies a chance to check it out, before merging.

PR #1639, opened 2018-12-11
Get Balances for Wallet Nav Coins

### LINKED ISSUE #1638: Wallet Balances Aren't Showing
The coins in the wallet menu should have balances shown on the left, but usually don't. 

image

image

jjeffryes:
This will make sure the balances for the Nav Coins in the wallet are shown when the view loads.

Closes #1638

### DIFF
--- a/js/views/modals/wallet/Wallet.js
+++ b/js/views/modals/wallet/Wallet.js
@@ -51,7 +51,7 @@ export default class extends BaseModal {
       initialActiveCoin = navCoins.find(coin => isSupportedWalletCur(coin)) || null;
     }
 
-    // If at this point the initialActiveCoin and consequentally this.activeCoin
+    // If at this point the initialActiveCoin and consequently this.activeCoin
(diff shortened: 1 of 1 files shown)


js/views/modals/wallet/Wallet.js
```diff
@@ -80,7 +80,7 @@ export default class extends BaseModal {
       active: coin === opts.initialNavCoin,
       code: coin,
       name: app.polyglot.t(`cryptoCurrencies.${coin}`, { _: coin }),
-      balance: coin.confirmed,
+      balance: app.walletBalances.get(coin).get('confirmed'),
```
rmisio: Would you mind checking first if the balance model is there? It should be, but the server has given us quirkier things and I'm trying to more defensively program, at least on the low hangin fruit.

```
const balanceMd = app.walletBalances.get(coin);

...

balance: balanceMd && balanceMd.get('confirmed'),
```

PR #1640, opened 2018-12-12
removing polyTfallback and using native functionality

rmisio:
Removing our custom `polyTFallback` since polyglot natively offers the functionality.

### DIFF
--- a/js/data/currencies.js
+++ b/js/data/currencies.js
@@ -690,7 +690,7 @@ function getCurrencies(options = {}) {
 
   const walletCurs = (opts.includeWalletCurs ? supportedWalletCurs() : [])
     .map(cur => {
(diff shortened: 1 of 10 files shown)


js/data/currencies.js
```diff
@@ -691,7 +690,7 @@ function getCurrencies(options = {}) {
 
   const walletCurs = (opts.includeWalletCurs ? supportedWalletCurs() : [])
     .map(cur => {
-      const name = polyTFallback(`cryptoCurrencies.${cur}`, cur);
+      const name = app.polylgot.t(`cryptoCurrencies.${cur}`, { _: cur });
```
jjeffryes:
```suggestion
      const name = app.polyglot.t(`cryptoCurrencies.${cur}`, { _: cur });
```

rmisio: good catch thanks.

PR #1645, opened 2018-12-14
updating getFormData in the baseVw to handle checkbox groups

rmisio:
This augments #1637 by updating `getFormData` to handle checkbox groups.

> The old code was just adding a value of "true" if at least one checkbox was checked.

Well, in 95% of the cases, that's what we want. We simply want whether a checkbox is checked or not to map to a boolean. For example:

`<input type="checkbox" name="useTor" checked />`

would get you: `{ useTor: true }`

`<input type="checkbox" name="useTor" />`

would get you: `{ useTor: false }`

I think what you're looking for is checkbox group functionality which by convention is denoted by appending the `name` attribute with `[]`. In that scenario, the field maps to an array containing the `value` attribute of any checked checkboxes.

https://www.dyn-web.com/tutorials/forms/checkbox/same-name-group.php

This code updates `getFormData` to support that while leaving the existing checkbox to boolean functionality in place.

### DIFF
--- a/js/views/baseVw.js
+++ b/js/views/baseVw.js
@@ -87,23 +87,22 @@ export default class baseVw extends View {
         }
       }
 
-      const name = $field.attr('name');
+      let name = $field.attr('name');
       const isCheckboxGroup = field.type === 'checkbox' &&
         name.endsWith('[]');
+      if (isCheckboxGroup) name = name.slice(0, name.length - 2);
 
       if (name.indexOf('.') !== -1) {
         let deepVal = val;
-        let deepName = name;
 
         if (isCheckboxGroup) {
           deepVal = this._getCheckboxGroupData($formFields.filter(`[name="${name}"]`));
-          deepName = name.slice(0, name.length - 2);
         } else if (field.type === 'checkbox') {
           deepVal = field.checked;
         }
 
         // handle nested model
-        setDeepValue(data, deepName, deepVal);
+        setDeepValue(data, name, deepVal);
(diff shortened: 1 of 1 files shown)


js/views/baseVw.js
```diff
@@ -87,23 +87,22 @@ export default class baseVw extends View {
...
       const isCheckboxGroup = field.type === 'checkbox' &&
         name.endsWith('[]');
+      if (isCheckboxGroup) name = name.slice(0, name.length - 2);
 
       if (name.indexOf('.') !== -1) {
         let deepVal = val;
-        let deepName = name;
 
         if (isCheckboxGroup) {
           deepVal = this._getCheckboxGroupData($formFields.filter(`[name="${name}"]`));
```
jjeffryes: Since name doesn't have brackets on it any more, the line above will need to have them restored so it gets the right inputs.

js/views/baseVw.js
```diff
@@ -87,23 +87,22 @@ export default class baseVw extends View {
...
-          deepName = name.slice(0, name.length - 2);
         } else if (field.type === 'checkbox') {
           deepVal = field.checked;
         }
 
         // handle nested model
-        setDeepValue(data, deepName, deepVal);
+        setDeepValue(data, name, deepVal);
       } else if (isCheckboxGroup) {
         data[name] = this._getCheckboxGroupData($formFields.filter(`[name="${name}"]`));
```
jjeffryes: Same comment as above.

PR #1647, opened 2018-12-17
start the bundled server on testnet

rmisio:


jjeffryes:
@hoffmabc here's the PR for starting on testnet for beta testing builds.

PR #1651, opened 2018-12-18
Update Debian dependency for libgconf2.4

hoffmabc:


build.sh
```diff
@@ -65,7 +65,7 @@ case "$TRAVIS_OS_NAME" in
     npm install -g --save-dev electron-installer-redhat --silent
 
     # Install libgconf2-4
-    sudo apt-get install libgconf2-4
+    sudo apt-get install libgconf2-4 libgconf-2-4
```
jjeffryes: Should this have both `libgconf2-4` and  `libgconf-2-4`?

hoffmabc: One is a shim and the other is real

PR #1662, opened 2019-01-03
Add Chinese translation

fengzie:
To support decentralization market, and contribute together to it.


jjeffryes:
@fengzie I see that you're a member of the Transifex team for the Chinese language translation. 

Once that translation file reaches 80% translated, we will include it in future builds. The best way to assist is to focus on translating more of that file.

fengzie:
@jjeffryes , I've updated it in Transifex, and now it shows 100% translated from 52%. Please check whether it works.

jjeffryes:
Great! I'll check it and pull it in with the other languages before we release the final version of 2.3.

PR #1666, opened 2019-01-05
sanitizes input values

jjeffryes:
This adds sanitation to the getFormData method, to prevent situations where a user pastes malicious code into an input.

### DIFF
--- a/js/views/baseVw.js
+++ b/js/views/baseVw.js
@@ -1,5 +1,6 @@
 import _ from 'underscore';
 import $ from 'jquery';
+import sanitizeHtml from 'sanitize-html';
 import { setDeepValue } from '../utils/object';
 import { View } from 'backbone';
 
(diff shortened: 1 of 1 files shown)


jjeffryes:
I think we can leave this one out of 2.3, so Rob has a chance to look at it. He may have some ideas on a better way to make sure all the inputs are safe.

js/views/baseVw.js
```diff
@@ -61,7 +62,7 @@ export default class baseVw extends View {
       const $field = $(field);
       const varType = $field.data('var-type');
 
-      let val = $field.val();
+      let val = sanitizeHtml($field.val());
```
rmisio: I think it's safer the sanitization happens here:

```
      } else {
        data[name] = val;
      }
```

Doing it right off the bat is concerning to me because there are subsequent checks on the value that may not be compatible with what the sanitization is changing the value to.

PR #1676, opened 2019-01-13
add protocol if missing in template

jjeffryes:
A seller could add a url to a digital order fulfillment without a protocol, like "www.somesite.com". If they do so, the link is clickable but doesn't resolve in the browser, as it gets the file// protocol.

This PR will add https to the front of the URL if it's missing a protocol when displaying it in the order details.


jjeffryes:
This one doesn't need to be merged before we release 2.3.

jjeffryes:
What's happening is the code in .dom has two bugs:
1. it just adds "http://" to the front, so you get "http://file://"
2. the href includes the local directory path when it's turned into a link and becomes something like file:///Users/jjeffryes/openbazaar-desktop-master/openbazaar-desktop/.tmp/www.test.com

The fix is pretty simple though. It might have just been a typo in the original code.

PR #1681, opened 2019-01-18
Improve Mod Loading In Purchase

jjeffryes:
This addresses 2 minor issues:
- the spinner timer was being reset on render.
- if one of the mods never loaded, the button that suggests trying an unverified mod never appeared.

This changes the code to fix the timer reset, and to show the suggestion for an unverified mod when all mods have loaded and non are verified or when the timeout fires.


js/views/components/Moderators.js
```diff
@@ -397,6 +399,7 @@ export default class extends baseVw {
   }
 
   render() {
+    console.log("mods render")
```
rmisio: Can we remove the debugging code?

js/views/components/ModeratorsStatus.js
```diff
@@ -37,17 +37,24 @@ export default class extends BaseVw {
     super.remove();
   }
 
+  setTimeout() {
+    if (!this.spinnerTimeout) {
+      this.spinnerTimeout = setTimeout(() => {
+        let mode = this.getState().mode;
+        if (mode === 'loadingXofY') mode = 'loadingXofYTimedOut';
+        this.setState({ showSpinner: false, mode });
+        clearTimeout(this.spinnerTimeout);
```
rmisio: There's no need clear a timeout at this point. It's already been executed.

styles/modules/_moderatorCard.scss
```diff
@@ -21,9 +21,7 @@
     }
 
     .selectBtn {
-      min-width: 76px;
-      flex-shrink: 0;
-      flex-grow: 1;
+      width: 100%;
```
jjeffryes: This fixes an issue where the buttons weren't the same width on the moderator card.

jjeffryes:
@rmisio this is ready for re-review.

js/views/components/ModeratorsStatus.js
```diff
@@ -28,6 +29,19 @@ export default class extends BaseVw {
     };
   }
 
+  setState(state = {}, options = {}) {
+    const combinedState = { ...this.getState(), ...state };
+    // Any time the state is set to loading, set the spinner timer if needed.
+    if (state.loading && combinedState.showSpinner) {
```
rmisio: Can we please clear the timeout here before creating a new one to ensure multiple timeouts aren't queue up?

js/views/components/moderators/ModeratorCard.js
```diff
@@ -1,13 +1,13 @@
...
-import BaseVw from '../baseVw';
-import loadTemplate from '../../utils/loadTemplate';
-import app from '../../app';
-import Profile from '../../models/profile/Profile';
-import VerifiedMod, { getModeratorOptions } from './VerifiedMod';
-import { handleLinks } from '../../utils/dom';
-import { launchModeratorDetailsModal } from '../../utils/modalManager';
-import { anySupportedByWallet } from '../../data/walletCurrencies';
-import { getLangByCode } from '../../data/languages';
+import BaseVw from '../../baseVw';
```
rmisio: @jjeffryes ah, not sure if you notice my note in the web repo, but it's configured so we could use absolute paths, which would avoid all these path changing shenanigans when you move a file.

rmisio:
@jjeffryes Could you please fix the merge conflicts? Otherwise, it LGTM.

Issue #1684, opened 2019-01-24
Visual bug on open dispute view


placer14:
Brief Description: 
![screenshot_2019-01-23_22_41_22](https://user-images.githubusercontent.com/305332/51653216-8af2cc00-1f60-11e9-8f10-020195d9855a.png)

Operating System (OS and version): MacOS
OpenBazaar version (shown on `About OpenBazaar` page in menu): 2.3.0/0.13.0

Reproducible (Always / Almost Always / Sometimes / Rarely / Couldn't Reproduce): Always

Steps to reproduce:
1. Begin dispute for fulfilled order.
2. Observe visual bug on form page as shown.

PR #1688, opened 2019-01-30
Bulk coin update

jjeffryes:
This should *not* be merged until after the corresponding server PR is merged. 
https://github.com/OpenBazaar/openbazaar-go/pull/1423

This PR fixes two unrelated bugs I saw while working on it, the HR line layout issue #1684 and a typo in a polyglot in wallet/spend.js.


js/views/modals/Settings/Store.js
```diff
@@ -126,9 +141,49 @@ export default class extends baseVw {
...
+    // If the state is set to stop processing, let the timer finish.
+    if (state.hasOwnProperty('isBulkCoinUpdating') &&
+      !state.isBulkCoinUpdating &&
+      this.processingTimer) {
+      delete state.isBulkCoinUpdating;
+    }
+
+    super.setState(state, options);
+  }
+
```
jjeffryes: @rmisio I'm not sure the above it the best way to handle preventing new state changes from stopping the spinner before the user has had a chance to see it, suggestions are welcome here.

js/templates/modals/settings/bulkCoinUpdateBtn.html
```diff
@@ -0,0 +1,18 @@
+<%= ob.processingButton({
+className: `btn clrP clrBr clrT clrSh2 ${ob.isBulkCoinUpdating ? 'processing' : ''} js-applyToCurrent`,
```
rmisio: Could we pleas indent these two lines?

js/utils/bulkCoinUpdate.js
```diff
@@ -0,0 +1,53 @@
...
+
+export function bulkCoinUpdate(coins) {
+  if (!Array.isArray(coins)) {
+    throw new Error('Please provide an array of accepted cryptocurrency codes.');
+  }
+
+  // dedupe the list
+  const newCoins = [...new Set(coins)];
+
+  if (!newCoins.length) {
```
rmisio: These first two shoeError calls IMHO don't make sense as user facing errors. They're really developer errors and this function should just throw them as exceptions, similar to if you're not passing in a proper array.

The user should never be able to get to this point if an update is already in progress, because the button should have a spinner.

If the user doesn't select a coin, then the view should handle it prior to calling this function whether they want to show an inline error (which is what we do in 90% of the other forms) or take the easy route (which I'm fine with) and show a simple message.

js/utils/bulkCoinUpdate.js
```diff
@@ -0,0 +1,53 @@
...
+    showError({ msg: app.polyglot.t('settings.storeTab.bulkListingCoinUpdate.NoCoinsError') });
+  } else if (isBulkCoinUpdating()) {
+    showError({ msg: app.polyglot.t('settings.storeTab.bulkListingCoinUpdate.InProgressError') });
+  } else {
+    events.trigger('bulkCoinUpdateStarted');
```
rmisio: Can this be called `bulkCoinUpdating` to follow suit with how most every other similar util module is doing it?

js/views/modals/Settings/BulkCoinUpdateBtn.js
```diff
@@ -0,0 +1,101 @@
...
+      'click .js-applyToCurrentCancel': 'clickApplyToCurrentCancel',
+      'click .js-applyToCurrentConfirm': 'clickApplyToCurrentConfirm',
+    };
+  }
+
+  startProcessingTimer() {
+    if (!this.processingTimer) {
+      this.processingTimer = setTimeout(() => {
+        this.processingTimer = null;
+        this.setState({ isBulkCoinUpdating: isBulkCoinUpdating() });
```
rmisio: Can this just be:

```
if (!isBulkCoinUpdating()) this.setState({ isBulkCoinUpdating: false });
```

Otherwise the way you're doing it is you're ensuring the call takes at least 500ms, then if it's still going you'll set the `isBulkCoinUpdating` state to true which will result in another timeout of at least 500ms. At this point we don't need to slap another 500ms onto the call... it's already gone at least that amount, so it's best to let it end when it ends.

PR #1697, opened 2019-02-25
Metrics tweaks

jjeffryes:
Some tweaks to the data collected by the metrics.

### DIFF
--- a/js/views/search/Search.js
+++ b/js/views/search/Search.js
@@ -217,7 +217,10 @@ export default class extends baseVw {
   deleteProvider(md = this.sProvider) {
     if (md.get('locked')) {
       openSimpleMessage(app.polyglot.t('search.errors.locked'));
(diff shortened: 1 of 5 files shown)


js/views/search/Search.js
```diff
@@ -61,7 +61,10 @@ export default class extends baseVw {
       recordEvent('Discover_InvalidDefaultProvider', { url: this.providerUrl });
     }
 
-    if (options.query) recordEvent('Discover_SearchFromAddressBar');
+    if (options.query) {
+      recordEvent('Discover_SearchFromAddressBar');
+      recordEvent('Discover_Search', { type: 'addressBar' });
```
jjeffryes: Several of these are duplicated with two events, because I want to keep the old event for future analytics, but also be able to run a query that treats all three as just one event.

It's kind of kludgey, but Countly doesn't have a way to migrate old event data to a new format.

PR #1704, opened 2019-03-01
Improve Quantity Error Check

jjeffryes:
In the setModelQuantity method, an error is thrown when `cur` is not a string, even though `cur` is only used when changing the quantity of cryptocurrency listings. This PR updates it to only throw the error for crypto listings, and to check for empty values. 

This works currently because the API returns a `metadata.coinType` value of "" (which is still a string), but when fetching from IPFS that parameter is missing for non-crypto listings.

### DIFF
--- a/js/views/modals/purchase/Purchase.js
+++ b/js/views/modals/purchase/Purchase.js
@@ -309,7 +309,7 @@ export default class extends BaseModal {
   }
 
   setModelQuantity(quantity, cur = this.cryptoAmountCurrency) {
-    if (this.listing.isCrypto && typeof cur !== 'string') {
+    if (this.listing.isCrypto && (typeof cur !== 'string' || !cur)) {
       throw new Error('Please provide the currency code as a string.');
     }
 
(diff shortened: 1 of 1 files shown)


rmisio:
> This works currently because the API returns a `metadata.coinType` value of "" (which is still a string), but when fetching from IPFS that parameter is missing for non-crypto listings.

Might be a lot of rabbit holes if the data between ipns and ipfs is not the same.

jjeffryes:
> Might be a lot of rabbit holes if the data between ipns and ipfs is not the same. We'd really need to test listing with all different types of data to be sure (coupons, shipping, variants, etc...).

I agree. So far that's the only breaking difference I noticed, but I didn't check every possible thing.

js/views/modals/purchase/Purchase.js
```diff
@@ -309,7 +309,7 @@ export default class extends BaseModal {
   }
 
   setModelQuantity(quantity, cur = this.cryptoAmountCurrency) {
-    if (typeof cur !== 'string') {
+    if (this.listing.isCrypto && (typeof cur !== 'string' || !cur)) {
       throw new Error('Please provide the currency code as a string.');
```
rmisio: Can we please update the error message to say "non-empty string"?

PR #1706, opened 2019-03-01
Show Dispute Started By and IDs for Mod

### LINKED ISSUE #1703: Request: a few more details in dispute resolution screen
client 2.3.0, server 0.13.0
client OSX, server Debian

On the Dispute Summary I think it would be nice to see the UIDs of the buyer and seller.
Currently it just shows the Order Details, Payment Details, and "Dispute Started" which indicates which _user_ opened the dispute, but not what their role is (buyer/seller).

I think it would be helpful if "Dispute Started" indicated if the dispute-opener was vendor or buyer, and also listed the counterparty. These UIDs are listed in the main "Cases" table, but it would make moderation easier if these details were on the summary page.

jjeffryes:
Adds a "started by buyer/seller" to the case view for the moderator, and the peer IDs of the two parties to make things more convenient for the moderator.

Closes #1703

image

### DIFF
--- a/js/languages/en_US.json
+++ b/js/languages/en_US.json
@@ -1732,7 +1732,10 @@
         "reviewLabel": "%{name}'s Review:"
       },
       "disputeStarted": {
+        "buyer": "Buyer",
+        "seller": "Seller",
         "heading": "Dispute Started",
+        "modHeading": "Dispute Started by %{openedBy}",
         "partyIsDisputing": "%{name} is disputing the order:",
         "resolveBtn": "Resolve Dispute",
         "genericIsDisputed": "The order is being disputed:",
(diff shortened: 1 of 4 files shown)


rmisio:
@morebrownies What do you think of this?

We already have `<name> is disputing the order:`. Would it have been better just to put `(buyer)` or `(seller)` after the name rather than changing the heading?

What do you these of these new buyer and seller lines? Do you like that we're just showing peer Id's there?

jjeffryes:
> We already have <name> is disputing the order:

The request in the ticket was de

jjeffryes:
I'm closing this since we never came to consensus on it.

PR #1721, opened 2019-03-14
upgrade nodemon

rmisio:
After upgrading to the latest node version, the app would fail on startup for me with a `[nodemon] Internal watch failed: EMFILE: too many open files, watch`. Upgrading the nodemon version has seemed to resolve it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "openbazaar-desktop",
-  "version": "2.3.0",
+  "version": "2.3.1",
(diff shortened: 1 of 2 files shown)


jjeffryes:
@hoffmabc Travis is failing with a `The command "eval npm ci  " failed.` error, is that just a fluke, or somehow caused by changing the nodemon version?

hoffmabc:
Probably need to update the package lock file.

https://github.com/OpenBazaar/openbazaar-desktop/blob/master/package-lock.json

PR #1727, opened 2019-03-20
Fix case crypto quantity

### LINKED ISSUE #1723: Dispute over cryptocurrency purchase: UI should display `quantity64` instead of `quantity`
Moderating a dispute over a ZEC sale. Note how in the dispute window it displays the quantity purchased as `0`:

Screen Shot 2019-03-16 at 9 43 41 AM

--

However deep in the contract JSON the real purchased value is stored:

Screen Shot 2019-03-16 at 9 44 43 AM

jjeffryes:
It looks like there was a typo where cases for cryptocurrency orders were dividing the wrong quantity field, which always resulted in a quantity of zero.

Closes #1723

### DIFF
--- a/js/models/order/Case.js
+++ b/js/models/order/Case.js
@@ -81,6 +81,10 @@ export default class extends BaseOrder {
     contract.buyerOrder.items.forEach((item, index) => {
       const listing = contract.vendorListings[index];
 
+      // standardize the quantity field
+      item.quantity = item.quantity === 0 ?
+        item.quantity64 : item.quantity;
+
       if (listing.metadata.contractType === 'CRYPTOCURRENCY') {
         const coinDivisibility = listing.metadata
           .coinDivisibility;
--- a/js/models/order/Case.js
+++ b/js/models/order/Case.js
@@ -77,15 +77,19 @@ export default class extends BaseOrder {
     return false;
   }
 
-  convertCryptoQuantity(contract = {}) {
+  convertQuantity(contract = {}) {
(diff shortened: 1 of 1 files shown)


js/models/order/Case.js
```diff
@@ -85,7 +85,7 @@ export default class extends BaseOrder {
         const coinDivisibility = listing.metadata
           .coinDivisibility;
 
-        item.quantity = item.quantity / coinDivisibility;
+        item.quantity = item.quantity64 / coinDivisibility;
```
rmisio: I think this should be fixed in parse, similar to how it's done in the Order model.

```
        // standardize the quantity field
        item.quantity = item.quantity === 0 ?
          item.quantity64 : item.quantity;
```

If I remember correctly, old nodes would return quantity whereas new ones would return the quantity64. Probably shouldn't be many or any nodes that old, but it doesn't hurt to account for both.

jjeffryes: It's in the convertCryptoQuantity method, which is being called by parse. 

I originally had the code above, but since it's only being called when the contractType is cryptocurrency, it felt like it wasn't necessary.

jjeffryes: I see what you're saying now. I was thinking the quantity64 was just for crypto listings, I forgot it's for all newer listings.

jjeffryes:
I noticed the quantity was showing zero when the vendor opened the order for all listing types, so I fixed things to always standardize the quantity for both the buyer and seller contracts.

PR #1740, opened 2019-03-26
Update explorers

### LINKED ISSUE #1718: Replace Blocktrail for Testnet Addresses
The blocktrail explorer is no longer available, we should switch to other providers for testnet addresses in the wallet and order details. 

This should just require changing the data in the coin data file.

### LINKED ISSUE #1732: Switch to privacy-respecting blockchain explorers
We link to multiple blockchain explorers in the client. When possible, we should link to explorers that publicly commit to giving users privacy, such as Blockchair (which we already use for some coins):

https://blockchair.com/privacy

jjeffryes:
This replaces blocktrail with other explorers for testnet, now that block trail is no longer available, and updates BTC, BCH, and LTC to use blockchair.

Closes #1718
Closes #1732

### DIFF
--- a/js/data/walletCurrencies.js
+++ b/js/data/walletCurrencies.js
@@ -20,12 +20,12 @@ const currencies = [
     url: 'https://bitcoin.org/',
     getBlockChainAddressUrl: (address, isTestnet) => (
       isTestnet ?
-        `https://www.blocktrail.com/tBTC/address/${address}` :
+        `https://chain.so/address/BTCTEST/${address}` :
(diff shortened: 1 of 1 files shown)


rmisio:
@jjeffryes @SamPatt Have either of you tested these with segwit addresses to make sure the explorers support them? (not sure which coins segwit applies to)

jjeffryes:
@rmisio I tested all of them to make sure they supported segwit, but it wouldn't hurt to have a second person do that in case I missed one.

PR #1742, opened 2019-03-28
Use Grid for Mixed Results

jjeffryes:
This fixes an issue where if you select cryptocurrency and another listing type in the type filter of the search page all the results will be formatted as cryptocurrencies (a list). 

If more than one type is selected, the results will be shown as a grid instead.

My assumption is this was missed when we changed the listing type filter from a radio group where only one could be selected at a time to a set of checkboxes.


jjeffryes:
This was incorporated into #1742, I'm closing this so we don't have any weird merge issues.

PR #1744, opened 2019-04-03
Update Electron to v4

jjeffryes:
This updates Electron to the latest v4 version available.

The change only affected the package.json, package-lock.json, and the rule for asterisks in the form css.


jjeffryes:
I'm creating a new PR for this based on the latest master so we don't have a bunch of awful merge conflicts.

PR #1746, opened 2019-04-09
Improve discover

### LINKED ISSUE #1728: Only Show Select Provider, Make Default if Multiple Providers
The UI elements for the select provider tooltip and make default link should only be shown if there are more than one providers to choose from. If only one is in the list, hide them.

This should incorporate #1600

### LINKED ISSUE #1730: Make Product Search Look like .com
The product search section should be updated to look closer to the version used in openbazaar.com.

- [ ] change the default view to show the same sections, with see all buttons.
- [ ] don't add a featured section, though a placeholder could be added for when easily updated featured store data is added to the search endpoint.

Ideally we'd eventually have an OBIP to allow any provider to return listings in arbitrary groups instead of us hard coding this layout, so it could be a single call to the API and be updatable by the search provider.

jjeffryes:
This is a pretty extensive rework and refactor of the discover section. The main change was to create a category-based initial experience similar to the one on openbazaar.com.

To see the category view, click on the discover button. If OB1 is your default provider, you will see a set of categories. If a different provider is your default, you're shown a blank search, since we can't know if they support the same search parameters (for instance, a type filter for cryptocurrency listings, which is needed for the Trade category).

The filters and sortBy bar have been broken out into their own views/templates, to make it easier to add them conditionally. If a provider doesn't provide filters, the filter column won't be shown and the results will fill the space instead (the last row will have 2 results instead of 4, but we have to send the page size parameter before we find out the endpoint doesn't return filters).

A simple search object was added that holds all the data needed to make a search call, to make it easier to keep the data passed to different modules consistent. It works similarly to a state object when it's in the main search view, updating it will trigger a search fetch.

Closes #1730
Closes #1728

### DIFF
--- a/js/collections/search/SearchProviders.js
+++ b/js/collections/search/SearchProviders.js
@@ -3,7 +3,6 @@ import is from 'is_js';
 import app from '../../app';
 import Provider from '../../models/search/SearchProvider';
 import LocalStorageSync from '../../utils/lib/backboneLocalStorage';
-import { sanitizeResults } from '../../utils/search';
 import { getCurrentConnection } from '../../utils/serverConnect';
 
 /**
@@ -98,11 +97,5 @@ export default class extends Collection {
       this[idString] = md.id;
       localStorage[storageString] = md.id;
     }
-  }
-
-  fetch(options = {}) {
-    return super.fetch({
-      data: sanitizeResults(options.data || {}),
-    });
   }
 }
--- a/js/collections/search/SearchProviders.js
+++ b/js/collections/search/SearchProviders.js
@@ -59,7 +59,7 @@ export default class extends Collection {
   }
 
   /**
-   * This will retrieve the first provider that has the given URL for one of it's endpoints.
+   * This will retrieve the first provider that has the given URL for one of its endpoints.
    * @param {string} url - The endpoint URL being looked for.
    * @returns {object} - A search provider model.
    */
--- a/js/collections/search/SearchProviders.js
+++ b/js/collections/search/SearchProviders.js
@@ -61,6 +61,8 @@ export default class extends Collection {
    * @returns {object} - A search provider model.
    */
   getProviderByURL(url) {
+    if (!url || is.not.url(url)) throw new Error('Please provide a valid URL.');
+
     return this.models.find(md => {
       let match = false;
       ['listings', 'torListings', 'vendors', 'torVendors'].forEach(type => {
--- a/js/collections/search/SearchProviders.js
+++ b/js/collections/search/SearchProviders.js
@@ -1,8 +1,20 @@
 import { Collection } from 'backbone';
+import is from 'is_js';
 import LocalStorageSync from '../../utils/lib/backboneLocalStorage';
-import { baseUrl } from '../../utils';
 import { sanitizeResults } from '../../utils/search';
 import Provider from '../../models/search/SearchProvider';
+
+/**
+ * Returns the URL minus any query parameters.
+ * @param {string} url - A URL.
+ * @returns {string} - The URL domain and pathname.
+ */
+function baseUrl(url) {
+  if (!url || is.not.url(url)) throw new Error('Please provide a valid URL.');
+
+  const tempUrl = new URL(url);
+  return (`${tempUrl.host}${tempUrl.pathname}`).replace(/\/$/, '');
+}
 
 export default class extends Collection {
   localStorage() {
@@ -47,8 +59,8 @@ export default class extends Collection {
     return this.models.find(md => {
       let match = false;
       ['listings', 'torListings', 'vendors', 'torVendors'].forEach(type => {
-        const typeUrl = md.get(type) && baseUrl(md.get(type), true);
-        if (typeUrl && baseUrl(url, true) === typeUrl) match = true;
+        const typeUrl = md.get(type) && baseUrl(md.get(type));
+        if (typeUrl && baseUrl(url) === typeUrl) match = true;
       });
       return match;
     });
--- a/js/collections/search/SearchProviders.js
+++ b/js/collections/search/SearchProviders.js
@@ -48,13 +48,18 @@ export default class extends Collection {
   }
 
   set defaultProvider(md) {
-    this.setProvider(md);
+    this.setDefaultProvider(md);
   }
 
   set defaultTorProvider(md) {
-    this.setProvider(md, true);
+    this.setDefaultProvider(md, true);
   }
 
+  /**
+   * This will retrieve the first provider that has the given URL for one of it's endpoints.
+   * @param {string} url - The endpoint URL being looked for.
+   * @returns {object} - A search provider model.
+   */
   getProviderByURL(url) {
     return this.models.find(md => {
       let match = false;
@@ -67,7 +72,12 @@ export default class extends Collection {
     });
   }
 
-  setProvider(md, tor = false) {
+  /**
+   * This will set the default provider for clear or Tor mode.
+   * @param {object} md - A provider model.
+   * @param {boolean} tor - Whether to save the default for clear or Tor mode.
+   */
+  setDefaultProvider(md, tor = false) {
     if (!md instanceof Provider) {
       throw new Error('Please provide a model as a Provider instance.');
     }
--- a/js/collections/search/SearchProviders.js
+++ b/js/collections/search/SearchProviders.js
@@ -1,9 +1,8 @@
 import { Collection } from 'backbone';
 import is from 'is_js';
-import app from '../../app';
 import Provider from '../../models/search/SearchProvider';
 import LocalStorageSync from '../../utils/lib/backboneLocalStorage';
-import { getCurrentConnection } from '../../utils/serverConnect';
+import { getCurConnTor } from '../../utils/serverConnect';
 
 /**
  * Returns the URL minus any query parameters.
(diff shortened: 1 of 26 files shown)


js/data/defaultSearchProviders.js
```diff
@@ -4,12 +4,10 @@ const defaultSearchProviders = [
     name: 'OB1',
     logo: '../imgs/ob1searchLogo.png',
     localLogo: '../imgs/ob1searchLogo.png',
-    search: 'https://search.ob1.io/search',
     listings: 'https://search.ob1.io/listings/search',
-    reports: 'https://search.ob1.io/reports',
-    torsearch: 'http://my7nrnmkscxr32zo.onion/search',
     torlistings: 'http://my7nrnmkscxr32zo.onion/listings/search',
-    locked: true,
+    vendors: 'https://search.ob1.io/profiles/search?type=vendor',
```
jjeffryes: The search endpoint uses "vendor" for the vendor search, and doesn't return a "search" endpoint any more, this just updates our code to match.

js/templates/search/sortBy.html
```diff
@@ -0,0 +1,37 @@
+<div class="tx5 flexExpand">
```
jjeffryes: This is also mostly unchanged, just broken out into its own template.

js/utils/dom.js
```diff
@@ -130,3 +130,11 @@ export function handleLinks(el) {
...
   });
 }
+
+/**
+ * This will scroll the pageContainer to the top of the contentFrame. Use this when calling
+ * scrollIntoView on a page view's root element doesn't scroll the pageContainer to zero.
+ */
+export function scrollPageIntoView() {
+  document.getElementById('pageContainer').scrollIntoView();
+}
```
jjeffryes: I added this function because the normal scroll into view was leaving the search page partially scrolled. This happened because the actual scrolling container is the grandparent of the search element, #contentFrame, and you want to scroll #pageContainer, not the search view's element.

js/views/search/Search.js
```diff
@@ -18,113 +19,143 @@ import { selectEmojis } from '../../utils';
...
-      this.mustSelectDefault = true;
-      recordEvent('Discover_InvalidDefaultProvider', { url: this.providerUrl });
-    }
+    const queryKeys = ['q', 'p', 'ps', 'sortBy'];
+
+    this._defaultSearch = {
+      q: '*',
+      p: 0,
+      ps: 66,
+      searchType: 'listings',
```
jjeffryes: The search type is setting things up for later when we add vendor searches.

rmisio:
@jjeffryes Should we remove the bottom hr of the last category?

jjeffryes:
> Should we remove the bottom hr of the last category?

Good catch. I forgot to do that.

rmisio:
One thing that occurs that isn't ideal is if you scroll down on search and then click or add a new provider, it will load the new content, but leave you scrolled at the top. Can we please scroll to the top on a new search provider?

rmisio:
>  If a different provider is your default, you're shown a blank search, since we can't know if they support the same search parameters (for instance, a type filter for cryptocurrency listings, which is needed for the Trade category).

I'm actually seeing that the old Discover experience is shown in that case. Which IMHO is a much much better experience than a blank screen.

rmisio:
Is there a reason why when OB1 is not your default and you go to it, it doesn't show the categoried experience? I would think we should show the category experience, unless you are searching for a term.

rmisio:
not super high priority... but if this could be fixed in an hour or two (and you have a free hour or two), it wouldn't hurt:

image

rmisio:
@jjeffryes @morebrownies @hoffmabc 

image


Since the inventory will always be `unknown` until we revamp inventory which is lower priority thing that probably won't happen for months or years.... should we hide the column for now?

jjeffryes:
> I'm actually seeing that the old Discover experience is shown in that case. Which IMHO is a much much better experience than a blank screen.

Sorry, I was unclear, I meant a blank search with * as the query and no filters.

jjeffryes:
> One thing that occurs that isn't ideal is if you scroll down on search and then click or add a new provider, it will load the new content, but leave you scrolled at the top. Can we please scroll to the top on a new search provider?

Good idea, I'll add that.

> Is there a reason why when OB1 is not your default and you go to it, it doesn't show the categoried experience? I would think we should show the category experience, unless you are searching for a term.

Good point. I guess I was thinking at that point you'd be in search mode and wouldn't want to wait for categories to load, but in practice they aren't much different from the random search.

Re: blue outline, I'll take that out, I'm not sure how it crept back in, unless it's always been that way?

Re: inventory, I think taking it out for now makes sense, since it isn't actually functional.

rmisio:
> If a different provider is your default, you're shown a blank search, since we can't know if they support the same search parameters (for instance, a type filter for cryptocurrency listings, which is needed for the Trade category).

Would it not be better to just omit the crypto box on non-OB1 providers, but give them the other categories?

One the one hand, I think many providers would appreciate having a more curated landing page. On the other, if they are uber specialized, those categories might not make sense.

I could go either way, just want to point out that `filter for cryptocurrency listings` doesn't have to be the sole determiner of categoried experience or not.

js/views/search/Search.js
```diff
@@ -417,177 +412,172 @@ export default class extends baseVw {
...
-    // use the initial set of results data to create the results view
-    if (data) this.createResults(data, state.searchUrl);
+    this.categoryViews.forEach(cat => cat.remove());
+    if (this.filters) this.filters.remove();
+    if (this.sortBy) this.sortBy.remove();
+
+    if (state.showHome) {
+      // Create a disposable copy to be shifted by the addNextCategory function.
+      this._categorySearchesToAdd = [...this._categorySearches];
+      this.addNextCategory();
```
rmisio: Not suggesting some massive refactor at this point... but (I've banged this drums a few times over the years... lol), it's pretty much an anti-pattern to kick off side effects in render. You have to be very careful. For example, with this structured as it is, if you do:

setState({ showHome: true, somethingLabel: 'Hi There' }) // triggers a category fetch
// while the category is fetching -or- some subsequent one that addNextCategory kicked off, you need to update some state, eg.
setState({ showHome: true, somethingLabel: 'Bye Bye!' }) // triggers a another add category fetch and now you have to addCategoryNext cycles going

You've basically made it that any state change or render will result in a fetch regardless of if that state change or render has anything to do with the fetch.

styles/components/_listingsGrid.scss
```diff
@@ -2,12 +2,12 @@
    the store and search results views. Make sure to check both when making changes to this file.
  */
 
-$crytoListViewTradeFromWidth: 57px;
-$crytoListViewTradeToWidth: 104px;
-$crytoListViewTradeArrowWidth: 48px;
-$crytoListViewVendorWidth: 185px;
-$crytoListViewPriceWidth: 125px;
-$crytoListViewInventoryMaxWidth: 138px;
+$crytoListViewTradeFromWidth: 9%;
```
rmisio: hmmm... a lot of tinkering was put in place to make sure different permutations of the variable length content would all look nice. If you're implementation is bigger, we're ok. If it's smaller, maybe not.

js/collections/search/SearchProviders.js
```diff
@@ -34,14 +48,36 @@ export default class extends Collection {
...
+    this.setDefaultProvider(md);
   }
 
   set defaultTorProvider(md) {
-    this.setProvider(md, true);
+    this.setDefaultProvider(md, true);
+  }
+
+  /**
+   * This will retrieve the first provider that has the given URL for one of it's endpoints.
```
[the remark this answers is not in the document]
jjeffryes: The airbnb linter we use sets the max to 100, which is what I have it set to.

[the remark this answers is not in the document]
rmisio: ok, that works. as long as we're enforcing something.

js/collections/search/SearchProviders.js
```diff
@@ -58,4 +94,10 @@ export default class extends Collection {
       localStorage[storageString] = md.id;
     }
   }
+
+  fetch(options = {}) {
```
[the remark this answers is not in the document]
jjeffryes: I think this was a copy paste error, it was supposed to go in the results collection.

jjeffryes:
@rmisio this is ready for re-review now.

rmisio:
Something is still messed up with routing:

(make sure to have at least two providers)

- click Discover when OB1 is your default
- click See All on the Crypto Trading Box
- click on a second provider
- click Back // you land on the See All result from step 2, which is correct
- click Back // you land on the See All result from step 2, which is *not* correct, you should land on OB1 home
- click Back // still stuck on the See All result from step 2

js/router.js
```diff
@@ -43,6 +43,7 @@ export default class ObRouter extends Router {
       ['(ob://)transactions/:tab(/)', 'transactions'],
       ['(ob://)connected-peers(/)', 'connectedPeers'],
       ['(ob://)search(?query)', 'search'],
+      ['(ob://)search/home', 'searchHome'],
```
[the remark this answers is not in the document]
jjeffryes: That's a great point. I was stuck thinking it needed to be a tab when I was wiring in tab support for when we add a vendor path later.

Plus that opens the door for having a separate "home" experience for the vendor search path if we wanted one (ie: tab is vendor but there is no query).

jjeffryes:
@rmisio this is ready again.

PR #1781, opened 2019-06-25
Update Dependencies for 2.3.5

### LINKED ISSUE #1654: Evaluate and Update all Dependencies
Some of our dependencies are old, we should evaluate each one for:
1. is it still used?
2. is it obscure or abandoned, which may make it a future security concern? If so, can we replace it with a better supported library or our own code?
    - This may require a separate issue if it's a significant change. The CropIt library, for instance, hasn't been updated in a while, but would be complicated to replace.
3. is the version up to date? If not, update it to the latest version, and make any changes needed for compatibility. 

This should be done after or in conjunction with #1366.

### LINKED ISSUE #1771: Check Dependency Compatibility with the Latest Node Version
Node-Sass requires v4.12 (updated in #1769) to be compatible with Node JS version 12. 

Since Node 12.2 is available, we should check our dependencies to see if any more of them need to be upgraded for compatibility.

### LINKED ISSUE #1774: Update Old Dependencies
Github has some warnings about old dependencies, they should be updated. 

This is similar to #1771 and should be done together with it.

jjeffryes:
This updates a variety of dependencies for release 2.3.5 or later. This is primarily a precaution, so we don't end up too far behind on our libraries.

The major changes here are Babel and Electron.

Babel's syntax is different, I used https://github.com/babel/babel-upgrade to auto-generate the new dependencies. 

Electron has been updated to v5.0.5, it now defaults to `nodeIntegration: false`, which required it to be set to true when the browser window is created.

Note: after updating Node to 10.16.0, I had to delete my node_modules file and reinstall everything with `npm install` to get around a too many files open issue when browser sync started watching the sass files.

Also, one of the currency tests started to fail, and required a small tweak to get it working right. The PLN currency test has a space between the PLN and the amount now.

Closes #1654
Closes #1771
Closes #1774


test/utils/currency.js
```diff
@@ -108,7 +108,7 @@ describe('the currency utility module', () => {
         'the resulting value', () => {
         expect(cur.convertAndFormatCurrency(500, 'USD', 'PLN', { locale: 'en-US' }))
           .to
-          .equal('PLN2,097.31');
+          .satisfy(val => ['PLN2,097.31', 'PLN 2,097.31'].includes(val));
```
jjeffryes: This is a little ugly, but on my desktop it adds the space, and on Travis it doesn't, so this is the only way I've been able to get it to pass tests on both.

.babelrc
```diff
@@ -1,13 +1,105 @@
 {
   "env": {
     "development": {
-       "presets": ["env", "stage-1"]
-     },
+      "presets": [
```
rmisio: At some point I think we could pare these down because I know we don't use a lot of these things (e.g. decorators). Just don't think we want to needlessly increase the file size of the bundled app or the amount of deps that have to come down the wire.

PR #1793, opened 2019-08-05
Fix Test Failing Because of Space

jjeffryes:
The international formatting is adding a non-breaking space, which breaks a test. This will fix it by checking against the non-breaking space character code.

See https://stackoverflow.com/questions/54242039/intl-numberformat-space-character-does-not-match


jjeffryes:
This was the wrong way to fix this, the correct way is in #1781

PR #1808, opened 2019-08-13
Css updates for235

jjeffryes:
This fixes three visual bugs:

1. the vertical spacing on the load moderators section of settings.store was way too big. A div inside a flex container was set to 100% height, which picked up the height of the entire container. That was an error, my guess is it worked before due to a bug in Chromium that was fixed in an Electron update.

2. the placeholder box with the load mods button was getting a height that pushed it outside the bottom of its container into the next div, probably due to the same change in Chromium. It was tweaked and no longer overlaps the text below it.

3. when the app was loading, the chat container would briefly have a black border. This was happening because the container border color was set to transparent, with a transition, so it could become grey when it opened. 

Again, probably due to a Chromium update, the browser behavior changed and the transition was being applied to the default border color of black. I'm not sure why it would do that, it might be a Chromium bug. Removing the transition (which is barely noticeable when it changes from transparent to grey) was the simplest solution.

### DIFF
--- a/js/templates/modals/settings/store.html
+++ b/js/templates/modals/settings/store.html
@@ -88,16 +88,18 @@
                 <div class="rowLg"><% // just a spacer %></div>
               </div>
               <div class="row">
-                <div class="flexVCentClearMarg tx5 rowTn">
-                  <h5 class="flexExpand"><%= ob.polyT('settings.storeTab.availableModerators') %></h5>
-                  <input type="checkbox" id="storeVerifiedOnly" class="js-storeVerifiedOnly"<% if (ob.showVerifiedOnly) print('checked') %>>
-                  <label class="tx5b" for="storeVerifiedOnly"><%= ob.polyT('settings.storeTab.verifiedOnly') %></label>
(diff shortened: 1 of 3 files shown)


rmisio:
> Removing the transition (which is barely noticeable when it changes from transparent to grey) was the simplest solution.

@jjeffryes I take it if we hard-code the default border-color to transparent, it doesn't fix the issue?

jjeffryes:
> @jjeffryes I take it if we hard-code the default border-color to transparent, it doesn't fix the issue?

Do you mean hard code it for all divs? I didn't try that. It's already being set to transparent in `#chatContainer` which should override any default values, which is what made this perplexing.

jjeffryes:
@rmisio I can't figure this one out. No matter what styles I put on #chatContainer, it's briefly shown with a black border and a white background when the app is loading. 

If I use an inline style tag in index.html, it's applied, but of course then it's stuck with that color.

It's as if the layout CSS rules are being applied, but not the color rules, for a brief second.

PR #1809, opened 2019-08-13
1807

### LINKED ISSUE #1807: Resync Bug
When the resync button is clicked, an error is being thrown on _server.save()

```
_server.save(...).done is not a function
    at Object.<anonymous> (C:\Users\User\AppData\Local\OpenBazaar2\app-2.3.5\resources\app.asar\prod\utils\resyncBlockchain.js:185:10)
    at fire (C:\Users\User\AppData\Local\OpenBazaar2\app-2.3.5\resources\app.asar\node_modules\jquery\dist\jquery.js:3291:31)
    at Object.fireWith [as resolveWith] (C:\Users\User\AppData\Local\OpenBazaar2\app-2.3.5\resources\app.asar\node_modules\jquery\dist\jquery.js:3421:7)
    at done (C:\Users\User\AppData\Local\OpenBazaar2\app-2.3.5\resources\app.asar\node_modules\jquery\dist\jquery.js:9523:14)
    at XMLHttpRequest.<anonymous> (C:\Users\User\AppData\Local\OpenBazaar2\app-2.3.5\resources\app.asar\node_modules\jquery\dist\jquery.js:9775:9)
```

rmisio:
closes #1807

### DIFF
--- a/js/utils/resyncBlockchain.js
+++ b/js/utils/resyncBlockchain.js
@@ -16,7 +16,7 @@ let server;
 
 // If you change this, be sure to change anywhere in the GUI you may have output how
 // long its unavailable.
-const resyncInactiveTime = 1000 * 60 * 60 * 1; // 1 hour
+const resyncInactiveTime = 1000 * 60;
 
 function checkCoinType(coinType) {
   if (typeof coinType !== 'string' && !coinType) {
@@ -45,6 +45,8 @@ function __isResyncAvailable(coinType) {
 let _isResyncAvailable = {};
 
 function setResyncAvailable(coinType, bool = __isResyncAvailable(coinType)) {
+  if (resyncInactiveTime <= 0) return;
+
   checkCoinType(coinType);
 
   if (typeof bool !== 'boolean') {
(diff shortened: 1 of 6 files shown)


js/utils/resyncBlockchain.js
```diff
@@ -170,14 +172,21 @@ export default function resyncBlockchain(coinType) {
           lastBlockchainResync = typeof lastBlockchainResync === 'object' ?
             lastBlockchainResync : {};
           lastBlockchainResync[coinType] = (new Date()).getTime();
-          _server.save({ lastBlockchainResync })
-            .done(() => {
+          const serverSave = _server.save({ lastBlockchainResync });
```
rmisio: So the problem was that `isjs` was updated and it looks like it not longer considers `localhost` as a valid IP, so I had to manually add that condition to the mode validation.

What happens is `Model.save()` returns `false` if validation fails as opposed to the jqXhr promise if it passes validation. So handling the case here. It really shouldn't ever fail validation unless it's something weird like this and if it does, it will have better console output than the seemingly random exception that was in place before.

js/utils/resyncBlockchain.js
```diff
@@ -16,7 +16,7 @@ let server;
 
 // If you change this, be sure to change anywhere in the GUI you may have output how
 // long its unavailable.
-const resyncInactiveTime = 1000 * 60 * 60 * 1; // 1 hour
+const resyncInactiveTime = 1000 * 60;
```
rmisio: Since the server POST comes back immediately and independent of the actual blockchain resync process, I decided to leave a 1 minute inactive time in to at least limit the chance of concurrent server side calls being fired off.

jjeffryes:
@rmisio looks good to me, tests out fine. Are you ready for it to merge?

rmisio:
@jjeffryes yes, it's ready on my end.

PR #1811, opened 2019-09-04
Fix Mismatch with Notification Types

jjeffryes:
I'm not sure how we missed this, but the notification types we were expecting don't match the ones the server sends for orderDeclined and cancel.

This resulted in notifications for a declined offline order or a cancelled offline order not showing any text in the notifications list, and not being clickable.

 https://github.com/OpenBazaar/openbazaar-go/blob/6eca0f933482c206aee7b124137baa76532faad3/repo/constants.go

### DIFF
--- a/js/collections/Notifications.js
+++ b/js/collections/Notifications.js
@@ -77,7 +77,8 @@ export function getNotifDisplayData(attrs, options = {}) {
     text = app.polyglot.t('notifications.text.orderConfirmation', {
       vendorName,
     });
-  } else if (attrs.type === 'orderDeclined') {
+    // Check for the new 'orderDeclined' value, but maintain compatibility with older notifications.
+  } else if (attrs.type === 'orderDeclined' || attrs.type === 'declined') {
     const vendorName = opts.native ?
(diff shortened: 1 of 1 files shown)


jjeffryes:
@rmisio do you have any ideas on how we missed this? I'm certain I've tested these notifications over the past year, and neither the server nor the client code has changed since 2018.

@placer14 or @cpacia did something else change on the server that would have switched the type in the order declined from `declined` to `orderDeclined` and cancel to `cancel` from `cancelled`? Maybe something was overriding the constants before?

placer14:
We did update those notifications to use constants instead of strings but it's possible they were updated wrong? I'll have to check it out. 😓

placer14:
Yeah, we changed the strings when we started using the constants: https://github.com/OpenBazaar/openbazaar-go/commit/1f0a7cd8ededbe3c0530edb21261c02873f92a9e

We probably need to include this fix in our next release.

jjeffryes:
I've updated the code so it keeps compatibility with the old notification types, I had a node that had both versions so I could test against that. We shouldn't need to migrate the server or anything like that.

PR #1837, opened 2019-11-20
Bump node-sass to 4.13.0 cause binding node asset is unavailable

M0Rf30:
During the latest building process of openbazaar 2.3.6 a message about the
node-sass/4.12.0/linux-x64-79_binding.node
pop out about the non existing resource
This commit fix the issue

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7928,9 +7928,9 @@
       }
     },
     "node-sass": {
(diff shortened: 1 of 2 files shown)


jjeffryes:
Thank you for submitting this, @M0Rf30

PR #1838, opened 2019-11-26
Dispute payout precision

### LINKED ISSUE #1841: Eth-Precision: Moderator Fee in Settings Shown as USD when it is ETH
The server is returning amountCurrency:{ code: "ETH", divisibility: 8}, but in settings/moderation the moderator's fixed fee is showing USD as the currency.

This requires the user to set their currency back to ETH before saving.

rmisio:
Implements conversions from string based integers to numbers for the dispute resolution section of  order data.

closes #1841

### DIFF
--- a/js/models/listing/Item.js
+++ b/js/models/listing/Item.js
@@ -166,9 +166,6 @@ export default class extends BaseModel {
       'bigPrice'
     );
 
-    console.log('the faddie is *billy*');
-    window.billy = attrs.bigPrice;
-
     if (!attrs.images.length) {
       addError('images', app.polyglot.t('itemModelErrors.imageRequired'));
     } else if (attrs.images.length > max.images) {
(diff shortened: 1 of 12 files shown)


jjeffryes:
With the latest changes, I get a "The moderation settings were not saved." error from the client when I try to save moderation settings with a percentage fee.

jjeffryes:
In the save.fail on settings/moderation.js, I'm seeing 
```
message: "Unexpected end of JSON input"
stack: "SyntaxError: Unexpected end of JSON input↵    at parse (<anonymous>)↵    at ajaxConvert (/Users/jjeffryes/openbazaar-desktop-mast
```

jjeffryes:
@rmisio is this ready for re-review?

rmisio:
@jjeffryes Probably not because there are still ob-go bugs that need to be resolved. I don't think they'll require any client action, but it's still a potentially broken experience until they are fixed.

rmisio:
@jjeffryes I think you could re-test this one since it looks like the dependent ob-go bugs have presumedly had fixes merged.

jjeffryes:
This looks good on re-testing.

PR #1845, opened 2019-12-09
Tweaks

rmisio:
Various tweaks I'v been fixing as I test ETH (any future fixes will go into a subsequent PR):

- Limiting the max decimals for crypto prices in the transactions table to 6 (this will extend if 6 or more would result in a 0 price)
- Fixed an issue where the listing detail modal was bombing on some v4 (old) listing data.
- Limiting the max decimals for crypto prices on the listing card to 6 (this will extend if 6 or more would result in a 0 price)
- On the listing card, bumping down the font size of the price line if the rating and price hav a lot of characters (this would most notable happen if you have an incredibly low price, e.g. 1 WEI).

### DIFF
--- a/js/templates/components/cryptoPrice.html
+++ b/js/templates/components/cryptoPrice.html
@@ -16,7 +16,8 @@
       ob.currencyMod.convertAndFormatCurrency(
         ob.priceAmount,
         ob.priceCurrencyCode,
-        ob.displayCurrency
+        ob.displayCurrency,
+        ob.convertAndFormatOpts
       )
     %>&nbsp;<span class="priceSymbol"><%= priceSymbol %></span>
   </span>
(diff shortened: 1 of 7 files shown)


jjeffryes:
Most of the listings in discover don't show a price, and throw an error in the console for currency.js, is that expected behavior since the listing format converter for v4 to v5 on the server isn't ready yet? I think it is, but want to make sure.

rmisio:
> Most of the listings in discover don't show a price, and throw an error in the console for currency.js, is that expected behavior since the listing format converter for v4 to v5 on the server isn't ready yet? I think it is, but want to make sure.

@jjeffryes Yes, that is the graceful way of handling the situation. It's unable to determine the price because the divisibility is not provided, so it just leaves the price blank. The exception you're seeing in the console is just logged so we know what's going on (i.e. it's expcted).

jjeffryes:
Cool. Should this wait to be merged until after the other PR?

rmisio:
> Cool. Should this wait to be merged until after the other PR?

I don't think so. I don't think they touch the same code and if they do, I'll just deal with the merge conflicts.

PR #1848, opened 2019-12-16
More tweaks

### LINKED ISSUE #1825: Eth-Precision Price Rounding in Fiat is Inconsistent
When a listing is priced at less than 0.01 of fiat currency, it's displayed with unexpected rounding in the purchase flow.

For example, if the price is displayed as $0.003, and the quantity is 2, the total is shown as $0.01.

I'd expect the use of 3 digits to be consistent, and the price displayed as $0.006.

I saw something similar with $0.00003 x3 displayed as $0.0001 instead of $0.00009.

Below are the listings I saw that behavior on, if you want to check them:

ob://QmcF2eG2tpRCZddxi8oCnpk5E8pHyRGp3nFbFzPMBGpyGt/store/test-inventory-surcharge-for-eth-precision

ob://QmcF2eG2tpRCZddxi8oCnpk5E8pHyRGp3nFbFzPMBGpyGt/store/test-phys-eth-precision

image

### LINKED ISSUE #1852: ETH - Payment Amount Becomes NaN after Paying
After paying for a listing priced in ETH, the payment amount changes to NaN. It was normal before the button in the pay tooltip was clicked.

This happened when paying in LTC. It doesn't happen when paying in ETH.

This is using #1838 and eth-master for the server.

image

rmisio:
- fixed bug where divisibility was not being properly validated in a listing due to the decimal places utility function not properly handling scientific notations
- putting in a translations for a server purchase error code
- fixed issue where the coupon discount type was not showing correctly in the edit listing modal
- adjusting the max number of decimal places in the bigNumber config

This was because a memory leak was observed on certain divisions (depending on the numbers being divided). 

- added messaging to the purchase receipt total

This was discussed with Josh and Wolf. The messaging will reflect that the receipt total is an estimate.

- updated the Payment notification handler in Purchase to handle an update to the schema in the data being sent over.
- tweaks to the wallet stats (balance, counts...) area to better handle space limitations (e.g. an ETH balance with 18 digit precision)
- Updated the mod component to pass in an asyncID rather than relying on the server to provide one. This will eliminate a potential race condition if the socket returns before the HTTP api call. (I thought the race condition was happening, but it turned out I was observing #1851.)

closes #1852 
closes #1825


js/views/modals/purchase/Purchase.js
```diff
@@ -278,6 +281,19 @@ export default class extends BaseModal {
...
+    const superReturn = super.setState(state, options);
+
+    if (
+      this.receipt &&
+      this.getState().phase !== 'pay'
+    ) {
+      this.receipt.showTotalTip = false;
+    }
+
+    return superReturn;
```
jjeffryes: Would it be simpler to just return super.setState(state, options) here instead of setting it as superReturn first?

jjeffryes: That's super minor, so up to you if you want to change it. Either way we can merge this.

rmisio: The reason it's done this way is so the `this.getState().phase !== 'pay'` check happens after the super call changes state. So we're basically checking if the resulting phase is not 'pay', we hide that tooltip.

I suppose I could have just checked if the passed in state.phase !== 'pay', but it feels like maybe there are or could be some edge cases if the state were somehow changed a different way, e.g. maybe someone sets the initial state to not be 'pay'.

jjeffryes: Ah, ok, that makes sense. Thanks!

PR #1856, opened 2020-01-09
Only Count Unique Incoming TXs

jjeffryes:
In the wallet, we are double-counting txs that come in through the wallet socket message, because it sends one message for UNCONFIRMED and a second one for PENDING, which will show a "X new transactions Refresh" popup with a total of 2x the number of new transactions.

This tweak will only count each tx once instead.

### DIFF
--- a/js/views/modals/wallet/transactions/Transactions.js
+++ b/js/views/modals/wallet/transactions/Transactions.js
@@ -40,7 +40,7 @@ export default class extends BaseVw {
     this.transactionViews = [];
     this.fetchFailed = false;
     this.fetchErrorMessage = '';
-    this.newTransactionCount = 0;
+    this.newTransactionsTXs = new Set();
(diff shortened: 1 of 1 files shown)


jjeffryes:
@rmisio I know you're on vacation, I added you as a reviewer just to make sure you were aware of this when you get back.

rmisio:
> we are double-counting txs that come in through the wallet socket message, because it sends one message for UNCONFIRMED and a second one for PENDING, which will show a "X new transactions Refresh" popup with a total of 2x the number of new transactions.

Are you sure?

In the block below it checks 
``
          if (transaction) {
            // existing transaction has been confirmed
            transaction.set(transaction.parse(_.omit(e.jsonData.wallet, 'wallet')));
          } else {
            // new transaction
            this.newTransactionsTXs.add(e.jsonData.wallet.txid);

PR #1857, opened 2020-01-09
Handle Mod With Missing Fixed Fee Currency

jjeffryes:
This prevents an unhandled error when a moderator has '' for the currency code in their fixedFee object by validating the model when it's added to the collection. This was causing loading moderators in settings/store to fail.

Now all moderators will be validated before being added to the moderator view, and if they are invalid, they will not be shown.

### DIFF
--- a/js/views/components/moderators/Moderators.js
+++ b/js/views/components/moderators/Moderators.js
@@ -153,7 +153,7 @@ export default class extends baseVw {
     const validCur = anySupportedByWallet(modCurs);
 
     if ((!!validMod && validCur || this.options.showInvalid)) {
-      this.moderatorsCol.add(new Moderator(data, { parse: true }));
+      this.moderatorsCol.add(new Moderator(data, { parse: true, validate: true }));
       this.removeNotFetched(data.peerID);
     } else {
       // remove the invalid moderator from the notFetched list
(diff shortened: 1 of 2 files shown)


jjeffryes:
@rmisio I know you're on vacation, the review here is just so you take a look after you're back. Since this fixes a bug that could break loading moderators, I merged it for the 2.3.8 release.

js/views/components/moderators/Moderators.js
```diff
@@ -153,7 +153,8 @@ export default class extends baseVw {
     const validCur = anySupportedByWallet(modCurs);
 
     if ((!!validMod && validCur || this.options.showInvalid)) {
-      this.moderatorsCol.add(new Moderator(data, { parse: true }));
```
rmisio: I would rename `validCur` to something less ambiguous, maybe `walletSupportsModAcceptedCurs`. I know it's a mouthful, but with the very generic `validCur` it took me a few minutes to understand how that was different than the fixed fee cur being invalid.

`this.options.showInvalid`:

So, the user has an option to show invalid mod cards, but in this case they won't be shown even if the user passes in that flag as true. That's a confusing API in my opinion. If it's too much of a rabbit hole to show them when their model is invalid, at least I would update the comment in that flag to make it clear it won't affect data which the model validates as invalid.

jjeffryes: I totally agree the naming is confusing, I can (and should) clean that up.

The showInvalid option shows the mods if they have turned off moderation, which is necessary in your settings if a mod you've already added isn't a mod any more, and you want to remove them. There wasn't anything that handled a mod with actual invalid data (despite the poor naming), and invalid data just broke the render.

This protects against that situation, but then the user is still left with a moderator on their list that isn't valid. A later PR will be needed to handle that, so you can see moderators that have bad data and remove them.

PR #1858, opened 2020-01-13
Add Processing Error to Order Progress Bar

jjeffryes:
We've had a report of an order where the progress bar in the order details was blank, and we confirmed the order state was PROCESSING_ERROR.

I didn't see anything in the summary.js code that would handle that situation.

I'm not sure if this is a complete fix, we'd also want to show the error, or some message to the user about the error, so they can do something to fix it. I don't have an order in an error state though, so I don't know what data about the error we'd have access to. 

Screen cap from the affected user:

image

### DIFF
--- a/js/languages/en_US.json
+++ b/js/languages/en_US.json
@@ -1785,7 +1785,8 @@
           "refunded": "Refunded",
           "decided": "Decided",
           "resolved": "Resolved",
-          "disputed": "Disputed"
+          "disputed": "Disputed",
+          "errored": "Error"
         },
         "heading": "Order Details",
         "shipToHeading": "Ship to",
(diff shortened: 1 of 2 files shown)


rmisio:
I would close this PR because I think there needs to be more of an investigation into what exactly the user's issue was, because I don't think their order has a processing error.

When an order has a processing error, the expectation is the contract has an error field with a list. The list can be empty or have a list of error strings. In any case as long as it's an Array the "processing error" UI will show.

What that consists of is a generic red message under the progress bar explaining that the vendor was unable to process the order. Underneath the generic message will be any string errors the server provided in the contract. The generic error will vary depending on different circumstances (e.g. whether it's a moderated order, what your role is in the order (vendor, moderator, buyer)).

The progress bar is not left blank. The UI tries to determine the state of the order based on other information in the order data and the progress will continue based on subsequent user actions (e.g. canceling a non-moderated offline order, opening and progressing a dispute, etc...).

rmisio:
@jjeffryes Closing this. If you feel it's needed, please feel free to log an issue to investigate the user's issue.

PR #1859, opened 2020-01-14
Refactor mod loading

### LINKED ISSUE #1851: Messaging needed when no valid mods are available
Steps:
1.) Create a listing.
2.) Have vendor add a sole mod who has moderation services turned off.
3.) From a 3rd node (buyer), go to the purchase screen for the listing.

image

Observed:

The mod box is blank and empty. It looks kind of broken.

Expected:

Maybe some type of messaging that none of the mods are valid.

jjeffryes:
This cleans up and improves the handling of loading mods. 

It also fixes an issue where sometimes the empty placeholder was being shown instead of one of the messages that explained no moderators were being shown. Now the loading message will be shown while loading, and one of the other messages will be shown if no valid moderators were loaded.

Closes #1851

### DIFF
--- a/js/templates/components/moderators/moderators.html
+++ b/js/templates/components/moderators/moderators.html
@@ -1,12 +1,5 @@
 <div class="moderatorsWrapper clrBr <% print(ob.wrapperClasses) %> js-moderatorsWrapper">
-  <% // placeholder so border collapse doesn't erase the table border when there are no table cells %>
-  <% if (ob.placeholder) { %>
-    <div class="moderatorCard clrBr">
-      <div class="moderatorCardInner">
-      </div>
-    </div>
-  <% } %>
-  <% if (ob.totalIDs && !ob.totalShown) { %>
+  <% if (!ob.totalShown) { %>
     <div class="moderatorCard moderatorsMessage clrBr">
       <div class="moderatorCardInner">
         <div class="flexCent">
@@ -15,10 +8,12 @@
                const showOnlyCur = ob.showOnlyCur ? 'Matching' : '';
                const opts = ob.showOnlyCur ? { coin: ob.showOnlyCur } : {};
                const msgPath = `moderators.noModsMsg.no${showOnlyCur}${showOnlyVer}Moderators`;
+               const title = ob.loading ?
+                 ob.polyT('moderators.moderatorsLoading') : ob.polyT(`${msgPath}.title`, opts);
             %>
-            <h4><%= ob.polyT(`${msgPath}.title`, opts) %></h4>
+            <h4><%= title %></h4>
             <% // The section below is only relevant if the moderators are loaded in a purchasing context. %>
-            <% if (ob.purchase) { %>
+            <% if (ob.purchase && !ob.loading) { %>
               <div class="tx4 clrT2"><%= ob.polyT(`${msgPath}.body`) %></div>
               <% if (showOnlyVer) { %>
                 <div class="padTn"><% // just a spacer %></div>
--- a/js/views/components/moderators/Moderators.js
+++ b/js/views/components/moderators/Moderators.js
@@ -234,8 +234,6 @@ export default class extends baseVw {
               data.forEach(mod => {
                 if (!excluded.includes(mod.peerId)) this.processMod(mod.profile);
               });
-              this.unfetchedMods = [];
-              this.checkNotFetched();
               if (!data.length) this.trigger('noModsFound', { guids: this.modsToFetch });
             }
           })
--- a/js/views/components/moderators/Moderators.js
+++ b/js/views/components/moderators/Moderators.js
@@ -48,7 +48,6 @@ export default class extends baseVw {
       useCache: true,
       moderatorIDs: [],
       excludeIDs: [],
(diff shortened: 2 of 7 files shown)


js/views/modals/purchase/Purchase.js
```diff
@@ -162,7 +162,7 @@ export default class extends BaseModal {
       singleSelect: true,
       radioStyle: true,
       initialState: {
-        showOnlyCur: currencies[0],
+        showOnlyCur: '',
```
jjeffryes: This was only showing moderators that accept BTC, I assume it was there because originally we automatically selected the first currency. Since we changed it to require the user to select a currency (none are selected by default) this should also have no default.

js/views/components/moderators/Status.js
```diff
@@ -38,7 +38,7 @@ export default class extends BaseVw {
         let mode = this.getState().mode;
         if (mode === 'loadingXofY') mode = 'loadingXofYTimedOut';
         this.setState({ showSpinner: false, mode });
-      }, 10000);
+      }, 60000);
```
jjeffryes: I made this longer since I was seeing the message for no mods appear then mods would load, which was pretty clunky.

js/views/components/moderators/Moderators.js
```diff
@@ -172,37 +175,36 @@ export default class extends baseVw {
...
-    });
+    // don't get any that have already been added or excluded.
+    const excluded = [...this.allIDs, ...this.excludeIDs];
+    this.modsToFetch = _.without(op.moderatorIDs, excluded);
+    this.unfetchedMods = [...this.modsToFetch];
+    this.fetchingVerifiedMods = app.verifiedMods.matched(this.modsToFetch);
 
     // Either a list of IDs can be posted, or any available moderators can be retrieved with GET
-    if (IDs.length || op.method === 'GET') {
+    if (this.modsToFetch.length || op.method === 'GET') {
```
jjeffryes: This was moved so it wouldn't set the state to loading if there were no mods to fetch.

jjeffryes:
Closed while I fix a bug.

js/templates/components/moderators/moderators.html
```diff
@@ -1,12 +1,5 @@
 <div class="moderatorsWrapper clrBr <% print(ob.wrapperClasses) %> js-moderatorsWrapper">
-  <% // placeholder so border collapse doesn't erase the table border when there are no table cells %>
-  <% if (ob.placeholder) { %>
-    <div class="moderatorCard clrBr">
-      <div class="moderatorCardInner">
-      </div>
-    </div>
-  <% } %>
-  <% if (ob.totalIDs && !ob.totalShown) { %>
```
jjeffryes: The placeholder got refactored so the messages below take care of everything with less duplication.

js/views/components/moderators/Moderators.js
```diff
@@ -93,9 +94,9 @@ export default class extends baseVw {
 
     super(opts);
     this.options = opts;
-    this.excludeIDs = opts.excludeIDs;
+    this.excludeIDs = [...opts.excludeIDs, app.profile.id];
```
jjeffryes: It seemed a lot smarter just include the user's own profile here, instead of adding it later.

We could add a parameter to not exclude it in the future if that use case ever came up.

js/views/modals/Settings/Store.js
```diff
@@ -286,22 +286,11 @@ export default class extends baseVw {
           const remAvail = this.modsAvailable.removeModeratorsByID(this.modsAvailable.selectedIDs);
 
           this.modsByID.excludeIDs = this.currentMods;
-          this.modsByID.moderatorsStatus.setState({
-            hidden: true,
-          });
```
jjeffryes: These three sections of code were removed, they were just duplicating things already handled by the view and creating some weird edge case behavior (like if you loaded moderators, then closed and reopened settings the status bar would appear).

js/views/components/moderators/Moderators.js
```diff
@@ -403,19 +403,16 @@ export default class extends baseVw {
   render() {
     const state = this.getState();
     const showMods = this.modCards.filter(mod => this.modShouldRender(mod.model));
-    const unVerCount = this.modCards.filter(mod =>
-      mod.model.hasModCurrency(state.showOnlyCur) && !mod.model.isVerified).length;
-    const totalIDs = this.allIDs.length;
+    const unVerCount = this.modCards.filter(mod => (!state.showOnlyCur ||
+      mod.model.hasModCurrency(state.showOnlyCur)) && !mod.model.isVerified).length;
```
jjeffryes: Since a default currency isn't being passed in (removed below), this was updated to be count all the unverified moderators if there is no currency set.

rmisio:
This one happens on this branch:

https://github.com/OpenBazaar/openbazaar-desktop/issues/1862

Not sure if it was introduced by this branch since I haven't tested it on master. If it does happen on master we could punt on it and prioritize later.

jjeffryes:
This is closed in favor of #1864, since non-critical updates should be happening on eth-master instead.

PR #1861, opened 2020-01-15
Handle invalid mods better

jjeffryes:
This changes the behavior in settings/store so invalid moderators (moderators with models that don't validate) are still shown.

The card will validate the model and show an error if it is invalid, with the specific errors listed in the card.

Sometimes isValid returned true when I manually put in invalid data for the models, which is weird. I wasn't able to figure out the cause, so there's an extra layer of checking in the card template that protects the fee data, and shows the existing currency mismatch message instead of the currency information if the currency code for the fixedFee is unknown or invalid.

### DIFF
--- a/js/views/components/moderators/Card.js
+++ b/js/views/components/moderators/Card.js
@@ -91,7 +91,9 @@ export default class extends BaseVw {
   }
 
   get hasValidCurrency() {
-    const validFeeCur = isFiatCur(this.fixedFeeCur) || anySupportedByWallet([this.fixedFeeCur]);
+    const isFeeFiat = this.fixedFeeCur && isFiatCur(this.fixedFeeCur);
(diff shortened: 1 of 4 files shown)


js/views/components/moderators/Card.js
```diff
@@ -86,7 +91,10 @@ export default class extends BaseVw {
   }
 
   get hasValidCurrency() {
-    return anySupportedByWallet(this.modCurs);
+    const isFeeFiat = this.fixedFeeCur && isFiatCur(this.fixedFeeCur);
+    const isFeeCrypto = this.fixedFeeCur && anySupportedByWallet([this.fixedFeeCur]);
+    const validFeeCur = isFeeFiat || isFeeCrypto;
```
jjeffryes: I didn't see an existing method for checking if a currency code was a known fiat or crypto, if there is one that could be swapped in here.

jjeffryes:
This is closed in favor of #1864, since non-critical updates should be happening on eth-master instead.

PR #1864, opened 2020-01-17
[WIP] Various mod improvements

### LINKED ISSUE #1862: Loading text doesn't go away when searching for mods by peerId
This was tested against the `refactorModLoading` branch.

Steps to reproduce:
- In Settings > Store, search for a mod by peerId
- When the mod loads, click Find to search for the same mod.

Observed:
`Loading moderator 1 of 1` is shown and never goes away.

image

Expectations:
- Perhaps `Loading moderator 1 of 1` shows briefly or not at all, but it doesn't stay perpetually.

*Similar issue:*

Steps to reproduce:
- In Settings > Store, search for a valid Mod by peerId and wait until the mod loads.
- Search for a non-existent peer by peerId (the peerId should be a valid multihash, just not map to a reachable OB node)

Observed:
An error message is shown indicating the node was not reached and `Loading moderator 1 of 1` is shown and never goes away.

image

Expectations:
- `Loading moderator 1 of 1` disappears after it's determined the peerId was not found and the error message is shown.

jjeffryes:
This adds the same changes from #1859 and #1861 to the eth-master branch.

This also closes #1862.

This is built on top of #1863, and should not be merged until that branch is merged. The WIP label is just there to make sure that doesn't happen.


rmisio:
@jjeffryes Would you mind updating this with the latest `eth-master` (I merged #1863) so the change-set is a little lighter?

jjeffryes:
@rmisio I merged in eth-master, it's a lot more sane now. :)

rmisio:
When I search for a mod, I get double "loading" texts:

image

js/views/components/moderators/Moderators.js
```diff
@@ -215,13 +214,15 @@ export default class extends baseVw {
             const eventData = event.jsonData;
             if (eventData.error) {
               // errors don't have a message id, check to see if the peerID matches
-              if (IDs.includes(eventData.peerId)) {
-                // provide the expected capitalization of peerID
-                eventData.peerID = eventData.peerId;
-                delete eventData.peerId;
+              if (this.modsToFetch.includes(eventData.peerID)) {
```
rmisio: the "peer id" coming from the socket is `peerId` not `peerID`.

rmisio:
@jjeffryes One other thing I noticed is that it appears that when I toggle the Verified Moderators checkbox the whole payment screen is re-rendering. I haven't looked in the code, but I'm basing that on the fact that the images across the whole view flicker.

If that's the case, seems way too heavy handed. I would think just the moderators view would need some state updated.

Anyhow, I know that probably wasn't introduced in this PR, so let me know if it's better I just log an issue we could prioritize for some other time?

rmisio:
@jjeffryes Also please let me know if I should log a seperate issue for this or if you want to fix it in this PR:

https://github.com/OpenBazaar/openbazaar-desktop/issues/1862#issuecomment-581659203

jjeffryes:
> @jjeffryes One other thing I noticed is that it appears that when I toggle the Verified Moderators checkbox the whole payment screen is re-rendering. I haven't looked in the code, but I'm basing that on the fact that the images across the whole view flicker.
> 
> If that's the case, seems way too heavy handed. I would think that upon toggle just the moderators view would need some state updated.
> 
> Anyhow, I know that probably wasn't introduced in this PR, so let me know if it's better I just log an issue we could prioritize for some other time?

I'll see if I can fix that with a reasonable amount of effort when revising this PR, if not I'll create a new issue for it.

jjeffryes:
I made a separate issue for future rendering optimization. https://github.com/OpenBazaar/openbazaar-desktop/issues/1866

The rest of this should be ready for re-review.

